PR #30, opened 2015-12-15
Allow proj_string for AreaDefinition instantiation

mitkin:
Before a dictionary with proj4 parameters was required for AreaDefinition
instantiation. Now AreaDefinition can be instantiated with both,
dictionary and proj4 string

Reason to use proj4 string -- it is the most typical way
to define SRS (spatial referece system) parameters

Internally both dictionary and string are turned into a sorted OrderedDict
to make sure that the order of the parameters in the Proj definition is the same,
regardless of the variable type used to instantiate AreaDefinition


TomLav:
This is useful.

mitkin:
This is actually redundant since exactly same functionality has already been implemented in `utils.get_area_def()`. I overlooked it. That method accepts proj strings or keyword arguments, exactly what was implemented in this pull request. Closing.

PR #47, opened 2016-10-20
Add option to choose colormap

mitkin:
Make possible to indicate which colormap to use when plotting image (add 'cmap' kwarg).


davidh-ssec:
I changed the destination for this PR to pre-master since new code should go there first. I think master and pre-master were at the same commit though so this shouldn't affect the tests.

mitkin:
Great, thanks @davidh-ssec . Forgot to pick the right branch.

davidh-ssec:
The only other thing I can think of that might be needed with this is a unit test, but it's a pretty simple feature. I'll merge it and we can add a test later if @mraspaud thinks it is needed.

PR #71, opened 2017-09-20
Add image container for bilinear interpolation

pnuu:
This PR adds image container for bilinear interpolation. Data with shapes `(x, y)` and `(x, y, z)` are supported.


davidh-ssec:
Appveyor notification broke, but it looks like this passes. LGTM.

PR #75, opened 2017-10-03
Fix bil mask deprecation

pnuu:
This PR fixes how the invalid coordinates were handled in bilinear interpolation, indicated by deprecation warning raised by Numpy. If there's a better way of doing this (compare to the original version), I'm happy to update the PR.

Also UserWarning about possible extra neighbours inside search radius is now suppressed, as the number of retrieved neighbours is more or less static.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
 - '2.7'
-- '3.3'
 - '3.4'
 - '3.5'
 - '3.6'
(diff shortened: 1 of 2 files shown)


davidh-ssec:
So is the catch warnings supposed to catch a UserWarning or the warning from numpy when handling invalid coordinates and assigning NaN?

pnuu:
I should only catch the UserWarning, but now that look at it it might catch everything. I need to look more what happens.

pnuu:
After some testing, the only warning raised is the UserWarning about more than N neighbors within the search radius. This is unrelated to #72.

davidh-ssec:
So @pnuu is this ready to be merged then?

pnuu:
Yes, ready for merging.

PR #89, opened 2017-12-21
Fix masks in grid.get_resampled_image

### LINKED ISSUE #87: ImageContainerQuick does not mask fill values in high resolution images
``ImageContainerQuick.resample()`` does not mask fill values in high resolution images. The output is a masked array, but its mask is ``False``. It works fine with low resolutions though. The problem seems to be independent of the projection (I didn't check all). 

I'm using pyresample-1.5.0 with numpy-1.13.0 .

Example:

    import numpy as np
    import pyresample.geometry
    import pyresample.image
    import pyresample.plot
    from pycmsaf.geometry import get_proj_corners
    
    for nx, ny in zip([360, 3600], [180, 1800]):
        # Generate some test data with masked elements
        x = np.arange(nx, dtype='f4')
        y = np.arange(ny, dtype='f4')
        x *= 2 * np.pi / x.max()
        y *= 2 * np.pi / y.max()
        mx, my = np.meshgrid(x, y)
        data = np.sin(np.sqrt(mx ** 2 + my ** 2))
        data = np.ma.masked_where(data > 0.98, data)
    
        # Define regular lon-lat grid (lower left corner: [-180, -90], upper right
        # corner: [180, 90])
        regular_grid = pyresample.geometry.AreaDefinition(
            area_id='my_regular_grid',
            name='Regular 1x1 degree grid',
            proj_id='my_id',
            proj_dict={'proj': 'eqc', 'lon_0': 0},
            x_size=nx,
            y_size=ny,
            area_extent=[-20037508.342789244, -10018754.171394622,
                         20037508.342789244, 10018754.171394622])
    
        # Define web-mercator grid (lower left corner: [-180, -85], upper right
        # corner: [180, 85])
        proj4_str = '+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 ' \
                    '+x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null'
        web_mercator = pyresample.utils.get_area_def(
            area_id='my_web_mercator',
            area_name='Web Mercator',
            proj_id='my_id',
            proj4_args=proj4_str,
            x_size=ny,
            y_size=ny,
            area_extent=[-20037508.342789244, -19971868.88040857,
                         20037508.342789244, 19971868.880408563]
        )
    
        # Resample data from regular lon-lat grid to web-mercator grid
        con_reg = pyresample.image.ImageContainerQuick(
            image_data=data, geo_def=regular_grid, fill_value=None)
        con_merc = con_reg.resample(web_mercator)
        data_merc = con_merc.image_data
        print data_merc.mask
    
        # Plot data
        fig, ax = plt.subplots()
        bmap = pyresample.plot.area_def2basemap(web_mercator, resolution='c')
        bmap.ax = ax
        bmap.drawcoastlines()
        bmap.imshow(data_merc, origin='upper')
        ax.set_title('nx={0}, ny={1}'.format(nx, ny))
    
    plt.show()

Output:

lo
hi

Funkensieper:
Fix #87 

Use numpy.ma version of row_stack to prevent loosing the mask of large images (rows > cut_off).

### DIFF
--- a/pyresample/grid.py
+++ b/pyresample/grid.py
@@ -236,7 +236,7 @@ def get_resampled_image(target_area_def, source_area_def, source_image_data,
                 # First iteration
                 result = next_result
             else:
-                result = np.row_stack((result, next_result))
+                result = np.ma.row_stack((result, next_result))
 
         return result
     else:
--- a/pyresample/grid.py
+++ b/pyresample/grid.py
@@ -236,7 +236,11 @@ def get_resampled_image(target_area_def, source_area_def, source_image_data,
                 # First iteration
                 result = next_result
             else:
-                result = np.ma.row_stack((result, next_result))
+                if isinstance(next_result, np.ma.core.MaskedArray):
+                    stack = np.ma.row_stack
+                else:
+                    stack = np.row_stack
+                result = stack((result, next_result))
 
         return result
     else:
--- a/pyresample/test/test_grid.py
+++ b/pyresample/test/test_grid.py
@@ -152,6 +152,22 @@ class Test(unittest.TestCase):
         self.assertAlmostEqual(
             cross_sum, expected, msg='Resampling of image failed')
 
+    def test_resampled_image_masked(self):
+        # Generate test image with masked elements
+        data = np.ma.ones(self.msg_area.shape)
+        data.mask = np.zeros(data.shape)
+        data.mask[253:400, 1970:2211] = 1
+
+        # Resample image using multiple segments
+        target_def = self.area_def
+        source_def = self.msg_area
+        res = grid.get_resampled_image(
+            target_def, source_def, data, segments=4, fill_value=None)
+
+        # Make sure the mask has been preserved
+        self.assertGreater(res.mask.sum(), 0,
+                           msg='Resampling did not preserve the mask')
+
     @tmp
     def test_generate_linesample(self):
         data = np.fromfunction(lambda y, x: y * x * 10 ** -6, (3712, 3712))


davidh-ssec:
@Funkensieper So I'm guessing your low resolution image in #87 was resulting in only one segment? And your higher resolution image was split in to multiple segments?

Funkensieper:
Exactly! The threshold is at 500 rows.

davidh-ssec:
@Funkensieper what happens if the result is a normal numpy array? Does `np.ma.row_stack` convert them to a masked array?

pyresample/grid.py
```diff
@@ -236,7 +236,7 @@ def get_resampled_image(target_area_def, source_area_def, source_image_data,
                 # First iteration
                 result = next_result
             else:
-                result = np.row_stack((result, next_result))
+                result = np.ma.row_stack((result, next_result))
```
davidh-ssec: This forces the result to be a numpy masked array. Could you do:

```
if i == 0:
    ...
elif isinstance(result, np.ma.MaskedArray):
    result = np.ma.row_stack((result, next_result))
else:
    result = np.row_stack((result, next_result))
```

Funkensieper: Good point! I had already pushed when I saw your comment :) It is slightly different now

davidh-ssec: Not sure you needed the `core` part, but eh if it works then OK.

davidh-ssec: I'll wait for tests to pass then merge.

Funkensieper: Indeed, it works both with and without the ``core`` part. Good to know, thanks!

Funkensieper:
Please wait with merging, I will add a testcase for get_resampled_image with masked arrays later.

davidh-ssec:
Sounds good. Thanks.

I'm hoping to merge this PR and mine from last night some time today and get a new version out since these issues are causing problems for a lot of people.

Funkensieper:
Great, thanks!

PR #93, opened 2018-01-16
Add dask extra to setup.py to specify minimum dask version

davidh-ssec:
pyresample uses dask meshgrid which came in version 1.9. This PR adds that to setup.py even though it is only an 'extra'.


davidh-ssec:
Hm nevermind, this PR may be wrong. At least the wrong version. I was looking at http://dask.pydata.org/en/latest/array-api.html#dask.array.meshgrid and I thought it said it was added in version 1.9, but it says it was changed. BUT conda-forge only has dask 0.16.1 so not sure what's going on.

davidh-ssec:
Superseded by #94.

PR #94, opened 2018-01-22
Optimize xarray

davidh-ssec:
This optimizes how dask arrays are computed and used in the nearest neighbor resampling. It reduces the usage so that dask arrays *should* only be computed once and when possible should share as many calculations as possible. Two things that should be looked at:

- [ ] I added a FIXME to line 971 in `kd_tree.py` to double check that dask doesn't compute things for pixels that we don't use. We use some complicated indexing so I could see how it might not be able to. I had trouble producing a test case that matched what is done here.
- [ ] The `get_sample_from_neighbour_info` method is still not using dask arrays and should be changed to do that at some point. As discussed on slack there is a new method being added to xarray that should help with this (https://github.com/pydata/xarray/pull/1751), but from the simple tests I did I got terrible performance. Perhaps our use case is too large or perhaps I was just doing it wrong.

Note the backwards compatibility breaking change this PR includes where `BaseDefinition` no longer checks lon/lat input arrays for -180/180,-90/90 validity since doing that with dask arrays would be a huge waste of time and it is a huge waste of computation time even for normal numpy arrays. The usual use case is a user who already has valid inputs.

 - [ ] Passes ``git diff origin/develop **/*py | flake8 --diff`` (remove if you did not edit any Python files)
 - [x] Fully documented (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

### DIFF
--- a/pyresample/test/test_geometry.py
+++ b/pyresample/test/test_geometry.py
@@ -836,12 +836,12 @@ class TestSwathDefinition(unittest.TestCase):
 
         lons = np.array([[-45., 0., 45.],
                          [-90, 0., 90.],
-                         [-135., 180., 135.]]).T
+                         [-135., -180., 135.]]).T
 
         area = geometry.SwathDefinition(lons, lats)
         lons, lats = area.get_edge_lonlats()
 
-        np.testing.assert_allclose(lons, [-45., -90., -135., -135., 180., 135.,
+        np.testing.assert_allclose(lons, [-45., -90., -135., -135., -180., 135.,
                                           135., 90., 45., 45., 0., -45.])
         np.testing.assert_allclose(lats, [80., 80., 80., 80., 80., 80., 80.,
                                           80., 80., 80., 80., 80.])
--- a/pyresample/test/test_geometry.py
+++ b/pyresample/test/test_geometry.py
@@ -91,43 +91,6 @@ class Test(unittest.TestCase):
         self.assertTrue((cart_coords.sum() - exp) < 1e-7 * exp,
                         msg='Calculation of cartesian coordinates failed')
 
-    def test_base_lat_invalid(self):
-
-        lons = np.arange(-135., +135, 20.)
-        lats = np.ones_like(lons) * 70.
-        lats[0] = -95
-        lats[1] = +95
-        self.assertRaises(
-            ValueError, geometry.BaseDefinition, lons=lons, lats=lats)
-
-    def test_base_lon_wrapping(self):
-
-        lons1 = np.arange(-135., +135, 50.)
-        lats = np.ones_like(lons1) * 70.
-
-        with catch_warnings() as w1:
-            base_def1 = geometry.BaseDefinition(lons1, lats)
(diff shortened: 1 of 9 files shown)


davidh-ssec:
I couldn't get flake8 to run on my new local python environment. I kept getting:

```
  File "/Users/davidh/anaconda/envs/polar2grid_py36/lib/python3.6/site-packages/flake8/checker.py", line 435, in run_check
    return plugin['plugin'](**arguments)
  File "/Users/davidh/anaconda/envs/polar2grid_py36/lib/python3.6/site-packages/utils/checks.py", line 51, in sphinx_has_header
    if os.path.samefile(filename, './sphinx/util/smartypants.py'):
  File "/Users/davidh/anaconda/envs/polar2grid_py36/lib/python3.6/genericpath.py", line 97, in samefile
    s2 = os.stat(f2)
FileNotFoundError: [Errno 2] No such file or directory: './sphinx/util/smartypants.py'
```

davidh-ssec:
One more thing, pyresample should define a global default chunk size for arrays or a function to determine the chunk size. I had success with using a `blocksize` (the name of the keyword argument in pyresample) to 2048 compared to the default 5000. For a 2D array this results in a dask chunk size of 4096. I think if we assume most modern CPUs to have a close-to-CPU cache of 6-8MB then we should be ok. I do think having a power of 2 default would be best. No matter what is chosen it should come from one location in the code and would probably be best to have an environment variable override `CHUNK_SIZE = os.getenviron('PYRESAMPLE_CHUNK_SIZE', 4096)`.

davidh-ssec:
Another thing to note with the validity check being removed is that there isn't really a good way to put a deprecation warning on this since it was always happening whether the user wanted it to or not. We could leave the check in and add a deprecation warning...but let's not.

davidh-ssec:
image

This is what my simple satpy test case of loading 7 VIIRS SDR granules, loading I01, and resampling it looks like. The first large block of empty/white space I believe if the construction of the KDTree. The second smaller block is the generation of the arrays used `get_sample_from_neighbour_info` (can we rename this method to US English?).

pyresample/utils.py
```diff
@@ -513,6 +516,34 @@ def wrap_longitudes(lons):
     return (lons + 180) % 360 - 180
 
 
+def check_and_wrap(lons, lats):
```
mraspaud: We need a unit test for this function

pyresample/test/test_geometry.py
```diff
@@ -755,27 +718,6 @@ def test_swath(self):
         self.assertFalse(id(lons1) != id(lons2) or id(lats1) != id(lats2),
                          msg='Caching of swath coordinates failed')
 
-    def test_swath_wrap(self):
-        lons1 = np.fromfunction(lambda y, x: 3 + (10.0 / 100) * x, (5000, 100))
```
mraspaud: Maybe this could be reused ?

PR #103, opened 2018-02-28
Add comparison methods to area definitions

davidh-ssec:
Based on the python docs I thought it would be better to implement each method separately since it sounds like there might be a small performance penalty to using the `total_ordering` decorator. 
This PR includes better chunk handling in AreaDefinitions so that you can ask for each dimension separately.

I should probably add some tests for this.

 - [ ] Tests added (for all bug fixes or enhancements)
 - [ ] Tests passed (for all non-documentation changes)
 - [ ] Passes ``git diff origin/develop **/*py | flake8 --diff`` (remove if you did not edit any Python files)
 - [ ] Fully documented (remove if this change should not be visible to users, e.g., if it is an internal clean-up, or if this is part of a larger project that will be documented later)

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -119,6 +119,41 @@ class BaseDefinition(object):
 
         return not self.__eq__(other)
 
+    def _comparable(self):
+        return self.shape
+
+    def __lt__(self, other):
+        """Which area is higher resolution (larger size)."""
+        if not isinstance(other, self.__class__):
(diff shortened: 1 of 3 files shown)


davidh-ssec:
Closing this PR after reconsideration of how area/swath definitions should be compared. Comparing just on resolution may not be what we need in the future. This was originally intended to help with finding the highest resolution area out of a list of them, but implementing the comparisons in the classes was overkill.

PR #138, opened 2018-08-13
Add `create_area_def` utility method and refactor AreaDefinition arguments

wroberts4:
- [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented 

Starting near the bottom of utils.py, from_params will take an area_id and projection along with shape, top_left_extent, center, area_extent, resolution, and/or radius and attempts to find an area definition from what has been given. Units can be in radians, degrees, meters, or any parameter from cs2cs. This feature is overall an ease of life addition. Multiple functions in utils.py have also been updated to call from_params.

If not enough information is provided to create an area definition, but either shape or area_extent can be found, a DynamicAreaDefinition will be created. Else a ValueError occurs.

There are four new convenience methods in AreaDefinition: from_extent, from_circle, from_area_of_interest, and from_geotiff (located right below from_params) which have positional arguments that will guarantee an area definition is created (except for from_circle which needs either pixel_size or shape). These each call from_params.


docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
  Number of rows: 425
  Area extent: (-5326849.0625, -5326849.0625, 5326849.0625, 5326849.0625)
 
+The function **from_params** attempts to find shape and area_extent with the given data below:
```
djhoese: I think we should start using sphinx's links to python objects when possible. See https://raw.githubusercontent.com/pytroll/satpy/master/doc/source/dev_guide/xarray_migration.rst for some examples. I can't find the actual documentation for this but pycharm does a pretty good job of autocompleting if I remember correctly. If you add this syntax for the first use or two of a python object (function, method, class, etc) it should give the user a nice obvious way to get to the API docs.

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+* **projection**: projection parameters as a proj4_dict or proj4_string
+
+Optional (keyword) arguments:
 
-The **load_area** function can be used to parse area definitions from a
-configuration file. Assuming the file **areas.yaml** exists with the following
-content
+* **area_id**, **proj_id**, and **area_extent**: same as AreaDefinition
+* **units**: Default projection units (meters/radians/degrees)
+* **shape**: (height, width). Note: if height = width, then only height or width needs to be passed.
```
djhoese: Maybe add `in number of pixels` to this description and remove the mention of shape in the `units` part below. Also maybe remove the Note part from each of these and put it after and say that `shape, pixel_size, and radius can be specified with one value when they are the same (I'm starting to think accepting one value was a mistake code-wise, but you've already done it so leave it).

I think you can combine the "where" descriptions with the individual arguments here and simplify the description to something like `x and y coordinate of upper left corner...`.

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+* **units** accepts anything with 'm', 'rad', 'deg' or '°'. The order of default is: units expressed
+  with each variable, units passed to **units**, units used in **projection**, then meters. **shape** is not affected
+  by units. To add units to a specific variable, you can make an xarray.DataArray with a variable and then use
+  {'units': unit_of_choice} as its attrs attribute.
+
+**from_params** returns the following outcomes:
+
+1. If shape and area_extent are found, an AreaDefinition is returned
+2. If only shape or area_extent can be found, a DynamicAreaDefinition is returned
+3. If neither shape nor area_extent can be found, an error will occur
```
djhoese: "an exception is raised"

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+* **height**: number of pixels in y direction (number of grid rows)
+* **width**: number of pixels in x direction (number of grid columns)
+* **x_size**: projection size of pixels in the x direction
+* **y_size**: projection size of pixels in the y direction
+* **x_length**: projection length from center to left/right outer edge
+* **y_length**: projection length from center to top/bottom outer edge
+* **units** accepts anything with 'm', 'rad', 'deg' or '°'. The order of default is: units expressed
+  with each variable, units passed to **units**, units used in **projection**, then meters. **shape** is not affected
+  by units. To add units to a specific variable, you can make an xarray.DataArray with a variable and then use
+  {'units': unit_of_choice} as its attrs attribute.
```
djhoese: I'm not a huge fan of discussing this here in this document. It complicates things a lot. Might be better just to keep that information in the docstring.

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+
+.. doctest::
+
+ >>> from pyresample import utils
+ >>> from xarray import DataArray
+ >>> area_id = 'ease_sh'
+ >>> description = 'Antarctic EASE grid'
+ >>> proj_id = 'ease_sh'
+ >>> projection = {'a': '6371228.0', 'units': 'm', 'lon_0': '0', 'proj': 'laea', 'lat_0': '-90'}
+ >>> area_extent = DataArray((-135.0, -17.516001139327766, 45.0, -17.516001139327766), attrs={'units': 'degrees'})
```
djhoese: Similar to above, I'm not sure I like showing examples of using this `DataArray` logic. This is a last resort as is and I'm not sure I want to subject users to it. Especially since a lot of users reading this code are new to projects and xarray so adding on this complicated construct/object is probably not great.

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+When composing a yaml file, things to keep in mind are:
+
+* If the elements of **shape**, **pixel_size**, or **radius** respectively are the same, you may define the
+  variable by the element alone. Example: [0, 0] = 0.
+* **area_id** defaults to the name used to specify the area definition (in this case ease_sh or ease_nh).
+* You may add a key to a variable with the same name as the variable itself. This variable-key has the
+  exact same parameters as the variable it is named after. Note: This only makes sense if you are adding units.
+* Only a variable/variable-key, **lower_left_xy**, and **upper_right_xy** can be expressed as a list.
+* A variable-key will override other data included except for **units**. An example using **shape**:
+  If shape: [425, 425], height: 850, and width: 850 are provided, **shape** will be [425, 425].
```
djhoese: This should raise an error if the user is specifying the same thing multiple ways, at least if I'm understanding this correctly.

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+ Number of columns: 425
+ Number of rows: 425
+ Area extent: (-5326849.0625, -5326849.0625, 5326849.0625, 5326849.0625)
+
+The **load_area** function can be used to parse area definitions from a configuration file by giving it the area file
+name and regions you wish to load. **load_area** takes advantage of **from_params** and hence uses the same arguments.
+
+When composing a yaml file, things to keep in mind are:
+
+* If the elements of **shape**, **pixel_size**, or **radius** respectively are the same, you may define the
```
djhoese: I don't think this is needed.

docs/source/geo_def.rst
```diff
@@ -88,10 +89,83 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+  variable by the element alone. Example: [0, 0] = 0.
+* **area_id** defaults to the name used to specify the area definition (in this case ease_sh or ease_nh).
+* You may add a key to a variable with the same name as the variable itself. This variable-key has the
```
djhoese: This is extremely confusing. Maybe leave it to examples below to describe these situations by showing instead of telling (I don't think you even need to describe it with words). This entire set of "keep in mind" is probably best as a show rather than a tell.

docs/source/geo_def.rst
```diff
@@ -109,23 +183,42 @@ content
    area_extent:
      lower_left_xy: [-5326849.0625, -5326849.0625]
      upper_right_xy: [5326849.0625, 5326849.0625]
+   top_left_extent: [-5326849.0625, 5326849.0625]
```
djhoese: Why are the extents and the area_extent both specified? I don't think this is a realistic use case. The examples listed here should stick with the original 3 or 4 use cases that this PR was originally addressing and maybe a couple extra that show the same/similar use cases but with units specified or a 2-element parameter specified with one element.

pyresample/geometry.py
```diff
@@ -677,7 +675,7 @@ def compute_domain(self, corners, resolution=None, size=None):
             x_resolution = (corners[2] - corners[0]) * 1.0 / (x_size - 1)
             y_resolution = (corners[3] - corners[1]) * 1.0 / (y_size - 1)
 
-        if resolution:
+        elif resolution:
```
djhoese: this could probably just be `else`, right? Maybe it is clearer this way.

pyresample/geometry.py
```diff
@@ -729,10 +726,11 @@ def freeze(self, lonslats=None,
             yarr[yarr > 9e29] = np.nan
             corners = [np.nanmin(xarr), np.nanmin(yarr),
                        np.nanmax(xarr), np.nanmax(yarr)]
-
-            domain = self.compute_domain(corners, resolution, size)
+            # For shape and area_extent to be defined, resolution or size must be defined.
+            domain = None
+            if not (resolution is None and size is None):
+                domain = self.compute_domain(corners, resolution, size)
             self.area_extent, self.x_size, self.y_size = domain
```
djhoese: Won't this line fail if either of them is specified? Maybe `compute_domain` should check for neither of them to be specified and for both of them specified (as it is doing already)?

docs/source/geo_def.rst
```diff
@@ -88,61 +96,189 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
+
+.. doctest::
+
+ >>> from pyresample import utils
+ >>> from xarray import DataArray
+ >>> area_id = 'ease_sh'
+ >>> projection = {'a': '6371228.0', 'units': 'm', 'lon_0': '0', 'proj': 'laea', 'lat_0': '-90'}
+ >>> area_def = utils.from_params(area_id, projection, center=(0, -90),
+ ...                              radius=(45, -17.516001139327766),
+ ...                              pixel_size=(45, -89.681194), units='degrees',
```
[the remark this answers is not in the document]
wroberts4: The conversion between angles and meters is a little goofy in these cases. The function treats the data as points on a grid; the second angle is used to find the length of the hypotenuse from (0,0) to (25067.525,25067.525), while the first angle is used to find the x and y components of said hypotenuse. Thus pixel_size=(45, -89.681194) means that x_size = cos(45)*sin((-89.681194+90)/2)*12742456 and y_size = sin(45)*sin((-89.681194+90)/2)*12742456 where 12742456 is the max distance. Note: all these angles are in degrees.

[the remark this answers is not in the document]
wroberts4: Conceptually, the seconds angle is going from -90 to what was provided which specifies an arc, multiplying by distance from -90 to 90 gives us a distance, and then basic trig using the first angle breaks down the x and y components.

pyresample/geometry.py
```diff
@@ -802,8 +799,8 @@ class AreaDefinition(BaseDefinition):
         Pixel width in projection units
     pixel_size_y : float
         Pixel height in projection units
-    pixel_upper_left : list
-        Coordinates (x, y) of center of upper left pixel in projection units
+    top_left_extent : list
+        Coordinates (x, y) of extent of upper left pixel in projection units
```
mraspaud: I think `center` is an important word here, it shouldn't be removed

wroberts4: After consulting with @djhoese, we decided it was better to use the top-left corner of the top-left pixel for area definitions, rather than the center of the top-left pixel. Hence, leaving "center" in the description would be incorrect.

docs/source/geo_def.rst
```diff
@@ -1,41 +1,41 @@
 Geometry definitions
 ====================
-The module **pyresample.geometry** contains classes for describing different kinds of types
-of remote sensing data geometries. The use of the different classes is described below.
+The module :mod:`pyresample.geometry <geometry>` contains classes for describing different kinds
```
djhoese: Does this make the text say "pyresample.geometry" and link to a document called "geometry" or does it make the text say "geometry" linked to "pyresample.geometry"?

If the latter I think this can be changed to ":mod:`~pyresample.geometry`".

docs/source/geo_def.rst
```diff
@@ -1,41 +1,41 @@
...
+Use function :mod:`utils.wrap_longitudes <utils.wrap_longitudes>` for wrapping longitudes yourself.
 
 AreaDefinition
 --------------
 
-The cartographic definition of grid areas used by Pyresample is contained in an
-object of type AreaDefintion. The following arguments are needed to initialize
-an area:
+The cartographic definition of grid areas used by Pyresample is
+contained in an object of type :mod:`AreaDefinition <geometry.AreaDefinition>`
```
djhoese: Does this `mod` work? `AreaDefinition` is not a module.

docs/source/geo_def.rst
```diff
@@ -63,22 +62,25 @@ Creating an area definition:
...
 ****************
-The utils module of pyresample has convenience functions for constructing
-area defintions. The function **get_area_def** can construct an area definition
-based on area extent and a proj4-string or a list of proj4 arguments.
+
+The :mod:`utils <utils>` module of Pyresample
+has convenience functions for constructing area definitions. The function
+:mod:`get_area_def <utils.get_area_def>` can construct an
+:mod:`AreaDefinition <geometry.AreaDefinition>` object based on
+area_extent and a proj4-string/dict or a list of proj4 arguments.
```
djhoese: What does "list of proj4 arguments" mean here?

docs/source/geo_def.rst
```diff
@@ -63,22 +62,25 @@ Creating an area definition:
...
  >>> area_id = 'ease_sh'
  >>> description = 'Antarctic EASE grid'
  >>> proj_id = 'ease_sh'
- >>> projection = '+proj=laea +lat_0=-90 +lon_0=0 +a=6371228.0 +units=m'
- >>> x_size = 425
- >>> y_size = 425
- >>> area_extent = (-5326849.0625,-5326849.0625,5326849.0625,5326849.0625)
- >>> area_def = utils.get_area_def(area_id, description, proj_id, projection,
- ...                  			   x_size, y_size, area_extent)
+ >>> proj_string = '+a=6371228.0 +units=m +lon_0=0 +proj=laea +lat_0=-90'
```
djhoese: Is there a reason you reordered these? The original order is fairly common and is easier to understand from a user perspective (proj first, then other parameters, then units last).

docs/source/geo_def.rst
```diff
@@ -88,15 +90,200 @@ based on area extent and a proj4-string or a list of proj4 arguments.
  Number of rows: 425
  Area extent: (-5326849.0625, -5326849.0625, 5326849.0625, 5326849.0625)
 
+The function :mod:`from_params <utils.from_params>` attempts to return
```
djhoese: Same question as before, `mod` doesn't seem right here.

docs/source/geo_def.rst
wroberts4: I didn't touch the wording of that, but I can reword it! It means that instead of providing '+a=6371228.0 +units=m +lon_0=0 +proj=laea +lat_0=-90', you could provide ['a=6371228.0', 'units=m', 'lon_0=0', 'proj=laea', 'lat_0=-90']. Since I didn't make the get_area_def function, I cannot speak as to why it is this way. I could add an example to show this functionality though!

docs/source/geo_def.rst
wroberts4: I used what was already in the docs for ordering, but I can order it as you suggested!

docs/source/geo_def.rst
wroberts4: The functionality is :any:`Name_Displayed <Path_To_Thing>`. So it displays "pyresample.geometry" and links to the geometry module.

docs/source/geo_def.rst
wroberts4: :mod:`AreaDefinition <geometry.AreaDefinition>` does work. However it's probably misleading for people reading the code. I will use "func" for functions and "mod" for modules!

docs/source/geo_def.rst
djhoese: I almost think we should remove the list of proj4 arguments in the documentation. This is probably only supported because of the legacy format and how it formatted things. @mraspaud may have other thoughts.

pyresample/utils.py
```diff
@@ -692,3 +697,388 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+                            'Rounding shape to {1}'.format(shape, (height, width)))
+    return height, width
+
+
+def _extrapolate_information(area_extent, shape, center, radius, resolution, top_left_extent, units, p, proj_dict):
+    """Attempts to find shape and area_extent based on data provided.
+
+    Parameters are used in a specific order to determine area_extent and shape.
+    Providing some parameters may have no effect if other parameters could be
+    to used determine this information. The order of
```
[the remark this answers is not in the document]
djhoese: Oops I committed something I had started typing and didn't finish. Will fix in a little bit.

docs/source/conf.py
```diff
@@ -239,3 +238,16 @@ def __getattr__(cls, name):
 
 # If false, no module index is generated.
 #latex_use_modindex = True
+
```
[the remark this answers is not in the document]
djhoese: This comes from the sphinx template. I don't see a reason to update it.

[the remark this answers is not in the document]
mraspaud: I thought we wanted to be pep8 compliant all the way ? :)

pyresample/test/test_geometry.py
```diff
@@ -1295,21 +1296,111 @@ def test_append_area_defs(self, adef):
...
+
+        # Tests that incorrect lists do not create an area definition, that both projection strings and
+        # dicts are accepted, and that degrees, meters, and radians all create the same area definition.
+        # area_list used to check that areas are all correct at the end.
+        area_list = []
+        for projection in projection_list:
+            for units in units_list:
+                for center in center_list:
+                    # essentials = center, radius, upper_left_extent, resolution, shape.
+                    if 'm' in units:
```
mraspaud: that includes `km`, `cm`, etc...

pyresample/utils.py
```diff
@@ -124,53 +146,60 @@ def _parse_yaml_area_file(area_file_name, *regions):
...
+        params['center'] = _get_list(params, 'center', ['x', 'y'])
+        params['area_extent'] = _get_list(params, 'area_extent', ['lower_left_xy', 'upper_right_xy'])
+        params['resolution'] = _get_list(params, 'resolution', ['dx', 'dy'])
+        params['radius'] = _get_list(params, 'radius', ['dx', 'dy'])
+        params['rotation'] = _get_list(params, 'rotation', [])
+        res.append(from_params(**params))
     return res
 
 
+def _get_list(params, var, arg_list):
```
mraspaud: `params` isn't used in this function appart from getting `var` from it. I would rather have it called like `_get_list(params['shape'], ['height', 'width'])` for example.

pyresample/utils.py
```diff
@@ -124,53 +146,60 @@ def _parse_yaml_area_file(area_file_name, *regions):
...
+            values = variable.get(arg)
+            if values is not None:
+                if arg == var:
+                    list_of_values = values
+                elif arg in ('lower_left_xy', 'upper_right_xy') and isinstance(values, list):
+                    list_of_values.extend(values)
+                else:
+                    list_of_values.append(values)
+        except AttributeError:
+            raise ValueError('Incorrect yaml: {0} has too many arguments: Both {0} and {1} were specified.'.format(var,
```
mraspaud: What does yaml have to do with it ?

pyresample/utils.py
```diff
@@ -124,53 +146,60 @@ def _parse_yaml_area_file(area_file_name, *regions):
...
+    variable = params.get(var)
+    if not isinstance(variable, dict):
+        return variable
+    list_of_values = []
+    # Add var as a key in case users want to express the entire variable with units. Insert
+    # at 0 to allow Attribute errors when too many arguments are specified.
+    arg_list.insert(0, var)
+    # Iterate through dict.
+    for key in variable.keys():
+        if key not in arg_list and (var == 'shape' or key != 'units'):
```
mraspaud: I'm not understanding this, aren't area defs allowed to have a shape ?

pyresample/utils.py
```diff
@@ -124,53 +146,60 @@ def _parse_yaml_area_file(area_file_name, *regions):
...
+        params['area_extent'] = _get_list(params, 'area_extent', ['lower_left_xy', 'upper_right_xy'])
+        params['resolution'] = _get_list(params, 'resolution', ['dx', 'dy'])
+        params['radius'] = _get_list(params, 'radius', ['dx', 'dy'])
+        params['rotation'] = _get_list(params, 'rotation', [])
+        res.append(from_params(**params))
     return res
 
 
+def _get_list(params, var, arg_list):
+    """Reads a list-like param variable."""
```
mraspaud: The docstring could be expanded, an example added.

pyresample/utils.py
```diff
@@ -692,3 +711,411 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+        If only shape or area_extent can be found, a DynamicAreaDefinition object is returned
+
+    Raises
+    ------
+    ValueError:
+        If neither shape nor area_extent could be found
+
+    Notes
+    -----
+    * **units** accepts 'deg', 'degrees', 'rad', 'radians', 'meters', and any parameter from cs2cs
```
mraspaud: why not stick to proj style ? m, deg, rad

pyresample/utils.py
```diff
@@ -692,3 +711,411 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+    proj_dict = _get_proj_data(projection)
+    p = Proj(proj_dict, preserve_units=True)
+
+    # If no units are provided, try to get units used in proj_dict. If still none are provided, use meters.
+    if units is None:
+        units = proj_dict.get('units', 'm')
+
+    # Makes sure list-like objects are list-like, have the right shape, and contain only numbers.
+    center = _verify_list('center', center, 2)
+    radius = _verify_list('radius', radius, 2)
```
mraspaud: aren't these 2 allowed to have just one value ?

pyresample/utils.py
```diff
@@ -692,3 +711,411 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+    else:
+        center = p(*center, inverse=True, errcheck=True)
+        if abs(abs(center[1]) - 90) < error:
+            center = (center[0], _sign(center[1]) * 90)
+        center = p(*center, errcheck=True)
+    return center
+
+
+def _distance_from_center_forward(var, center, p, is_radians):
+    """Convert distances in degrees to projection units."""
```
mraspaud: radians seem to be accepted too ?

pyresample/utils.py
```diff
@@ -692,3 +711,411 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+    # Interprets radius and resolution as distances between latitudes/longitudes.
+    # Since the distance between longitudes and latitudes is not constant in
+    # most projections, there must be reference point to start from.
+    if center is None:
+        raise ValueError('center must be given to convert radius or resolution from an angle to meters')
+
+    center_as_angle = p(*center, radians=is_radians, inverse=True, errcheck=True)
+    unit_conversion = 90
```
mraspaud: could you find a better variable name ?

pyresample/utils.py
```diff
@@ -692,3 +711,411 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+    elif not is_angle and inverse:
+        # Converts list-like from meters to degrees.
+        var = p(*var, inverse=True, errcheck=True)
+    if name in ['radius', 'resolution']:
+        var = (abs(var[0]), abs(var[1]))
+    return var
+
+
+def _validate_variable(var, new_var, var_name, input_list):
+    """Makes sure data given does not conflict with itself."""
```
mraspaud: Could this be explained a bit more ?

pyresample/utils.py
```diff
@@ -692,3 +711,411 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+    projection : dict or str
+        Projection parameters as a proj4_dict or proj4_string
+    description : str, optional
+        Description/name of area. Defaults to area_id
+    proj_id : str, optional
+        ID of projection (being deprecated)
+    units : str, optional
+        Default projection units: meters, radians, or degrees
+    shape : list, optional
+        Number of pixels in the x and y direction (height, width)
```
mraspaud: please be consistent, it should by `y` and `x`

docs/source/geo_def.rst
```diff
@@ -1,84 +1,69 @@
...
 
 AreaDefinition
 --------------
 
-The cartographic definition of grid areas used by Pyresample is contained in an
-object of type AreaDefintion. The following arguments are needed to initialize
-an area:
+An :class:`~pyresample.geometry.AreaDefinition`, or ``area``, is the primary
+way of specifying a uniformly spaced geographic region in pyresample. It is
+also one of the only geometry objects that understands geographic projects.
```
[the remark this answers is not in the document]
djhoese: Yes. This was a typo.

docs/source/geo_def.rst
```diff
@@ -1,84 +1,69 @@
...
+in X/Y coordinates in meters. See the PROJ.4 documentation for more
+information on projections and coordinate reference systems.
 
-* **area_id** ID of area
-* **name**: Description
+The following arguments are needed to initialize an area:
+
+* **area_id**: ID of area
+* **description**: Description
 * **proj_id**: ID of projection (being deprecated)
```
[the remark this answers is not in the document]
djhoese: Do you want a fully incompatible class interface? Right now the positional arguments still mean the same thing even though they have different names. I can add a warning if the `proj_id` is not None or doesn't equal area_id, but not sure that helps much since code would still break in the future.

docs/source/geometry_utils.rst
```diff
@@ -0,0 +1,430 @@
...
+The ``from_params`` function has the following required arguments:
+
+* **area_id**: ID of area
+* **projection**: Projection parameters as a proj4_dict or proj4_string
+
+and optional arguments:
+
+* **description**: Human-readable description. If not provided, defaults to **area_id**
+* **proj_id**: ID of projection (being deprecated)
+* **units**: Default projection units: meters, radians, or degrees
```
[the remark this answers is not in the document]
djhoese: I think this description is misleading. This `units` (as far as I remember, @wroberts4 correct me if I'm wrong) is to specify what units the other parameters (center, resolution, etc) are in and doesn't actually affect the projection.

docs/source/geo_def.rst
```diff
@@ -88,328 +73,161 @@ based on area extent and a proj4-string or a list of proj4 arguments.
...
- Area ID: ease_sh
- Description: Antarctic EASE grid
- Projection ID: ease_sh
- Projection: {'a': '6371228.0', 'lat_0': '-90.0', 'lon_0': '0.0', 'proj': 'laea', 'units': 'm'}
- Number of columns: 425
- Number of rows: 425
- Area extent: (-5326849.0625, -5326849.0625, 5326849.0625, 5326849.0625)
+Creating an ``AreaDefinition`` can be complex if you don't know everything
+about the region being described. Pyresample provides multiple utilities
+for creating areas as well as storing them on disk for repeated use. See
```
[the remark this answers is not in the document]
djhoese: Good point. I was more talking about the YAML reading, but I suppose we could talk about printing using the `create_areas_def`. Do we want to take this opportunity to consider other names for this function? `dump`? Just an idea after looking at the contents of the function. Not a very strong feeling.

pyresample/utils.py
djhoese: @mraspaud You are right that `params` isn't used more than that, but `var` is used a lot to know exactly what we are dealing with. Besides keeping it as is, another option would be to split the function up per-parameter but I think most of the functionality is shared so I'm not sure it would be pretty/easy.

pyresample/utils.py
djhoese: This could be made cleared, maybe "Invalid area specification" or something, but to be clear this function is only used when loading data from a YAML file.

pyresample/utils.py
djhoese: If I'm understanding this, this chunk of code only happens if the yaml for this parameter is a dictionary. If the key in the yaml sub-dict isn't `units` and it isn't what we asked for then that's an invalid argument. The `shape` check (I think) is because `shape` can't have `units`, it doesn't make sense. @wroberts4 could you add a comment on this?

pyresample/utils.py
djhoese: More flexible I guess. I'm ok limiting the amount of unit variations that are allowed. I just thought users might want to specify it different ways so I was ok with this. Thoughts? I suppose limited would be better.

pyresample/utils.py
djhoese: Suggestions? `conversion_factor`?

mraspaud: the point was that it is used in a subtraction also, so I felt it was more than just a unit conversion factor

docs/source/geometry_utils.rst
```diff
@@ -0,0 +1,430 @@
...
+ >>> area_def = AreaDefinition.from_area_of_interest(area_id, proj_dict, center, resolution, shape)
+ >>> print(area_def)
+ Area ID: ease_sh
+ Description: ease_sh
+ Projection: {'a': '6371228.0', 'lat_0': '-90.0', 'lon_0': '0.0', 'proj': 'laea', 'units': 'm'}
+ Number of columns: 425
+ Number of rows: 425
+ Area extent: (-5326849.0625, -5326849.0625, 5326849.0625, 5326849.0625)
+
+from_geotiff
```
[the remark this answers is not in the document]
mraspaud: `from_ul_corner` ?

pyresample/utils.py
mraspaud: `Invalid area specification` would be better indeed.

docs/source/geo_def.rst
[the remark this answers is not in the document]
mraspaud: By removing `get_area_def` we are breaking backwards compatibility, so this will be released as 2.x.x anyway...

docs/source/geo_def.rst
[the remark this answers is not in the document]
mraspaud: Yes, we can change the method name if you want.

docs/source/geometry_utils.rst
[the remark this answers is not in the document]
mraspaud: That can be made clearer indeed. But from what I understand, theses parameters and units are still passed to proj at som point, so units could be anything proj supports, right ?

pyresample/utils.py
mraspaud: then you need to add `metres` :)

pyresample/utils.py
djhoese: Yeah so maybe it should be something like `center_reference`.

docs/source/geo_def.rst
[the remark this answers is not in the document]
djhoese: I didn't remove `get_area_def`, only talked about it on slack. I was going to deprecate it but it's recent functionality of returning an AreaDefinition *or* a DynamicAreaDefinition is not available anywhere else. If `from_params` is renamed to `get_area_def` and provides the same interface I don't think we need to go all the way to 2.x.x.

I'd prefer to release this functionality as a 1.x release and have a real 2.x design process (with issues marked for 2.0 and really considering new interface opportunities).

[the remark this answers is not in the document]
mraspaud: We leave this be then.

docs/source/geometry_utils.rst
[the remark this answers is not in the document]
djhoese: I think you're right. I forget what @wroberts4 ultimately ended up implementing. I also think the possible units are linked to somewhere, they could be copied here.

docs/source/geometry_utils.rst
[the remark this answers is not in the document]
djhoese: That works for me.

docs/source/geo_def.rst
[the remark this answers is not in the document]
djhoese: I'll save that for a separate PR where I rewrite a lot of the documentation. Probably a PR specifically for this method rename. Any idea if people use `create_areas_def` outside of pyresample much? I feel like I've only seen it when trying to help people create an areas def file.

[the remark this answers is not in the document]
mraspaud: I'm not aware of people using it outside pyresample. But we could just add a deprecation warning and remove it in 2.x.

pyresample/utils.py
wroberts4: Correct, radius and resolution may be just one value. The reason why I check for a size of 2 is because if a single number is provided, it is converted into a list of size 2 (which is internally how these variables function); there is not a case where a single number is provided such that my code will complain about a user only providing one value. If the user provides a list with only one element (rather than a stand-alone number), I interpret that as a mistake and do not allow that.

docs/source/geometry_utils.rst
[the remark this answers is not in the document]
wroberts4: @mraspaud is correct in both regards (these units only affect how user arguments are interpreted, NOT the projection being made), the description was not updated to incorporate the same units that proj accepts. I provide more in-depth detail in the doc-string for my function, but will update this description.

mraspaud:
There is some documentation that needs to be change to reflect the `x_size` and `y_size` change: https://pyresample.readthedocs.io/en/latest/geo_def.html

djhoese:
@mraspaud You linked to documentation that was not built from this PR. The `geo_def.rst` module has been updated in this PR to reflect the changes to `x_size` and `y_size`. Please verify and I'll update anything that needs updating.

djhoese:
So are you approving this PR? Are these argument name changes big enough that we have to wait for a 2.0 release for this? If so, as mentioned before, I'd like to plan for much larger changes to pyresample before a 2.0 release.

djhoese:
As mentioned on slack and other issues (I think), I've been waiting on merging this issue so that I could perform larger changes/refactorings to pyresample like splitting the utils module in to a subpackage and cleaning up some of the documentation. My understanding of the discussion here is that this PR was as good as it was going to get without user feedback and wider use (to see what users expect and what doesn't make sense). Most of the edge cases that @wroberts4 handled in this PR were guesses based on my suggestions or his intuition and there are plenty of cases where we may have made bad choices. Time for users to tell us that.

So I am merging this PR so I can start work on other things with pyresample.

PR #140, opened 2018-09-07
Fix area parsing code to convert PROJ.4 parameters to float if possible

djhoese:
Recreation of #139 because I accidentally merged it when I pushed to the wrong remote.


djhoese:
@mraspaud As mentioned this PR changes the default behavior of the AreaDefinition to store all projection information as a float when possible. The main purpose of this PR is to fix the config loading functions so they don't pass these parameters as strings but for consistency I made the AreaDefinition do it too. I could see this being a big change for anyone who was specifically depending on the string but I also can't imagine that those people wouldn't be better off using the float.

So the question is, does this get released as a bug fix (what it really is) or as a minor release given that it is an important change?

PR #159, opened 2019-02-12
Fix 'typing' import with 'collections' for python 2 compatibility

djhoese:
The 'typing' module was used to import the `Mapping` class for a small utility function. This was causing compatibility issues for environments with python <3.5. The `Mapping` object should be available from the `collections` module in all supported versions of python.

 - [ ] Tests added 
 - [x] Tests passed 


djhoese:
Since I didn't hear anything from you guys, I assume this won't cause any issues. I'll merge this tomorrow morning.

PR #162, opened 2019-02-18
Fix create_area_def around poles

wroberts4:
- [x] Closes #xxxx 
 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented 

Currently entering a radius or pixel_size in degrees when the center is at the poles (and normal lat/long is not the same as the center lat/long) does not yield the expected result. This change makes sure that the x direction is treated the same as the y direction and hence makes a square when both the x and y dimensions are equal.

### DIFF
--- a/pyresample/utils.py
+++ b/pyresample/utils.py
@@ -176,7 +176,7 @@ def _capture_subarguments(params, arg_name, sub_arg_list):
           dx: 11
           dy: 22
           units: meters
-    returns DataArray((11, 22), attrs={'units': 'meters})
+        # returns DataArray((11, 22), attrs={'units': 'meters})
     """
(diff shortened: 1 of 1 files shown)


djhoese:
I have done a lot of reorganization of this code. You'll have to merge and put your changes in `pyresample/area_config.py` I'm guessing. Also please edit your pull request "comment" to add a description of what this PR does so people can get a summary of what has changed.

djhoese:
Coming back to this and as discussed in person with @wroberts4, I think this is a nice simplification. This is one of those things that most people won't care about or run in to and when someone does care a lot about this, they can make their own PR to change it.

PR #164, opened 2019-03-02
Delete unused requirements.txt file

djhoese:
Not used and only references basemap which is deprecated. @pytroll/core let me know if I'm missing something.


djhoese:
That's enough evidence for me. Merging.

PR #166, opened 2019-03-06
Inherit kd_tree resampler for bilinear interpolation

pnuu:
This PR updates bilinear interpolation so that it inherits `XArrayResamplerNN` as a base.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


pnuu:
So far this PR has only made the bilinear interpolation much slower. With the script below, it takes 8 m 40 s for the first run and 4 m 7 s for the second run on my laptop. The same figures for `master` branch are 4 m 40 s / 2 m 42 s.

Performance for `nearest` hasn't changed with the small changes that were needed.

```python
#!/usr/bin/env python

import glob                                                            
from satpy import Scene                                                
fnames = glob.glob('/home/lahtinep/data/satellite/geo/msg/H*')
glbl.load(['overview', 'natural_color', 'realistic_colors'])  
lcl = glbl.resample('euro4', resampler='bilinear', cache_dir='/tmp')
```

pyresample/kd_tree.py
```diff
@@ -926,11 +930,11 @@ def query_no_distance(target_lons, target_lats, valid_output_index,
     if index_array.ndim == 1:
         index_array = index_array[:, None]
 
-    # KDTree query returns out-of-bounds neighbors as `len(arr)`
+    # KDTree query returns out-of-bounds neighbours as `len(arr)`
```
djhoese: @pnuu We are trying to stick with US English spelling of things now. Your change makes this the UK English. Please undo these changes.

pnuu: That will break backwards compatibility, `neighbours` is used in the method and function names.

djhoese: True, but not sure comments needed to be changed. I'd be willing to deprecate the old neighbours methods over time (like until a 2.0 release).

pnuu:
I'm closing this as obsolete. All the other changes made after this was created make it hard to directly update. I might revisit this in a new PR started from more up-to-date starting point.

PR #167, opened 2019-03-11
Fix yaml area loading when projection is latlong (units degrees)

djhoese:
YAML parsing was assuming meters for units even for latlong projections.

@pnuu @wroberts4 Can you verify that this fixes the issue brought up on slack.


djhoese:
Ok, I made this PR from pycharm. Starting to figure out how it works...hopefully this is the last time I make a branch on the main repository.

PR #192, opened 2019-05-13
Add bucket resampling

pnuu:
This PR adds few methods to do "drop-in-a-bucket" resampling for calculating

- averages (average of all the values hitting a bin)
- hit-counts (number of values hitting a bin)
- fractions of different categorical values (e.g. water fraction of a bin)

when aggregating high resolution data to a lower resolution grid.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -26,6 +26,7 @@ when aggregating data to coarser scale grids.
 """
 
 import dask.array as da
+import xarray as xr
 import numpy as np
 from pyresample._spatial_mp import Proj
 
@@ -49,6 +50,9 @@ def get_bucket_indices(adef, lons, lats):
     """Get projection indices"""
 
     prj = Proj(adef.proj_dict)
+
+    lons = lons.ravel()
+    lats = lats.ravel()
 
     # Create output grid coordinates in projection units
     x_res = (adef.area_extent[2] - adef.area_extent[0]) / adef.width
@@ -79,27 +83,42 @@ def get_bucket_indices(adef, lons, lats):
     return x_idxs, y_idxs
 
 
-def get_sample_from_bucket_indices(data, x_idxs, y_idxs, target_shape):
+def get_sum_from_bucket_indices(data, x_idxs, y_idxs, target_shape):
     """Resample the given data with drop-in-a-bucket resampling.  Return
     the sum and counts for each bin as two Numpy arrays."""
 
-    sums = np.zeros(target_shape)
-    counts = np.zeros(target_shape)
-    data = np.ravel(data)
+    if isinstance(data, xr.DataArray):
+        data = data.data
+    data = data.ravel()
+    # Remove NaN values from the data when used as weights
+    weights = da.where(np.isnan(data), 0, data)
 
-    idxs = (x_idxs > 0) & (y_idxs > 0) & np.isfinite(data)
-    x_idxs = x_idxs[idxs]
-    y_idxs = y_idxs[idxs]
-    data = data[idxs]
+    # Convert X- and Y-indices to raveled indexing
+    idxs = y_idxs * target_shape[1] + x_idxs
+    # idxs = idxs.ravel()
 
-    # Place the data to proper locations
-    for i in range(x_idxs.size):
-        x_idx = x_idxs[i]
-        y_idx = y_idxs[i]
-        sums[y_idx, x_idx] += data[i]
-        counts[y_idx, x_idx] += 1
+    out_size = target_shape[0] * target_shape[1]
 
-    return sums, counts
+    # Calculate the sum of the data falling to each bin
+    sums, _ = da.histogram(idxs, bins=out_size, range=(0, out_size),
+                           weights=weights, density=False)
+    return sums.reshape(target_shape)
+
+
+def get_count_from_bucket_indices(x_idxs, y_idxs, target_shape):
+    """Resample the given data with drop-in-a-bucket resampling.  Return
+    the sum and counts for each bin as two Numpy arrays."""
+
+    # Convert X- and Y-indices to raveled index
+    idxs = y_idxs * target_shape[1] + x_idxs
+    # idxs = idxs.ravel()
+
+    out_size = target_shape[0] * target_shape[1]
+
+    # Calculate the sum of the data falling to each bin
+    counts, _ = da.histogram(idxs, bins=out_size, range=(0, out_size))
+
+    return counts.reshape(target_shape)
 
 
 def resample_bucket_average(adef, data, lats, lons,
@@ -107,10 +126,11 @@ def resample_bucket_average(adef, data, lats, lons,
     """Calculate bin-averages using bucket resampling."""
     if x_idxs is None or y_idxs is None:
         x_idxs, y_idxs = get_bucket_indices(adef, lons, lats)
-    sums, counts = get_sample_from_bucket_indices(data, x_idxs, y_idxs,
-                                                  adef.shape)
+    sums = get_sum_from_bucket_indices(data, x_idxs, y_idxs, adef.shape)
+    counts = get_count_from_bucket_indices(x_idxs, y_idxs, adef.shape)
+
     average = sums / counts
-    average = np.where(counts == 0, fill_value, average)
+    average = da.where(counts == 0, fill_value, average)
 
     return average
 
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -41,40 +41,40 @@ class BucketResampler(object):
 
     .. code-block:: python
 
-        >>> from pyresample.bucket import BucketResampler
-        >>> from satpy import Scene
-        >>> from satpy.resample import get_area_def
-        >>> fname = "hrpt_noaa19_20170519_1214_42635.l1b"
-        >>> glbl = Scene(filenames=[fname])
-        >>> glbl.load(['4'])
-        >>> data = glbl['4']
-        >>> lons, lats = data.area.get_lonlats()
-        >>> target_area = get_area_def('euro4')
+        from pyresample.bucket import BucketResampler
+        from satpy import Scene
+        from satpy.resample import get_area_def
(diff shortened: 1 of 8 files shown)


pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+        x_idxs, y_idxs = get_bucket_indices(adef, lons, lats)
+    results = {}
+    counts = get_count_from_bucket_indices(x_idxs, y_idxs, adef.shape)
+    counts = counts.astype(float)
+    for cat in categories:
+        cat_data = da.where(data == cat, 1.0, 0.0)
+
+        sums = get_sum_from_bucket_indices(cat_data, x_idxs, y_idxs, adef.shape)
+        results[cat] = sums.astype(float) / counts
+
```
sfinkens: Fill results with fill value where count is zero?

pnuu: We discussed this here, and thought that we should leave them as `nan`. But then the `fill_value` kwarg should be removed. I think it's better to allow the user to use a fill value if so wishes, so I'll add that.

sfinkens: :+1:

pnuu:
Hmm. I think I might need to rethink `resample_bucket_average()`, there's no point in requiring `lons` and `lats` if the indices are already calculated and passed via the kwargs.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+        Bin-wise count of hits for each target grid location
+    """
+
+    # Convert X- and Y-indices to raveled index
+    idxs = y_idxs * target_shape[1] + x_idxs
+
+    out_size = target_shape[0] * target_shape[1]
+
+    # Calculate the sum of the data falling to each bin
+    counts, _ = da.histogram(idxs, bins=out_size, range=(0, out_size))
```
mraspaud: Can `histogram` be replaced with `bincount` in this case ?

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,305 @@
...
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Code for resampling using bucket resampling.
+
+Bucket resampling is useful for calculating averages and hit-counts
+when aggregating data to coarser scale grids.
+"""
```
mraspaud: Can you add a couple of examples ?

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,305 @@
+# pyresample, Resampling of remote sensing image data in python
+#
+# Copyright (C) 2019  Pytroll developers
+#
+# Author(s):
+#
+#   Panu Lahtinen <panu.lahtinen@fmi.fi>
```
djhoese: I'd like if we stopped putting individual authors in the copyright header and changed "Pytroll developers" to "Pyresample developers".

setup.py
```diff
@@ -32,7 +32,7 @@
                   'rasterio': ['rasterio'],
                   'dask': ['dask>=0.16.1']}
 
-test_requires = ['rasterio', 'dask', 'xarray', 'cartopy', 'pillow', 'matplotlib', 'scipy']
+test_requires = ['rasterio', 'dask', 'xarray', 'cartopy', 'pillow', 'matplotlib', 'scipy', 'pyproj', 'numpy']
```
djhoese: `pyproj` and `numpy` are both hard requirements on the package as a whole so these shouldn't be needed here.

pyresample/bucket/__init__.py
pnuu: Doesn't look like it. In `get_bucket_indices()` I need to mark invalid indices with `-1`, and that doesn't work with `bincount()`. And as the indices need to be integers, using `np.nan` is out... But `histogram()` works :grin:

pyresample/bucket/__init__.py
pnuu: Ok. Just took the header from some other file.

setup.py
pnuu: This is always confusing to me. I always assume that the requirements are separate...

pyresample/bucket/__init__.py
pnuu: Examples added.

pnuu:
I'll convert this in to a class, which should make using a bit easier and less verbose.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,276 @@
...
+        weights = da.where(np.isnan(data), 0, data)
+
+        # Calculate the sum of the data falling to each bin
+        out_size = self.target_area.size
+        sums, _ = da.histogram(self.idxs, bins=out_size, range=(0, out_size),
+                               weights=weights, density=False)
+        return sums.reshape(self.target_area.shape)
+
+    def get_count(self):
+        """Count the number of occurances for each bin using drop-in-a-bucket
```
sfinkens: Typo: "occurrences"

pnuu: Typos fixed.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,291 @@
+# pyresample, Resampling of remote sensing image data in python
+#
+# Copyright (C) 2019  Pytroll developers
```
djhoese: `Pyresample developers`

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+import dask.array as da
+import xarray as xr
+import numpy as np
+from pyresample._spatial_mp import Proj
+
+
+class BucketResampler(object):
+
+    """Bucket resampling is useful for calculating averages and hit-counts
+    when aggregating data to coarser scale grids.
```
djhoese: The first line of a docstring should be one sentence and be one line long.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+        self.prj = Proj(self.target_area.proj_dict)
+        self.x_idxs = None
+        self.y_idxs = None
+        self.idxs = None
+        self._get_indices()
+        self.counts = None
+
+    def _get_proj_coordinates(self, lons, lats, x_res, y_res):
+        """Calculate projection coordinates and round them to the closest
+        resolution unit.
```
djhoese: Same here for single line docstring "subject" line.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+        Returns
+        -------
+        average : Dask array
+            Binned and averaged data.
+        """
+        sums = self.get_sum(data, mask_all_nan=mask_all_nan)
+        counts = self.get_count()
+
+        average = sums / counts
+        idxs = (counts == 0) | np.isnan(sums)
```
djhoese: These aren't indexes but ok.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,291 @@
...
+    def __init__(self, target_area, source_lons, source_lats):
+
+        self.target_area = target_area
+        self.source_lons = source_lons
+        self.source_lats = source_lats
+        self.prj = Proj(self.target_area.proj_dict)
+        self.x_idxs = None
+        self.y_idxs = None
+        self.idxs = None
+        self._get_indices()
```
djhoese: This (`_get_indices`) is a lot of logic to happen during init method, but I can see how it would be difficult to change it now.

pyresample/test/utils.py
```diff
@@ -178,3 +178,22 @@ def create_test_latitude(start, stop, shape, twist_factor=0.0, dtype=np.float32)
     lat_array = np.repeat(lat_col, shape[1], axis=1)
     lat_array += twist_array
     return lat_array
+
+
+class CustomScheduler(object):
+    """Custom dask scheduler that raises an exception if dask is computed
+    too many times."""
```
djhoese: Can you put this on one line too

djhoese:
We talked about this a bit in slack, but what if the fractions resampling returned a multi-dimensional `DataArray` object with a `category` dimension and a `my_data_arr.coords['category']` with the integer value of the category? I'm not sure what other attributes you would want to copy from the original data, possibly all of them.

If it was done this way then the Satpy interface could provide access to the fraction resampling too.

pnuu:
I decided not to use Xarray on the Pyresample side. I prefer to have plain dask/numpy arrays here so that people using the functionality directly don't need to wrap their data in `Xarray.DataArray`.

pyresample/bucket/__init__.py
pnuu: One workaround would be to call `self._get_indices()` in the resampler method calls. `if self.idxs is None: self._get_indices()`.

pyresample/bucket/__init__.py
```diff
@@ -0,0 +1,314 @@
...
+    Bucket resampling is useful for calculating averages and hit-counts
+    when aggregating data to coarser scale grids.
+
+    Below are examples how to use the resampler.
+
+    Read data using Satpy.  The resampling can also be done (apart from
+    fractions) directly from Satpy, but this demonstrates the direct
+    low-level usage.
+
+    .. code-block:: python
```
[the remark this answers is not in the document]
pnuu: Removing the `>>> ` seems to work. And now the class is actually in the documents.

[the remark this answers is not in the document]
djhoese: 🤦‍♂ Oops I see how poorly worded this is. I meant trying to remove `.. code-block:: python` because it would read better in a non-sphinx context.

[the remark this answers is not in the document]
pnuu: Ha, actually I first read the way you meant, but then went the short way and batch-removed `>>> `. Looking at the docs, all the other examples are using `>>> ` so went back to use that and removed the code blocks directive.

mraspaud:
Are we waiting for something or can this be merged ? @pnuu ?

pnuu:
Everything is ready on my part :-)

PR #195, opened 2019-05-30
Update docs for create_area_def and improve AreaDefinition property consistency

wroberts4:
I updated the imports and API for create_area_def docs.

Secondly, it would be very useful if the terminology between DynamicAreaDefinition and AreaDefinition were compatible; DynamicAreaDefinition uses resolution and AreaDefinition uses pixel_size_x/pixel_size_y, DynamicAreaDefinition uses width/height and AreaDefinition uses shape. This has made confusing code when trying to get properties from both DynamicAreaDefinition and AreaDefinition. It doesn't matter to me which terms we use as long as there is at least one shared term(s) for each property. ie not adding pixel_size_x/pixel_size_y to DynamicAreaDefinition and instead just rely on resolution.
 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented


pyresample/geometry.py
```diff
@@ -769,13 +769,17 @@ def __init__(self, area_id=None, description=None, projection=None,
...
+        self.width = width
+        self.height = height
         self.area_extent = area_extent
         self.optimize_projection = optimize_projection
         self.resolution = resolution
         self.rotation = rotation
 
+    @property
+    def shape(self):
+        return self.height, self.width
```
djhoese: Considering that height and width could be None I'm not sure I'm comfortable with this. I'll let @mraspaud weigh in though.

wroberts4: Yeah, I can change it to make shape None; I just added a description, but I just want consistency in properties between DynamicAreaDefinition and AreaDefinition

djhoese: I would rather not having the property at all, although consistency is nice.

PR #204, opened 2019-08-12
Add Example for Regular Lat-Lon Grid

sfinkens:
After introducing a couple of students/interns to pyresample I found that it would be useful to have an example explaining how to get the area definition for a given regular lat-lon grid.

I also separated the are definition examples into separate code blocks to improve the readability.

 - [ ] Tests passed 


djhoese:
Technically `eqc` is not a "lat/lon" grid, right? It is in meters. A lat/lon grid would be `+proj=latlong`.

sfinkens:
Yes, good point. Changed it to `latlong`.

sfinkens:
Hm, but now the example doesn't fit the "...even if the projection space is defined in meters" anymore. I'll just add an extra sentence for that.

PR #205, opened 2019-08-12
Exclude NaNs from Bucket Average

sfinkens:
Currently the bucket average counts NaNs in the data as valid observations. For example values [1, 2, 3, np.nan] are averaged to (1+2+3) / 4 = 1.5 instead of (1+2+3) / 3 = 2. Here is a snippet to reproduce:

```python
import pyresample.bucket
import numpy as np
import dask.array as da

area_def = create_area_def('my_area',
                           {'proj': 'eqc', 'lon_0': 0},
                           center=(0, 0),
                           width=2, height=2,
                           resolution=10,
                           units='degrees')

lons = da.from_array([[0, 0], [0, 0]], chunks=1)
lats = da.from_array([[0, 0], [0, 0]], chunks=1)
data = da.from_array([[1, 2], [3, np.nan]], chunks=1)

resampler = pyresample.bucket.BucketResampler(target_area=area_def,
                                              source_lons=lons,
                                              source_lats=lats)
avg = resampler.get_average(data=data, mask_all_nan=True, fill_value=np.nan).compute()
print(avg)
```

This PR attempts to exclude NaNs from the bucket average.

 - [X] Tests added 
 - [x] Tests passed 

### DIFF
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -234,7 +234,8 @@ class BucketResampler(object):
         LOG.info("Get average value for each location")
 
         sums = self.get_sum(data, mask_all_nan=mask_all_nan)
-        counts = self.get_count()
+        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int),
+                              mask_all_nan=False)
 
         average = sums / counts
         mask = (counts == 0) | np.isnan(sums)
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -237,9 +237,8 @@ class BucketResampler(object):
         counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int),
                               mask_all_nan=False)
 
-        average = sums / counts
-        mask = (counts == 0) | np.isnan(sums)
-        average = da.where(mask, fill_value, average)
+        average = sums / da.where(counts == 0, np.nan, counts)
+        average = da.where(np.isnan(average), fill_value, average)
 
         return average
 
--- a/pyresample/test/test_bucket.py
+++ b/pyresample/test/test_bucket.py
@@ -166,10 +166,10 @@ class Test(unittest.TestCase):
         with dask.config.set(scheduler=CustomScheduler(max_computes=0)):
             result = self.resampler.get_average(data, mask_all_nan=True)
         self.assertTrue(np.all(np.isnan(result)))
-        # By default all-NaN bins have a value of 0.0
+        # By default all-NaN bins have a value of NaN
         with dask.config.set(scheduler=CustomScheduler(max_computes=0)):
             result = self.resampler.get_average(data)
-        self.assertEqual(np.nanmax(result), 0.0)
+        self.assertTrue(np.all(np.isnan(result)))
 
     def test_resample_bucket_fractions(self):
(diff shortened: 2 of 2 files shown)


pyresample/bucket/__init__.py
```diff
@@ -234,7 +234,8 @@ def get_average(self, data, fill_value=np.nan, mask_all_nan=False):
         LOG.info("Get average value for each location")
 
         sums = self.get_sum(data, mask_all_nan=mask_all_nan)
-        counts = self.get_count()
+        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int),
+                              mask_all_nan=False)
 
         average = sums / counts
```
sfinkens: Shall we mask zero counts before division in order to avoid dask warnings?

pyresample/test/test_bucket.py
```diff
@@ -166,7 +173,8 @@ def test_get_average(self):
```
sfinkens: The last test now fails. Is it obsolete now?

pyresample/bucket/__init__.py
pnuu: Yes, that could make the output cleaner for the user.

pyresample/test/test_bucket.py
pnuu: Yes, I think `NaN` is a better default result. So maybe change the test on line 172 to `self.assertTrue(np.all(result))`.

sfinkens: I replaced it with `self.assertTrue(np.all(np.isnan(result)))` to make it more clear

pyresample/bucket/__init__.py
sfinkens: Done

pyresample/test/test_bucket.py
```diff
@@ -139,7 +144,8 @@ def test_get_count(self):
 
     def test_get_average(self):
         """Test averaging bucket resampling."""
-        data = da.from_array(np.array([[2., 4.], [2., 2.]]))
+        data = da.from_array(np.array([[2., 4.], [3., np.nan]]),
```
sfinkens: I already added a NaN to the data here, so that should cover the count issue.

pnuu: Oh, right! Missed that when browsing the changed rows.

pyresample/bucket/__init__.py
```diff
@@ -234,11 +234,11 @@ def get_average(self, data, fill_value=np.nan, mask_all_nan=False):
         LOG.info("Get average value for each location")
 
         sums = self.get_sum(data, mask_all_nan=mask_all_nan)
-        counts = self.get_count()
+        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int),
+                              mask_all_nan=False)
```
djhoese: Should this type of logic be something that can be made a kwarg in `get_count`? Could this functionality be defaulted the way you want it, but be controlled by a kwarg?

sfinkens: From our slack conversation: For buckets which we know will end up empty we have such a kwarg (``mask_all_nan``). But for the average-counts I wouldn't do that because the results are wrong. The only use case I can think of is speeding up the computation if the user knows that there are no NaNs in the data. Then ``get_average`` could directly use ``get_count``.

PR #207, opened 2019-08-16
Make output index tiling in bilinear interpolation work with dask

pnuu:
This PR makes output index tiling in bilinear interpolation more dask friendly. And actually work.

The existing version doesn't work with dask arrays. The `np.tile()` call requires an integer, when plain Numpy arrays can be tiled with a tuple argument.

Also includes a ton of flake8 fixes forced by pre-commit.

 - [x] Tests passed 


pyresample/bilinear/xarr.py
```diff
@@ -28,7 +30,8 @@ def __init__(self,
                  nprocs=1,
                  segments=None):
         """
-        Parameters
+        Parameters.
+
```
djhoese: Extra blank line here after Parameters.

pyresample/bilinear/xarr.py
```diff
@@ -357,18 +360,21 @@ def _mask_coordinates_dask(lons, lats):
...
-
     # Find four closest pixels around the target location
 
     # FIXME: how to daskify?
     # Tile output coordinates to same shape as neighbour info
     # Replacing with da.transpose and da.tile doesn't work
-    out_x_tile = np.transpose(np.tile(out_x, (neighbours, 1)))
-    out_y_tile = np.transpose(np.tile(out_y, (neighbours, 1)))
+    out_x_tile = np.reshape(np.tile(out_x, neighbours),
+                            (neighbours, out_x.size)).T
```
djhoese: Have you checked the dask source to see if this is intentionally not implemented (meaning, non-trivial)? Or did someone just not implement it yet and you could?

pyresample/bilinear/xarr.py
pnuu: Yes, that's what pre-commit and flake8 wanted. Wouldn't let me commit without it. The first line (the shoet description of the function/method) should end in a full stop, followed by an empty line.

pyresample/bilinear/xarr.py
pnuu: No, didn't have time, just made it to work :grin:

pyresample/bilinear/xarr.py
djhoese: Oh it's because you don't have a description of the method. It should be:

```
"""Initialize resampler.

Parameters
-----------

```

What you have now is "Parameters." is the "subject line" of the method.

pnuu: Yeah. Had to do a quick modification to quite many places so could push the real changes before the day ended. I'll update that and check also the other bits on monday.

pyresample/bilinear/xarr.py
pnuu: The source doesn't mention if this omission is by design or just a missing feature (https://github.com/dask/dask/blob/master/dask/array/creation.py#L802), but there's an issue from 2017 about this feature with new comments from June this year with a suggested solution: https://github.com/dask/dask/issues/2447

djhoese: We probably mentioned this on slack, but do you remember what the error was or what was wrong when using the dask version?

pnuu: Dask raises a `NotImplementedError` when trying to `tile` with a `tuple`. The call to `np.tile()` calls the dask version in the background for dask arrays and causes the error to be raised.

pnuu:
Indeed. All the bilinear resampler tests seem to be for non-xarray/dask version. Blargh.

PR #208, opened 2019-08-27
Bugfix bilinear resampler masking

pnuu:
This PR fixes the masking in bilinear interpolation.

The existing version incorrectly uses area definition of the output grid when masking input data.

 - [x] Tests passed 

Closes https://github.com/pytroll/satpy/issues/735


pnuu:
With this and https://github.com/pytroll/pyresample/pull/207 bilinear interpolation will again work properly, also via Satpy.

PR #212, opened 2019-09-06
Use slices in bilinear resampler

pnuu:
This PR changes the bilinear resampler to return data and mask slices from `get_bil_info()` method instead of the `valid_input_index` and `index_array`. The current version uses these latter to compute the slices every time, so some time will be saved.

Also includes "few" flake8 fixes in `geometry.py` forced by `pre-commit`.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/bilinear/xarr.py
+++ b/pyresample/bilinear/xarr.py
@@ -16,6 +16,14 @@ from pyresample._spatial_mp import Proj
 from pykdtree.kdtree import KDTree
 from pyresample import data_reduce, geometry, CHUNK_SIZE
 
+CACHE_INDICES = ['bilinear_s',
+                 'bilinear_t',
+                 'slices_x',
+                 'slices_y',
+                 'mask_slices',
+                 'out_coords_x',
+                 'out_coords_y']
+
 
 class XArrayResamplerBilinear(object):
     """Bilinear interpolation using XArray."""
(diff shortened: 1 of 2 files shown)


pnuu:
Note: this will break bilinear interpolation for Satpy until https://github.com/pytroll/satpy/pull/897 is ready.

I'm not yet sure if, and how, to document this. The default bilinear resampler is the Numpy based version, and having two *almost* identical versions documented might be quite confusing.

Appveyor and Travis fail in a place where I have `dask` arrays locally (in linux), but they have `np.ndarray`, not sure what's happening there. The dask versions are the same in all three, 2.3.0. I'll try something.

And I'm going to ignore the "errors" Stickler reported.

pnuu:
That seemed to work.

For reference some timings (OMP_NUM_THREADS=1 DASK_NUM_WORKERS=4 PYTROLL_CHUNK_SIZE=1024, 2 composites with HRV data, 6 composites without HRV data, `generate=False`):
**current `pyresample` and `satpy` `master` branches**
- first run: 146.2 s
- cached run: 43.3 s
**PR branches of `pyresample` and `satpy`:
- first run: 134.0 s
- cached run: 27.2 s

PR #226, opened 2019-11-13
Fix inconsistent radius

loerum:
When spherical.SphPolygon is returning an instance of a SphPolygon (f.x in _bool_oper) it should pass the radius of it self. So area computations are consistent.


loerum:
Hi David, I have added a test to test/test_spherical.py

PR #227, opened 2019-11-16
Fix EWA resampling hanging when geolocation had a lot of NaNs

djhoese:
This was originally discovered by @spruceboy with the CSPP Polar2Grid project. I'm porting my fix for that implementation of EWA here. The P2G PR is https://github.com/ssec/polar2grid/pull/238.

For some reason the data files are being produced with the first 8 or so rows of geolocation (longitude and latitude) being all missing. In Polar2Grid these missing values are represented as NaNs. The calculations performed in fornav with the geolocation (actually the columns and rows from the ll2cr step) end up with a lot of the algorithm parameters being NaN. This causes EWA to either not fall in to or fall in to if satements that normally wouldn't be done causing processing to take an extremely long time (~1-5 rows per second compared to 100s or 1000s of rows per second).

This PR tries to address this by resetting the NaN parameters to the maximum distance parameter to limit how many pixels are processed when geolocation is NaN.

 - [x] Tests added 
 - [x] Tests passed 


djhoese:
This PR can wait to be merged unless we want to rush a release out. There is a new issue popping up in Polar2Grid that could still be related to these NaN values so more adjustments could be made here or in the future.

PR #235, opened 2019-11-29
Fix cartesian coordinate calculations when lon/lats are integers

### LINKED ISSUE #233: get_neighbour_info can not handle SwathDefinitions with lat lon of type integers

```python
import numpy as np
from pyresample.kd_tree import  get_neighbour_info
from pyresample.geometry import SwathDefinition

lon = np.array([0, 10, 25])
source_def  = SwathDefinition(*(lon, lon))
target_def = SwathDefinition(*(lon, lon))
valid_in, valid_out, indices_int, distances = get_neighbour_info(source_def, target_def, 1000, neighbours=1)
lon = np.array([0, 10, 25]).astype(np.float64)
source_def  = SwathDefinition(*(lon, lon))
target_def = SwathDefinition(*(lon, lon))
valid_in, valid_out, indices_float, distances = get_neighbour_info(source_def, target_def, 1000, neighbours=1)
print(indices_float)
print(indices_int)

```
#### Problem description

When using integer lat/lon get_neighbour_info returns only zeros. 

#### Expected Output

I expected indices_float and indices_int to have the same values ([0, 1, 2]).

>>> print(indices_float)
[0 1 2]
>>> print(indices_int)
[0 0 0]

pyresample version 1.13.2
python version 3.6.7

I am not sure if it is SwathDefinition that should cast lat/lon to floats. Or warn that they are integers. Or if get_neighbour_info should handle also integers.

ninahakansson:
The main problem was the casting of pi/180 to the same type as the latitude-longitude data. In the case of integer data it is cast to zero.  Current solution uses np.deg2lats in case the pi/180 was cast to zero. 

 - [x] Closes #233 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/_spatial_mp.py
+++ b/pyresample/_spatial_mp.py
@@ -188,8 +188,8 @@ class Cartesian(object):
     def transform_lonlats(self, lons, lats):
 
         coords = np.zeros((lons.size, 3), dtype=lons.dtype)
-        deg2rad = lons.dtype.type(np.pi / 180)
-        if ne and deg2rad > 0:
+        if ne:
+            deg2rad = np.pi / 180
             coords[:, 0] = ne.evaluate("R*cos(lats*deg2rad)*cos(lons*deg2rad)")
             coords[:, 1] = ne.evaluate("R*cos(lats*deg2rad)*sin(lons*deg2rad)")
             coords[:, 2] = ne.evaluate("R*sin(lats*deg2rad)")
--- a/pyresample/_spatial_mp.py
+++ b/pyresample/_spatial_mp.py
@@ -189,15 +189,16 @@ class Cartesian(object):
 
         coords = np.zeros((lons.size, 3), dtype=lons.dtype)
         deg2rad = lons.dtype.type(np.pi / 180)
-        if ne:
+        if ne and deg2rad > 0:
             coords[:, 0] = ne.evaluate("R*cos(lats*deg2rad)*cos(lons*deg2rad)")
             coords[:, 1] = ne.evaluate("R*cos(lats*deg2rad)*sin(lons*deg2rad)")
             coords[:, 2] = ne.evaluate("R*sin(lats*deg2rad)")
         else:
-            coords[:, 0] = R * np.cos(lats * deg2rad) * np.cos(lons * deg2rad)
-            coords[:, 1] = R * np.cos(lats * deg2rad) * np.sin(lons * deg2rad)
-            coords[:, 2] = R * np.sin(lats * deg2rad)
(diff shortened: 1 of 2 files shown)


ninahakansson:
Sure I can update to not use the lons.dtype! But I am just thinking maybe it was there for a reason? Is it good to return data as type float32 if that was the original type? But my code now would anyway have destroyed that ...

Should I do as you suggest and remove lons.dtype, or should I change the code to:

if deg2read equals zero: use np.deg2rad
elif ne: as before
else: as before

djhoese:
So after re-reviewing the code, I don't think any of the current solutions would work for you. If you give it an integer array for lons and lats then the `coords` array will be an integer, right? Is that really what you want? We are doing these calculations to get new coordinates and even though the degrees coordinates may have been integers that doesn't mean the new coordinates will be nice round numbers.

I'm kind of thinking things should be cast to float if needed or that it should be an error to pass integer types. If I'm missing something, let me know.

ninahakansson:
I think it is perfektly fine for integer to be cast to floats. I do not need them to be integers after transformation. As you say it would not make sense. My question in the comment was just if we want float32 to stay float32. I thought (wrongly) that np.deg2rad would always return float64 ... I will update the PR according to your first comment.

djhoese:
I think worrying about preserving the float32 is a good idea, but I think in this case since we are always writing values to `coords` which was declared as the same dtype as `lons` it isn't a problem.

You could do `if np.issubdtype(lons.dtype, np.integer): lons = lons.astype(np.float)` or something like that. That needs testing though. @mraspaud other ideas?

pyresample/_spatial_mp.py
```diff
@@ -186,19 +186,22 @@ def __init__(self, *args, **kwargs):
...
 
     def transform_lonlats(self, lons, lats):
-
+        """Transform longitudes and latitues to cartesian coordinates."""
+        if np.issubdtype(lons.dtype, np.integer):
+            lons = lons.astype(np.float)
         coords = np.zeros((lons.size, 3), dtype=lons.dtype)
-        deg2rad = lons.dtype.type(np.pi / 180)
         if ne:
+            deg2rad = np.pi / 180  # noqa: F841
```
djhoese: Just curious, what does flake8 have a problem with on this line?

ninahakansson: That we define deg2rad but then never use it .... :)

djhoese: Ah of course. thanks.

djhoese:
Thanks again for this. The only failing tests here are unrelated and have been fixed in master now. Merging...

PR #243, opened 2019-12-10
Respect dtype when get_lonlats provide dask array

### LINKED ISSUE #242: AreaDefinition.get_lonlats ignores dtype option

```python
from satpy import Scene
from glob import glob
srcdir = "/data/gholl/cache/MTG"
fciscene = Scene(sensor="fci", reader=["fci_l1c_fdhsi"],
                 filenames=glob("/data/gholl/cache/MTG/*BODY*.nc"))
fciscene.load(["vis_09"])
(lon, lat) = fciscene["vis_09"].area.get_lonlats(chunks=1024, dtype="f4")
print(lon.dtype, lat.dtype)
```
#### Problem description

According to [the documentation](https://pyresample.readthedocs.io/en/latest/API.html#pyresample.geometry.AreaDefinition.get_lonlats), `pyresample.geometry.AreaDefinition.get_lonlats` has a keyword argument `dtype`:

> dtype (numpy.dtype, optional) – Data type of the returned arrays

However, even if I pass `dtype="f4"` (single precision), the returned dask array is double precision.

#### Expected Output

I expect the script above to print

```
float32 float32
```

In reality, the script above prints

```
float64 float64
```

Python 3.7.3, pyresample 1.13.2, satpy 0.18.2.dev131+ga08c6202

gerritholl:
Respect dtype when get_lonlats provide dask array.

 - [x] Closes #242  
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1822,8 +1822,7 @@ class AreaDefinition(BaseDefinition):
             from dask.array import map_blocks
             res = map_blocks(invproj, target_x, target_y,
                              chunks=(target_x.chunks[0], target_x.chunks[1], 2),
-                             new_axis=[2], proj_dict=proj_def,
-                             dtype=dtype)
+                             new_axis=[2], proj_dict=proj_def).astype(dtype)
             return res[:, :, 0], res[:, :, 1]
 
         if nprocs > 1:
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1822,7 +1822,8 @@ class AreaDefinition(BaseDefinition):
             from dask.array import map_blocks
             res = map_blocks(invproj, target_x, target_y,
                              chunks=(target_x.chunks[0], target_x.chunks[1], 2),
-                             new_axis=[2], proj_dict=proj_def)
+                             new_axis=[2], proj_dict=proj_def,
+                             dtype=dtype)
             return res[:, :, 0], res[:, :, 1]
 
         if nprocs > 1:
--- a/pyresample/test/test_geometry.py
+++ b/pyresample/test/test_geometry.py
@@ -1191,6 +1191,37 @@ class Test(unittest.TestCase):
                                                               area_extent[3]))
         self.assertEqual(reduced_area.shape, (928, 928))
 
+    def test_get_lonlats_options(self):
+        """Test that lotlat options are respected
+        """
+        area_def = geometry.AreaDefinition('areaD', 'Europe (3km, HRV, VTC)', 'areaD',
+                                           {'a': '6378144.0',
+                                            'b': '6356759.0',
+                                            'lat_0': '50.00',
+                                            'lat_ts': '50.00',
+                                            'lon_0': '8.00',
+                                            'proj': 'stere'},
+                                           800,
+                                           800,
+                                           [-1370912.72,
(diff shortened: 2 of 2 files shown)


pyresample/geometry.py
```diff
@@ -1822,7 +1822,8 @@ def get_lonlats(self, nprocs=None, data_slice=None, cache=False, dtype=None, chu
             from dask.array import map_blocks
             res = map_blocks(invproj, target_x, target_y,
                              chunks=(target_x.chunks[0], target_x.chunks[1], 2),
-                             new_axis=[2], proj_dict=proj_def)
+                             new_axis=[2], proj_dict=proj_def,
+                             dtype=dtype)
```
djhoese: What version of pyproj did you test this with? I noticed this recently too, but dask should be checking the dtype for us automatically. Are you seeing something different?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Version 1.13.3 (20xx-xx-xx)
+
+### Issues Closed
+
+* [Issue 242](https://github.com/pytroll/pyresample/issues/242) - Problem with dtype of dask arrays returned by AreaDefinition.get_lonlats ([PR 242](https://github.com/pytroll/pyresample/pull/243)
+
```
djhoese: Where did this changelog entry come from?

pyresample/geometry.py
gerritholl: See #242 and see it failing in Travis CI with [this commit](https://travis-ci.org/pytroll/pyresample/builds/623098241?utm_source=github_status&utm_medium=notification), as I committed and pushed the test before I did so for the fix.

djhoese:
So I just read your PR and understand what you're doing here more. If I'm not mistaken, pyproj is always returning a float64. I'm not sure we should depend on dask doing the casting for us from what our function returns and what dtype we asked for. What do you think about doing `.astype(dtype)` after the map_blocks call?

gerritholl:
In the non-dask case the casting happens in the lines

https://github.com/pytroll/pyresample/blob/ab6289cac42f5ae0b8a460fc16b23027d2b8566e/pyresample/geometry.py#L1835-L1836

but in the dask-case the dtype option was ignored prior to this PR.

I don't know about any semantical differences between passing `dtype=...` to `map_blocks` or calling `.astype(...)` afterward.  Doesn't `.astype` also become a dask ufunc, such that the casting is done by dask either way?  What is the advantage of calling `.astype(dtype)` after the `map_blocks` rather than passing `dtype=...` within the `map_blocks` call?

djhoese:
My understanding of `dtype` in `map_blocks` is not to tell dask how to convert the returned array, but rather to tell dask what to *expect* from the returned function. If pyproj is always returning float64 then `dtype` to `map_blocks` should always be float64. However, dask has a practice of passing dummy arrays to the mapped function to make an educated guess at what the returned dtype will be for all future calls.

So the use of `astype` is more explicit to me in that we are saying "we know we were probably returned a 64-bit float, but we want to make sure it is of type X".

My worry is that if dask developers decided that converting dtypes "magically" by passing a new `dtype=` is bad and then changed it, that pyresample would be broken in these cases.

PR #251, opened 2020-02-06
Fix bugs and use real data in the doc pages but hide code needed for …

adybbroe:
Fix bugs and use real data in the doc pages but hide code needed for doc testing

There was also one duplicated section which is now fixed.

The idea is that the user should be able more or less to generate the same plots by going from top to bottom in a notebook or ipython prompt.

Signed-off-by: Adam.Dybbroe <a000680@c21856.ad.smhi.se>

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Tests passed 
 - [x] Fully documented


adybbroe:
Should be ready now.
One thing which is lacking is some help on how to get the background blue marble ready for Cartopy. Didn't find a nice description/manual for this online when I was looking now. But locally I have this json file:
```
➜  raster more images.json 
{"__comment__": "JSON file specifying the image to use for a given type/name and resolution. Read in by cartopy.mpl.geoaxes.read_user_background_images.",
  "BM": {
    "__comment__": "Blue Marble Next Generation, July ",
    "__source__": "https://neo.sci.gsfc.nasa.gov/view.php?datasetId=BlueMarbleNG-TB",
    "__projection__": "PlateCarree",
    "low": "BM.jpeg",
    "high": "BMNG_hirez.png"},
    "pop": {
    "__comment__": "Population density from Nasa earth observations website",
    "__source__": "https://neo.sci.gsfc.nasa.gov/view.php?datasetId=SEDAC_POP",
    "__projection__": "PlateCarree",
    "high": "population.jpg"}
}
```
and a jpeg file `BM.jpeg` in the same directory. And that directory is pointed to by `CARTOPY_USER_BACKGROUNDS`

Any suggestions welcome!

docs/source/plot.rst
```diff
@@ -78,18 +96,23 @@ Assuming the file **areas.yaml** has the following area definition:
...
 
- >>> import numpy as np 
- >>> from pyresample import load_area, save_quicklook, SwathDefinition
- >>> from pyresample.kd_tree import resample_nearest
- >>> lons = np.zeros(1000)
- >>> lats = np.arange(-80, -90, -0.01)
- >>> tb37v = np.arange(1000)
- >>> area_def = load_area('areas.yaml', 'pc_world')
- >>> swath_def = SwathDefinition(lons, lats)
+ >>> area_def = load_area('areas.yaml', 'pc_world') # doctest: +SKIP
```
djhoese: Isn't this area load just overwriting the area_def created in the lines above?

adybbroe: Yes, right. But the stuff above is hidden, and does not show up on RTD. Therefore the additional line for RTD. The area is not defined and on Travis we don't have access to `areas.yaml` and certainly not in the home dir.

adybbroe:
I am also not pleased with it really. I am using this doctest-hide to help doctesting but hiding it for the user. 
I haven't found a way to skip a whole block when testing.

docs/source/plot.rst
djhoese: Ah, pyresample doesn't have a builtin areas :man_facepalming: Got it.

djhoese:
Looks like there are some import issues on the tests. @adybbroe Could you look at these?

adybbroe:
Closing and reopening the PR to try trigger a RTD build

adybbroe:
I think I am ready with this PR now, and now one can see the doc-pages being build on RTD for PRs.
Except that I here see the # doctest: +SKIP (I suppose they will be gone once merged?) I believe it is ready. Perhaps @TomLav @mraspaud @djhoese or someone else want to go through it quickly?

djhoese:
@mraspaud @adybbroe I think this should be good now. I made one important change to the code that could cause issues for some people: I remove the `matplotlib.use` line in the `save_quicklook` function. It was using a `warn` keyword argument that is being deprecated in matplotlib 3.3 and I personally don't think a function should be setting the matplotlib backend for the user. If this was a known problem and was needed in the past we can revisit it. Otherwise I'm going to assume modern matplotlib behaves better for this.

PR #255, opened 2020-02-24
Fix for python3 rpm-build

adybbroe:
Fix so that it is possible to make an rpm package for Python 3 (only)

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
I assume this means that RPMs can be made for CentOS versions where Python 3 is the default. Merging now with the assumption that this doesn't break something further down the line.

PR #265, opened 2020-04-03
Fix geocentric resolution favoring one area dimension over the other

djhoese:
This is a major fix for how nearest neighbor resampling works by default. In a previous PR (#225), I had added a `geocentric_resolution` method which could be used to come up with a best guess for the `radius_of_influence` of nearest neighbor resampling. However, all of my tests were based on square areas. While investigating a complaint by a coworker, I realized that `geocentric_resolution` actually depends on the size of the X and Y dimension.

The code I had did the following:

1. Create a horizontal midline and vertical midline of coordinates.
2. Calculate the geocentric distance along the points on those lines.
3. Take the histogram of all of those distances.
4. Take the largest bin (the one with the most points) and average the edge values.

The issue with this is that the histogram, and therefore the largest bin, is dependent on how many pixels go in to each midline. So if you have a non-square area, there is a good chance that the longer dimension is going to have the largest bin. This means you end up completely ignoring the other dimension which might actually have a larger resolution that needs to be accounted for to have an accurate `radius_of_influence`. So this PR changes it to perform step 3 and 4 for each midline, then take the maximum.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1997,7 +1997,27 @@ class AreaDefinition(BaseDefinition):
         return new_area
 
     def geocentric_resolution(self, ellps='WGS84', radius=None):
-        """Find best estimate for overall geocentric resolution."""
+        """Find best estimate for overall geocentric resolution.
+
+        This method is extremely important to the results of KDTree-based
+        resamplers like the nearest neighbor resampling. This is used to
+        determine how far the KDTree should be queried for valid pixels
+        before giving up (`radius_of_influence`). This method attempts to
+        make a best guess at what geocentric resolution (the units used by
+        the KDTree) represents the majority of an area.
+
+        To do this this method will:
+
+        1. Create a vertical mid-line and a horizontal mid-line.
+        2. Convert these coordinates to geocentric coordinates.
+        3. Compute the distance between points along these lines.
+        4. Take the histogram of each set of distances and find the
+           bin with the most points.
+        5. Take the average of the edges of that bin.
+        6. Return the maximum of the vertical and horizontal bin
+           edge averages.
+
+        """
         from pyproj import transform
         rows, cols = self.shape
         mid_row = rows // 2
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -2010,20 +2010,36 @@ class AreaDefinition(BaseDefinition):
             dst = Proj("+proj=cart +a={} +b={}".format(radius, radius))
         else:
             dst = Proj("+proj=cart +ellps={}".format(ellps))
+        # need some altitude, go with the surface (0)
         alt_x = np.zeros(x.size)
         alt_y = np.zeros(y.size)
+        # convert our midlines to (X, Y, Z) geocentric coordinates
         hor_xyz = np.stack(transform(src, dst, x, mid_row_y, alt_x), axis=1)
         vert_xyz = np.stack(transform(src, dst, mid_col_x, y, alt_y), axis=1)
+        # Find the distance in meters along our midlines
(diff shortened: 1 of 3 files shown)


pyresample/test/test_kd_tree.py
```diff
@@ -902,7 +902,7 @@ def test_nearest_area_2d_to_area_1n_no_roi(self):
         self.assertIsInstance(res.data, da.Array)
         res = res.values
         cross_sum = np.nansum(res)
-        expected = 32114793.0
+        expected = 87281406.0
```
djhoese, line 905: I realize that this looks fishy, but I tested it with real ABI data and I think it is accurate. The source area definition whose geocentric resolution is being used is 10x50 pixels (exactly what this PR is addressing). When I took that definition, moved it over to North America and tested it with Full Disk ABI data with this PR it fully covered the output. When I did it with the old resolution it had missing pixels.

djhoese:
The original code I used to test my coworkers use case:


```
import os
os.environ['OMP_NUM_THREADS'] = '1'
os.environ['DASK_NUM_WORKERS'] = '8'
from satpy import Scene
from glob import glob

area_def = create_area_def('way_south',
                        "+proj=lcc +datum=WGS84 +ellps=WGS84 +lat_0=-63.00000 +lat_1=-63.00000 +lon_0=-53.00000 +units=m +no_defs",
                        1200, 1000,
                        resolution=(500, -500),
                        upper_left_extent=(-299999.8122955704, 250000.29971955402)
                       )

scn = Scene(reader='abi_l1b', filenames=glob('/data/satellite/abi/2018253/*RadF*.nc'))
scn.load(['C02'])
new_scn = scn.resample(area_def)
new_scn['C02'].plot.imshow()
```

I then tried modifying it with 1200 columns and 5000 rows which I would consider an extreme case of a non-square area near the edge of a full disk and it did have some missing pixels on the edge but this is the edge of the disk with a really tall output area. I'm ok with that.

So is this PR perfect, no. Is it much better than it was, yes (IMO).

djhoese:
@peters77 and @simonrp84 you too are always really good for testing stuff. Would you mind installing this version of pyresample and see if your usual satpy resampling cases still produce valid images when you don't specify a `radius_of_influence` to the resample call with the 'nearest' resampler?

pnuu:
The default `radius_of_influence` seems to be still leaving gaps close to the edges: 
natural_color_20161128_110000

djhoese:
As mentioned on slack, I think this PR is good as-is for what it is trying to solve. I think my geocentric code handles the common case that users will run in to. I would consider the edge of the disk being a special case which I could add to this PR or another (mainly check if source is geostationary projection).

pyresample/geometry.py
```diff
@@ -2010,20 +2030,36 @@ def geocentric_resolution(self, ellps='WGS84', radius=None):
...
-            raise RuntimeError("Could not calculate geocentric resolution")
-        # return np.max(dist)  # alternative to histogram
+        # Get rid of any NaNs or infinite values
+        hor_dist = hor_dist[np.isfinite(hor_dist)]
+        vert_dist = vert_dist[np.isfinite(vert_dist)]
         # use the average of the largest histogram bin to avoid
-        # outliers and really large values
-        return np.mean(np.histogram_bin_edges(dist)[:2])
+        # outliers and really large values.
+        # Very useful near edge of disk geostationary areas.
```
pnuu, line 2047: The large values near the disk edges would be very *helpful*. But I guess that would require to inspect the intersection of source and target areas so that cases where "too high" radii are not desired aren't affected.

PR #277, opened 2020-05-08
Fix calculating area slices for flipped projections

pnuu:
This fixes a segmentation fault for Satpy data reduction when the source area is flipped and GEO data has space surrounding the full disk data. In this case `pykdtree` was passed coordinates that were `inf` which causes segmentation fault.

 - [x] Closes https://github.com/pytroll/pyresample/issues/274 
 - [x] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/pyresample/test/test_kd_tree.py
+++ b/pyresample/test/test_kd_tree.py
@@ -902,7 +902,7 @@ class TestXArrayResamplerNN(unittest.TestCase):
         self.assertIsInstance(res.data, da.Array)
         res = res.values
         cross_sum = np.nansum(res)
-        expected = 32114793.0
+        expected = 87281406.0
         self.assertEqual(cross_sum, expected)
 
         # pretend the resolutions can't be determined
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1997,7 +1997,27 @@ class AreaDefinition(BaseDefinition):
         return new_area
 
     def geocentric_resolution(self, ellps='WGS84', radius=None):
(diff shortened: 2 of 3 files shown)


gerritholl:
I tested this with my example and I no longer get a segmentation fault.  It successfully generates this image:

ir_123

Using:

```python
from satpy import Scene
from satpy.utils import debug_on
debug_on()
from glob import glob
sc = Scene(glob("/media/nas/x21308/2020_04_MTG_unofficial_Testdata/20130804_RC72/W_XX-EUMETSAT-Darmstadt,IMG+SAT,MTI1+FCI-1C-RRAD-FDHSI-FD--CHK-BODY--L2P-NC4E_C_EU*.nc"), reader="fci_l1c_fdhsi")
sc.load(["ir_123"])
nsc = sc.resample("fci_0deg_2km")
nsc["ir_123"].compute()
nsc.save_dataset("ir_123", "/tmp/ir_123.png")
```

pyresample/test/test_kd_tree.py
```diff
@@ -902,7 +902,7 @@ def test_nearest_area_2d_to_area_1n_no_roi(self):
         self.assertIsInstance(res.data, da.Array)
         res = res.values
         cross_sum = np.nansum(res)
-        expected = 32114793.0
+        expected = 87281406.0
```
djhoese, line 905: What happened here?

djhoese: Oh this is my change from #265

PR #271, opened 2020-05-04
Create AreaDefinition from a netCDF/CF file

TomLav:
- [x] Closes #269 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Passes ``git diff origin/master **/*py | flake8 --diff`` 
 - [x] Fully documented

PR #279, opened 2020-05-12
Refactor API docs to document things more easily

djhoese:
After talking with @TomLav about #271 and how to document the functions in sphinx, it seemed like a lot of changes needed to be made for a good long term solution to documenting the pyresample API. This PR is those changes.

NOTE: This does move around some of the modules, but nothing that should have been user-facing.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
This last commit has some interesting fixes for some interesting findings. In the past, I added a few things under `Attributes` in the class docstring for `AreaDefinition`. Sphinx is now actually detecting that these are duplicates of the docstrings in the AreaDefinition property definitions.

pyresample/geometry.py
```diff
@@ -1222,7 +1210,8 @@ def shape(self):
 
     @property
     def resolution(self):
-        """Return area resolution in X and Y direction."""
+        """Return area resolution in X and Y direction.
+        """
```
pnuu, line 1214:
```suggestion
        """Return area resolution in X and Y direction."""
```

PR #302, opened 2020-10-14
Replace inf with NaN in lat/lon coordinates

sfinkens:
Edit: We decided to add a `fill_value` kwarg and to *not* change the default fill value in order to stay consistent with pyproj/PROJ.

---

`AreaDefinition.get_lonlats()` returns `inf` for space pixels in the geostationary projection (coming from `PROJ`). Since `inf` may cause trouble in certain situations (see below), add a new keyword argument `fill_value=np.nan` which changes the default fill value to `NaN`. Examples for problems with `inf`:

a) Selecting high latitudes

```
>>> lat = np.array([10, 20, 30, np.nan, np.inf])
>>> lat > 25
array([False, False,  True, False,  True])
```

b) Satpy's CF writer (and xarray in general) uses `NaN` as default fill value, so that `inf` values are not being masked properly.

As discussed on slack, an upstream solution in pyproj or even PROJ would be preferred, but might take some time. See https://github.com/pyproj4/pyproj/issues/721 and https://github.com/OSGeo/PROJ/issues/2376 . 
An alternative solution could be to change the fill value in the `Proj` and `ProjMP` classes. This would target all projection operations. But since the long term wish is to get rid of at least the `Proj` class, the keyword argument solution seemed like a good compromise.

 - [X] Tests added 
 - [x] Tests passed 


djhoese:
Would it be possible to skip the conversion if the array dtype is integer (or rather doesn't support NaN)?

sfinkens:
Putting this on hold, maybe it can be resolved upstream in PROJ:

https://github.com/OSGeo/PROJ/issues/2376

sfinkens:
> Would it be possible to skip the conversion if the array dtype is integer (or rather doesn't support NaN)?

I've issued a warning if the data type doesn't support the fill value

pnuu:
You may want to merge master to this PR, there are some fixes to the different CI build systems that may help them pass.

sfinkens:
Thanks for the hint @pnuu !

sfinkens:
While trying to fix the tests I found that in many places the valid range of latitudes [-90, 90] and longitudes [-180, 180] is checked. These comparisons work differently with `NaN`, so I replaced them with a helper method. All the tests pass now, but I feel a little uncomfortable because I changed many methods I'm not familiar with.

djhoese:
Well if the pyresample tests didn't work with NaN but worked with Inf then maybe users are going to run into the same thing 😈

sfinkens:
Yes, but some users are also running into problems with Inf :) Pyresample should now be able to work with both Inf and Nan. We could keep the default fill value Inf, but issue a warning that this will change in a future release?

djhoese:
I think we've decided not to go ahead with this change, right? Can we close this?

sfinkens:
Yes, correct!

djhoese:
It was pointed out on slack that this PR doesn't actually have the reasons for why this was closed. I'll try to summarize what I remember from past discussions. Here are the pros and cons of allowing the user to fill invalid or space pixels with NaNs instead of the current default Inf. Note the distinction in some of these that NaN could be the new default with no option for Inf, or NaN would be the new default with a kwarg for going back to Inf, or Inf could still be the default with a kwarg to use NaN instead, or the ultimate alternative would be to not have pyresample use NaN at all in `get_lonlats`.

## Pros

1. More flexibility for how users can use the returned values.
2. Simpler arithmetic when the user gets the values they expect.

## Cons

1. Another keyword argument.
2. Inconsistency with PROJ/pyproj defaults for similar operations.
3. Handling of NaNs or Inf in other pyresample functions that expect lon/lats as input.
4. Unexpected results for some arithmetic versus others (`lons < 180` will return `True` for `inf` but `False` for `nan`).
5. Infinity is arguably more appropriate for pixels that are "beyond" the coordinate system for the Area versus NaN which typically has a connotation of invalid. Infinity is also returned by some projection calculations inside PROJ because it naturally results from the arithmetic involved.
6. Able to distinguish off-CRS versus invalid pixels. However, normal AreaDefinition operations should all be in the CRS domain until the user or pyresample wants to start masking invalid pixels from a data array. Even then, that information is not kept in the AreaDefinition itself but in an array that the user or a Resampler controls.

Note that out of all the projections and work that could be done related to projections these off-lon/lat pixels are rare. They mostly show up when you have space pixels most common in geostationary projections.

PR #310, opened 2020-11-05
Remove appveyor CI in favor of travis Windows and add Python 3.9 wheels

djhoese:
Appveyor support is deprecated by ci-helpers which we are using to easily download and install conda packages. We already have Windows being tested on travis but it is an allowed failure. This PR removes appveyor and moves the Windows environment out of the allowed failure on Travis.


djhoese:
One downside I'm realizing after all of these changes is that Python 3.9 wheels still aren't built because we are waiting on a pykdtree wheel (installing from source currently fails on the wheel docker images).

djhoese:
This is ready for review otherwise.

pnuu (approved):
LGTM

PR #311, opened 2020-11-08
[WIP] Compatibility with PyProj >= 3.0

avalentino:
This PR  should fix compatibility with PyProj >= 3.0.
Probably the implementation is not as fast as in the case of PyProj < 3.0 but at least it does not raise errors.

The PR also include the re-generated C file for ``_ll2cr.pyx``. Not sure if there is some specific constraint on the Cython version to be used. 

 - [ ] Tests added
 - [ ] Tests passed


avalentino:
Not sure if the error in RTD is due to changes in this PR. Can you please provide some guidance?

djhoese:
I have another PR open which fixes the RTD problem. Pyproj 3 requires using a newer pip than I was accidentally using with RTD.

djhoese:
Actually your changes should all be included in my PR #310

avalentino:
> Actually your changes should all be included in my PR #310

Indeed, I see it now.
Sorry I haven't looked at already open PRs before opening this one.
Closing.

PR #314, opened 2020-11-25
Add citation information

cgentemann:
- [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
Thanks!

Could you explain how this zenodo json file works? Does zenodo use that to fill in its entry for pyresample or did zenodo generate this file for you?

Side note: We should probably find a way to remove stickler bot from the authors list.

cgentemann:
;) yes, probably remove stickler bot from the authors list... ;)
I have mixed feelings about the .zenodo.json file.  It allows you to add metadata to the zenodo record.  
So, it is good because it links to orcids, institutions, you can put grant funding, etc. 
It is bad because it will have to be updated for each release.
I'm okay with not including if you feel like it might be hard to maintain.
Chelle

djhoese:
> It is bad because it will have to be updated for each release.

Aw, really? Yeah that's kind of a pain then. I like all the affiliation stuff, but don't want to have to update the version number and stuff every release. It is weird that it doesn't just supplement the information automatically gathered, but :man_shrugging: 

I think I'm :-1: on it then. @pytroll/core Opinions?

cgentemann:
@pytroll/core Opinions?    I don't undestand.

djhoese:
I'm asking the other pytroll (pyresample) maintainers what they think is best.

sfinkens:
Good initiative! I see that satpy already has a zenodo badge. I guess this is done using webhooks? https://guides.github.com/activities/citable-code/ .

djhoese:
@sfinkens The badge in satpy was manually added. Zenodo automatically finds github releases once it is told to look for them. Pyresample has already had a zenodo DOI (99% sure) but it was never linked in the README.

sfinkens:
Yes indeed: https://zenodo.org/record/4270378 . Then let's just add the badge :)

djhoese (approved):
Thanks!

djhoese:
This is only README changes and I don't want to forget about this PR again so I'm merging it before the tests pass.

PR #330, opened 2021-02-11
Fix a division-by-zero failure causing missing data in bilinear interpolation

### LINKED ISSUE #295: Undetermined values in bilinear resampling result when resampling swath to grid data with geographic output coordinates.
I came across an issue with the bilinear resampler in case I resample swath data to grid data with geographic coordinates in the output area definition.

I tried to create a minimal code example to demonstrate what I mean:

```python
import numpy as np
from matplotlib import pyplot as plt
from scipy import ndimage
from pyresample import bilinear
from pyresample.geometry import SwathDefinition, create_area_def

# generate some sample swath data that we want to resample to grid data later
swath_data = np.meshgrid(np.arange(100, 200), np.arange(100, 200))[0]

# get corresponding longitude/latitude arrays
lons, lats = np.meshgrid(np.linspace(0, 1, 200), np.linspace(51, 50, 200))
lons = ndimage.rotate(lons, 10, mode='constant')[50:150, 50:150]
lats = ndimage.rotate(lats, 10, mode='constant')[50:150, 50:150]

# create a target area definition (we want the swath data to be resampled to geographic coordinates)
#    NOTE: The area extent and shape are fictive, they should not really matter here.
target_def = create_area_def(area_id='',
                             projection='+proj=longlat +datum=WGS84 +no_defs',
                             area_extent=(0, 50, 1, 51),
                             shape=(100, 100),
                             units='degrees')

# define the source swath definition
source_def = SwathDefinition(lons=lons, lats=lats)

# resample as documented here: https://pyresample.readthedocs.io/en/latest/swath.html#resample-bilinear
result = bilinear.resample_bilinear(swath_data, source_def, target_def,
                                    radius=50e3, 
                                    neighbours=32,
                                    nprocs=1, 
                                    fill_value=0,
                                    reduce_data=True, 
                                    segments=None,
                                    epsilon=0)
```

The input arrays look like this:
swath_data:
grafik

lons:
grafik

lats:
grafik

I expect the output to look exactly the same like the swath data array, but it must be rotated somehow.
The rotation seems to be correct and pixel values make sense. However, the result contains undetermined values (pixel value is 0, i.e., the fill value) within the actual image boundaries:

grafik

How can I avoid these unfilled pixels in the output? Is this a bug in pyresample or what am I doing wrong? I already tried to increase or decrease all the keyword arguments of bilinear.resample_bilinear() but without success.

I am using a freshly installed Anaconda Python 3.8.5 with pyresample 1.16.0 (conda-forge).

pnuu:
- [x] Closes #295 
 - [ ] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/pyresample/bilinear/_base.py
+++ b/pyresample/bilinear/_base.py
@@ -382,11 +382,16 @@ def _solve_quadratic(a__, b__, c__, min_val=0.0, max_val=1.0):
     # Solve the quadratic polynomial
     x_1 = (-b__ + np.sqrt(discriminant)) / (2 * a__)
     x_2 = (-b__ - np.sqrt(discriminant)) / (2 * a__)
+    x_3 = -c__ / b__
 
     # Find valid solutions, ie. 0 <= t <= 1
     x__ = np.where(
-        find_indices_outside_min_and_max(x_1, min_val, max_val),
+        find_indices_outside_min_and_max(x_1, min_val, max_val) | np.isnan(x_1),
         x_2, x_1)
+
+    x__ = np.where(
+        find_indices_outside_min_and_max(x__, min_val, max_val) | np.isnan(x__),
+        x_3, x__)
 
     x__ = np.where(
         find_indices_outside_min_and_max(x__, min_val, max_val),


mraspaud (approved):
LGTM

PR #336, opened 2021-03-08
Switch to building cython extensions at installation time

djhoese:
We currently provide the C/C++ rendered files from Cython in our sdist and in the repository. This causes issues when Python changes its API and Cython needs to re-render the files to match. This PR changes the necessary pieces of the setup.py and CI so this rendering always happens on installation. Users without compilers will need to depend on the wheels on PyPI.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [x] Fully documented


pnuu (approved):
LGTM. This will clean-up also the diffs a lot for the following PRs touching the Cython code.

mraspaud (approved):
LGTM, thanks!

PR #343, opened 2021-03-22
Fix EWA resampling when the result is all fill values

djhoese:
This was a bug I discovered at the same time I found #342, but had a really difficult time reproducing. The main piece of code that is effected is *only* called when dask determines that it should call the "aggregate" function of the reduction with a specific result from the previous "combination" calls. I was able to force this by making the input one single chunk.

 - [x] Tests added 
 - [x] Tests passed 


mraspaud (approved):
LGTM

PR #356, opened 2021-05-19
Add `get_min` and `get_max` to Bucket resampler (experimental)

zxdawn:
This PR adds two functions to Bucket: `get_min` and `get_max`.

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -251,6 +251,7 @@ intersphinx_mapping = {
     'scipy': ('https://docs.scipy.org/doc/scipy/reference', None),
     'xarray': ('https://xarray.pydata.org/en/stable', None),
     'dask': ('https://docs.dask.org/en/latest', None),
+    'pandas': ('https://pandas.pydata.org/docs', None),
     'pyresample': ('https://pyresample.readthedocs.io/en/stable', None),
     'trollsift': ('https://trollsift.readthedocs.io/en/stable', None),
     'trollimage': ('https://trollimage.readthedocs.io/en/stable', None),
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -190,6 +190,35 @@ class BucketResampler(object):
             statistic = da.where(nan_bins > 0, np.nan, statistic)
         return statistic
 
+    def _call_scipy_binned_statistics(self, scipy_method, data, fill_value=None, skipna=None):
+        """Calculate statistics (min/max) for each bin with drop-in-a-bucket resampling."""
+        from scipy.stats import binned_statistic
+
+        if isinstance(data, xr.DataArray):
+            data = data.data
+        data = data.ravel()
+
+        # Remove NaN values from the data when used as weights
+        weights = da.where(np.isnan(data), 0, data)
+
+        # Rechunk indices to match the data chunking
+        if weights.chunks != self.idxs.chunks:
+            self.idxs = da.rechunk(self.idxs, weights.chunks)
+
+        # Calculate the min of the data falling to each bin
+        out_size = self.target_area.size
+        statistics, _, _ = binned_statistic(self.idxs, values=weights, statistic=scipy_method,
+                                            bins=out_size, range=(0, out_size))
+        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int)).ravel()
+
+        # TODO remove following line in favour of weights = data when dask histogram bug (issue #6935) is fixed
+        statistics = self._mask_bins_with_nan_if_not_skipna(skipna, data, out_size, statistics)
+
+        # set bin without data to fill value
+        statistics = da.where(counts == 0, fill_value, statistics)
+
+        return statistics.reshape(self.target_area.shape)
+
     def get_min(self, data, fill_value=np.nan, skipna=True):
         """Calculate minimums for each bin with drop-in-a-bucket resampling.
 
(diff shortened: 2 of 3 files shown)


zxdawn:
Example of processing ABI L2 CTH data:


image

djhoese:
You mentioned on slack that there is code repeated between the two methods. Would it be possible to make a generic method that is given the function name to pass to scipy?

djhoese:
Also how does this perform for some of the other functions provided by `binned_statistics` that are already implemented here (mean/average, count, sum, etc)? Would it be better to copy from the existing functions to support min/max if it means better performance? Or should those be switched to depend on scipy?

zxdawn:
As discussed in slack, we prefer the xarray or dask method.
1. xarray method
    `range()` isn't supported in xarray's `groupby_bins` function (see [this issue](https://github.com/pydata/xarray/issues/5358)).
2. dask method
    dask only supports part of scipy.stats (see the [Doc](https://docs.dask.org/en/latest/array-stats.html)). `binned_statistics` isn't in the list.

zxdawn:
@djhoese I switched to the xarray method. The `weights` needs to be DataArray and I set it to DataArray by `xr.DataArray()`. The speed of calculation is quite low ... It should be something related with dask. Do you have any suggestions?

djhoese:
I think overall the groupby operations by xarray don't perform well. There was a presentation at the dask summit this week about a dask-centric groupby package that is being worked on, but it is only experimental at this point. Otherwise I think a custom implementation would be needed. I don't know enough about how the bucket resampler works to give any more advice. Maybe @pnuu ?

zxdawn:
@djhoese Yes, the xarray groupby operations are too slow. So, instead I tested the pandas method.

# Test
## Method1: pandas groupby with pd.cut
```
length = int(1e5)
x = np.arange(length)
df = pd.DataFrame({'values': x*5})
groups = df.groupby(pd.cut(x, bins=np.linspace(0, length, length+1), right=False, include_lowest=True))
bin_min = groups['values'].min()
```
2.31 s ± 4.51 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

## Method2: pandas groupby with np.digitize
```
%%timeit

length = int(1e5)
x = np.arange(length)
df = pd.DataFrame({'values': x*5})
groups = df.groupby(np.digitize(x, bins=np.linspace(0, length, length+1)))
bin_min = groups['values'].min()
```
6.65 ms ± 4.33 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

##  Method3: dask pandas groupby with np.digitize
```
%%timeit

length = int(1e5)
df = pd.DataFrame({'x': np.arange(length), 
                   'values': np.arange(length)*5})

df_dask = dd.from_pandas(df, npartitions=3)

bin_min = df_dask.map_partitions(lambda part: part.groupby(np.digitize(part.x,
                                                                       bins=np.linspace(0, length, length+1)
                                                                      )
                                                          )['values'].min()).compute()
```
15.9 ms ± 772 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

##  Increasing the data length
The new result of consumed time is
```
method1: 24.1 s ± 101 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
method2: 108 ms ± 834 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
method3: 72.5 ms ± 7.25 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

If I change `npartitions` to 10, the speed is quicker:
```
method3: 59.3 ms ± 3.17 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
If `npartitions` is set to 1, it's much slower:
```
212 ms ± 15.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
## Summary
I think we can use dask pandas groupby with np.digitize. However, I don't know how to set the suitable npartitions automatically.

zxdawn:
I create the dask DataFrame directly from dask array. It works well now.

zxdawn:
The calculation speed is not quite good compared with other resampler method ...

```
$ blended_scene.resample(coarse_area, resampler='bucket_count')
25.3 ms ± 369 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

$ blended_scene.resample(coarse_area)
36.7 ms ± 2.46 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

$ blended_scene.resample(coarse_area, resampler='bucket_min')
14.3 s ± 93 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Any idea of how to speed up?

pnuu (approved):
LGTM, and thanks for adding the features!

As discussed on Slack, a short note on (current) slowness in the `get_min()` and `get_max()` docstrings would be nice.

zxdawn:
The docstring is added, but the cross-linked function `_call_pandas_groupby_statistics` can't be clicked:
image

djhoese:
This is likely because by default sphinx won't document private (`_`) methods. I think you should just link to the pandas method directly. You'll likely need to add reference to pandas in the intersphinx mapping in `doc/conf.py`.

djhoese (approved):
Looks good to me.

CC to @dcherian that this PR would really benefit from your dask groupby work. I saw your presentation at the dask summit and that's why the project is referenced in the docstring here as a better option for what @zxdawn is trying to do.

dcherian:
Thanks. @zxdawn you could test out https://github.com/dcherian/dask_groupby and see if there are any bugs. I think work is starting soon on integrating it in to xarray

zxdawn:
@dcherian Thanks. I will test a simple example and submit an issue if there's any bug.

PR #363, opened 2021-05-25
Fix version number not being preserved in wheels

### LINKED ISSUE #361: __version__ in pip wheel  is mangled
#### Various installation and version checks

```bash
pip install pyresample==1.19.0 --no-binary :all: # don't use wheel, install from tar.gz
python -c "import pyresample; print(pyresample.__version__)"
1.19.0

pip install pyresample==1.19.0
python -c "import pyresample; print(pyresample.__version__)"
0+unknown

pip install pyresample==1.18.0
python -c "import pyresample; print(pyresample.__version__)"
0+unknown

pip install pyresample==1.17.0
python -c "import pyresample; print(pyresample.__version__)"
0+unknown

pip install pyresample==1.16.0
python -c "import pyresample; print(pyresample.__version__)"
0+unknown

pip install pyresample==1.15.0. # first version to include wheels
python -c "import pyresample; print(pyresample.__version__)"
0+unknown

pip install pyresample==1.14.0  # no wheels available install from tar.gz
python -c "import pyresample; print(pyresample.__version__)"
1.14.0
```

#### Problem description
The internal version attribute for pyresample from the wheels is mangled.  Satpy utilizes this and will block functionality with the mangled version.  https://github.com/pytroll/satpy/blob/378a2c563f647a7bb7ce74b417f288d0738d2762/satpy/resample.py#L182

Should satpy be querying the version in a different way or do the wheels need to be fixed?

#### Expected Output
the correct version number when calling `pyresample.__version__`

edit: added comments regarding the various install versions to document the method to my madness

abrammer:
Make sure that the wheel distributions include the versioneer details correctly. 

 - [x] Closes #361  
 - [x] Tests added - added test in deploy to install and check internal version
 - [ ] Fully documented 
       N/A ?

Tests for this need to be local for now, 
```bash
python -m build -w
pip install dist/pyresample*.whl
python -c "import pyresample; print(pyresample.__version__)"
```

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -77,6 +77,14 @@ jobs:
             quay.io/pypa/${{ matrix.docker-image }} \
             /io/continuous_integration/build-manylinux-wheels.sh
 
+      - name: Check version number from inside wheel
+        if:  matrix.docker-image != 'manylinux1_i686'
+        run: |
+          mv pyresample unused_src_to_prevent_local_import
+          python -m pip install --find-links=./dist/ pyresample
+          python -c "import pyresample; print(pyresample.__file__, pyresample.__version__)"
+          python -c "import pyresample; assert 'unknown' not in pyresample.__version__, 'incorrect version found'"
+
       - name: Upload wheel(s) as build artifacts
         uses: actions/upload-artifact@v2
         with:
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,7 +10,7 @@ max-line-length = 120
 VCS = git
 style = pep440
 versionfile_source = pyresample/version.py
-versionfile_build =
+versionfile_build = pyresample/version.py
 tag_prefix = v
 
 [coverage:run]


djhoese:
Thanks for making this.

For tests, is there something we could look for in the wheel's files that wouldn't require actually installing the wheel? Since wheels are just zip files I wonder if we could unzip one and grep for something in the version file. Maybe add something to https://github.com/pytroll/pyresample/blob/main/continuous_integration/build-manylinux-wheels.sh?

abrammer:
Yeah, unzipping and checking the `pyresample/version.py` would be pretty simple.  pro or con of running the full test suite on each wheel.  I got it running for everything but the 32 bit linux,  but notably the windows 3.6 tests fail, not sure if it's a legit fail or a just an actions things.  

example of the deploy workflow with tests here, dropped the upload jobs from this file for test purposes.  
https://github.com/abrammer/pyresample/runs/2668214477?check_suite_focus=true

abrammer:
Pushed the much simpler version check, just import and check 'unknown isn't in the `__version__` rather than running the full test suite.  
Example fail on the old setup.cfg : https://github.com/abrammer/pyresample/actions/runs/876423241
Example success on this PR setup.cfg:  https://github.com/abrammer/pyresample/actions/runs/876422055

zxdawn:
I tested this PR. But it still doesn't work for installing it on a node without Internet.
```
$ pip install -e .
$ python

$ import pyresample
$ pyresample.__version__

'0+unknown'
```

abrammer:
@zxdawn can you detail the steps you took.  It looks like you transfered the repo to the node? Installing from the repo, relies on git to work out the version. If you didn't transfer the `.git` or have `git` installed on the node. Unknown might be the expected result.  
This PR was tackling the wheel versions from pypi. 

You could transfer one of the wheels to the node instead, and install that and then this PR would be relevant and should fix the issue.  There are corrected built wheels in the artififacts from the success example in my previous comment here. 
(Not that familiar with versioneer so might have missed something)

zxdawn:
@abrammer Thanks for the PR. Here're the details:

1. Clone the repo to the admin (the node with Internet)
2. Installed the modified repo
3. The version is correct on the admin, but not on the compute node which is without the Internet. Note that admin and compute node share the same storage.

BTW, I got this issue today and it worked well in the last few days ... Strange.

abrammer:
Does the main branch work? I'm thinking it's possibly unrelated to the PR. 

The related issue #361, might be the better place to discuss the issue if it's not related to the changes of the PR.

zxdawn:
> Does the main branch work? I'm thinking it's possibly unrelated to the PR.

The main branch also doesn't work.

I'm not familiar with the pip and not sure whether this PR should fix this problem. What's your opinion, @djhoese?

djhoese:
@zxdawn So you are `pip install -e .` on the admin/head node of a cluster and then using that shared installation directory (shared between nodes) and when you check the version you get unknown?

Installing things in editable mode (the `-e` in your `pip`) seems like a questionable approach on a multi-node cluster. My guess is that versioneer is creating a `version.py` that checks `git` to figure out the version. This is because of the editable mode. *BUT* this requires `git` to be installed on the node running the python code and importing/running the `version.py` module (as @abrammer said). I would be surprised this ever worked unless:

1. You had `git` installed in some environment and didn't realize it and were using that on the child nodes. Or:
2. You were installing it with `pip install .` (no `-e`) which probably should have worked in current `main` but should definitely work with this PR's fix. Or:
3. Versioneer has some dependence on the version of `git` installed and your nodes have different versions. Wild guess and unlikely, but still want to say it.

zxdawn:
@djhoese Thanks for your suggestions!
1. The git is available at `/use/bin/git` on the admin, but not on the compute node. I installed the git in the conda environment and `pip install -e .` works again now ;)
2. Yes, `pip install .` works without the need for git.

djhoese:
@abrammer Thanks for putting in this work. It looks like this is all behaving as expected. @mraspaud do you have any issues with this? Otherwise I'm thinking a merge and a patch release of pyresample this week.

djhoese:
Thank you!

PR #367, opened 2021-06-04
Fix AreaDefinition.get_area_slices returning non-integer slices

djhoese:
As part of https://github.com/pytroll/satpy/issues/1714 we found out that the `AreaDefinition.get_area_slices` method seems to actually depend on numpy 1.19.0 and above. Otherwise it produces slices with floating point numbers. We *could* fix this (if someone wants to go through the work of testing pyresample on older versions of numpy and fixing this bug then :+1: ), but the easier solution right now is to just bump the minimum version of numpy.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/continuous_integration/environment.yaml
+++ b/continuous_integration/environment.yaml
@@ -2,6 +2,7 @@ name: test-environment
 channels:
   - conda-forge
 dependencies:
+  - numpy >= 1.19.0
   - xarray
   - dask
   - distributed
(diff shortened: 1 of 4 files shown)


djhoese:
I'm working on updating the tests to make sure that slice elements are integers and it looks like I'm going to have to fix this properly. Some of the tests aren't producing integer indices even with newer numpy versions.

PR #368, opened 2021-06-05
Speed up Bucket `get_min` and `get_max`

zxdawn:
- [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented 

The pandas method is quite slow. I come up with the faster method by combing `numpy.ufunc.reduceat()` and `xr.apply_ufunc()`.


zxdawn:
The method sounds reasonable to me, but the result is wrong:
image

It seems that `get_min` and `get_max` functions are applied over some part of the data repeatedly ...
Do you have any ideas? @pnuu

pyresample/bucket/__init__.py
```diff
@@ -209,27 +206,50 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
+        # sort the slices
+        sort_index = da.map_blocks(np.argsort, slices.data)
+        slices = slices[sort_index]
+        weights = weights[sort_index]
+
+        # get the unique slices (for assignment later) and bins (for numpy_reduceat)
+        unique_slices, unique_bins = da.unique(slices.data, return_index=True)
+        statistics_sub = xr.apply_ufunc(numpy_reduceat,
+                                        weights,
+                                        unique_bins.compute_chunk_sizes(),
```
zxdawn, line 239: I have to use `compute_chunk_sizes()` because the chunks is unknown after `da.unique()` ...

zxdawn:
@pnuu I fixed the bug and the result looks well now:
image

However, the speed is slow again ...

I created a simple example below to test what leads to slow speed.
The pure numpy is much quicker (x4300) than dask method!
Any ideas? @djhoese @mraspaud 

To reproduce this example, you can copy them into jupyter notebook cells and execute them.

## Cell1: Create sample data
```
import random
import numpy as np
import xarray as xr
import dask.array as da

# create example data
out_size = int(1e5)
bins = np.arange(3, out_size-3)
idxs = np.arange(0, out_size)
random.shuffle(idxs)

weights = idxs
slices = np.digitize(idxs, bins)
```

## Cell2: Calculate bin_min using pure numpy
```
%%timeit
# get index and weights in bins
valid = (bins.max() >= idxs) & (idxs >= bins.min())
valid_slices = slices[valid]
valid_weights = weights[valid]

# sort slice and weights
sort_index = valid_slices.argsort()
valid_slices_sort = valid_slices[sort_index]
valid_weights_sort = valid_weights[sort_index]

unique_slices, unique_index = np.unique(valid_slices_sort, return_index=True)
res_sub = np.minimum.reduceat(valid_weights_sort, unique_index)

# initilize result
res = xr.DataArray(np.full((out_size), np.nan), dims=['x'])
res.loc[unique_slices-1] = res_sub
```

### Time

```
9.6 ms ± 34.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

## Cell3: Calculate bin_min using dask and xarray
```
def numpy_reduceat(data, bins, statistic_method):
    '''Calculate the bin_statistic using numpy.ufunc.reduceat'''
    if statistic_method == 'min':
        return np.minimum.reduceat(data, bins)
    elif statistic_method == 'max':
        return np.maximum.reduceat(data, bins)

# convert weights to DataArray using idxs as coords
weights = xr.DataArray(da.from_array(weights), coords=[idxs], dims=['x'])
slices = xr.DataArray(da.from_array(np.digitize(idxs, bins)), dims=['x'])

%%timeit
# convert weights to DataArray using idxs as coords
mask = xr.DataArray((idxs >= bins.min()) & (idxs <= bins.max()), dims=['x'])
valid_weights = weights.where(mask, drop=True)
valid_slices = slices.where(mask, drop=True)

# sort the slices
sort_index = da.map_blocks(np.argsort, valid_slices.data)
valid_slices = valid_slices[sort_index]
valid_weights = valid_weights[sort_index]

# get the unique slices (for assignment later) and bins (for numpy_reduceat)
unique_slices, unique_bins = da.unique(valid_slices.data, return_index=True)
statistics_sub = xr.apply_ufunc(numpy_reduceat,
                                valid_weights,
                                unique_bins.compute_chunk_sizes(),
                                kwargs={'statistic_method': 'min'},
                                input_core_dims=[['x'], ['new_x']],
                                exclude_dims=set(('x',)),
                                output_core_dims=[['new_x'],],
                                dask="parallelized",
                                output_dtypes=[weights.dtype],
                                dask_gufunc_kwargs={'allow_rechunk': True},
                            )

# initialize the output DataArray with np.nan
statistics = xr.DataArray(da.from_array(np.full((out_size), np.nan)), dims=['x'])
# assign the binned statistics
statistics.loc[unique_slices.astype('int')-1] = statistics_sub
statistics.compute()
```

### Time

```
41.3 s ± 91.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

pyresample/bucket/__init__.py
djhoese: So I won't even pretend to know how this algorithm is working, but there a couple things I can point out (since you asked for my review). The main thing is that with dask you can't blindly add a `map_blocks` to a function that doesn't work on chunks.

1. I'm guessing `argsort` is a function that can't be map_blocks'd. It needs to look at every chunk to know how to sort them. So the benefit of dask here kind of disappears. Let me know if you want a more complete explanation of this.
2. The `compute_chunk_sizes` function (as the dask documentation says) computes the result. This means that these steps to create unique_bins are not doing things lazily as a Satpy user may expect.
3. Below you do `da.from_array(np.full(...))`. This is creating a huge numpy array and then giving it to dask to be "lazy". The numpy array operation removes any dask-ness of this operation and takes up a lot of memory. Doesn't dask have a `full` function?
4. You then write to this `statistics` dask array which may be supported by dask (maybe), but it likely doesn't perform very well as it is again not a chunk-friendly operation.

Note that this algorithm you are trying to build is inherently difficult to do in a completely dask-friendly manner. The basic operation of "max" or "min" means you have to look at many data points that may not be in the same chunk.

The "best" (maybe) way to structure this algorithm would be if you could get each bin of points to be its own dask chunk. Then run `map_blocks` (or maybe a straight numpy min) along the axes of these chunks. That way the `min` function is getting a numpy array of all the points in a bin.

Sorry I can't be of more help.

zxdawn: Yes, the method is quite complicated. So, I try my best to post the question on [Stack Overflow](https://stackoverflow.com/questions/67882615/fastest-way-to-get-the-minimum-value-of-data-array-in-another-paired-bin-array). Hope someone can give us a simpler solution of calculating bin_min.

zxdawn: Someone has posted the quick method. But, I still have no idea how to apply it with dask ....
For the last example of the [answer](https://stackoverflow.com/a/67900968/7347925), it just costs 9 seconds! But, when I apply it to `get_min`, it stuck ...
I would appreciate any suggestions!

djhoese: Yeah doesn't look like that algorithm will be very dask-friendly. You could naively wrap it in a `@delayed` decorator and then do `da.from_delayed` to get a dask array back. It is no longer parallel, but for small amount of data that isn't a big deal. Otherwise you're getting close to having to write a custom dask algorithm with building your own task graph and stuff.

pyresample/bucket/__init__.py
```diff
@@ -209,27 +206,39 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
+        extended_bins[-1] = np.iinfo(bins.dtype).max  # last bin goes to infinity
+
+        # preallocate buffers to avoid array creation in main loop
+        chunk_size = max(len(weights)//int(1e3), len(bins))
+        mask_buffer = da.empty((chunk_size, len(self.idxs) + 1), dtype=bool)
+        mask_buffer[:, -1] = True
+        statistics = da.empty_like(bins, dtype=np.float64)
+
+        # calculate the statistics in each chunk
+        for low in range(0, len(bins), chunk_size):
```
zxdawn, line 234: @pnuu  @djhoese Now, I switched to the new method. But, have now idea of applying dask parallel or `da.from_delayed()` to this loop. Another problem is that I don't know the suitable `chunk_size` for the loop, and it is set manually ...

djhoese (commented):
The algorithm needs to be broken up into different parts. It seems there are parts that are chunk-friendly and parts that aren't. It also seems like you've tried converting the SO answer into dask by switching some `np.` calls to `da.`. In some cases this may be unnecessary as numpy will call dask's methods properly. In other cases it has no effect or is actually worse. For example, `da.empty` will create an empty dask array, but then you need to do assignments on it which will generally not perform well with dask.

You may also want to look at dask's `blockwise` if your algorithm ends up going from some chunked array shape to another chunked array shape which is typical for resampling. For example if you had input arrays to an algorithm that were (bins, y, x) and needs to go to (bins, y2, x2) then `blockwise` would be a good fit. https://docs.dask.org/en/latest/array-api.html#dask.array.blockwise

If you can separate the part of the algorithm that is dealing with multiple chunks at a time (sorting, assigning to random areas of the output array, etc) into its own function and then wrap that in delayed, you can do whatever you want with numpy functions inside that function.

The algorithm you were provided may be the fastest numpy algorithm and it will still perform well in a delayed function but there is probably another algorithm (or a version of this algorithm) that would use dask chunks in a better way. Even if they did this it might not perform as well as the pure numpy version, but it would at least be able to operate per-chunk and never load more data than it needed. How this dask-friendly algorithm would be coded is not obvious and would probably not be worth the effort unless you want to use the opportunity to learn how dask graphs work (really not that bad).

This would be much easier to talk about in person with a whiteboard in front of us.

zxdawn:
Thank you very much, @djhoese. The explanation is quite clear and I will try my best to improve this PR for dask tomorrow ;)

pyresample/bucket/__init__.py
```diff
@@ -209,27 +206,40 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
-                                           )['values'].max())
-
-        # fill missed index
-        statistics = (statistics + pd.Series(np.zeros(out_size))).fillna(0)
+        # create the output bins
+        bins = da.linspace(0, out_size-1, out_size).astype('int')
+
+        # sort idxs and weights based on weights
+        if statistic_method == 'min':
+            order = da.map_blocks(np.argsort, weights)
```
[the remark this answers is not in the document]
zxdawn: Yes, that should be the entire array.

zxdawn:
@djhoese Following your suggestions, I modified the creation of empty array and switched to `to_delayed()` and `da.from_delayed()`. But, I met two problems.

pyresample/bucket/__init__.py
```diff
@@ -211,35 +213,42 @@ def _call_bin_statistic(self, statistic_method, data, fill_value=None, skipna=No
...
-                                            (extended_bins[low+1:high+1, None] > idxs_sorted[None, :])
-            mask = mask_buffer[:chunk_size, ...]
-            weight_idx = np.argmax(mask, axis=-1)
-            statistics[low:high] = weights_sorted[weight_idx]
+
+        # calculate the statistics in each block
+        blocks = da.from_array(bins, chunks=chunk_size).to_delayed()
+        results = [da.from_delayed(get_block_statistic(block, mask_buffer,
+                                                       idxs_sorted, weights_sorted),
+                                   shape=(chunk_size,),
```
zxdawn, line 246: The `chunk_size` is set manually and will cause the wrong shape of statistics because the shape of some blocks are different. If I use `shape=(np.nan,)`, this could lead to errors of `statistics = da.where(counts == 0, fill_value, statistics)`, which needs to know the exact shape.

pyresample/bucket/__init__.py
```diff
@@ -211,35 +213,42 @@ def _call_bin_statistic(self, statistic_method, data, fill_value=None, skipna=No
...
+            order = np.argsort(weights)
         elif statistic_method == 'max':
-            order = da.map_blocks(np.argsort, weights)[::-1]
+            order = np.argsort(weights)[::-1]
+
+        @dask.delayed
+        def get_block_statistic(block, mask_buffer, idxs_sorted, weights_sorted):
+            extended_bins = np.append(block, np.iinfo(block.dtype).max)
+            # use the preallocated buffers
+            mask_buffer[:len(bins), :-1] = (bins[:, None] <= idxs_sorted[None, :]) &\
```
zxdawn, line 224: I use the preallocated buffers, but it could lead to a different result in each run. That's strange.
If I create the new mask in each run like this:
```
            mask = (block[:, None] <= idxs_sorted[None, :-1]) &\
                          (extended_bins[1:, None] > idxs_sorted[None, :-1])
            mask = np.hstack((mask, np.ones((mask.shape[0], 1), dtype=mask.dtype)))

```
it works well but will raise memory value.

pyresample/bucket/__init__.py
```diff
@@ -209,27 +208,47 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
+            order = np.argsort(weights)[::-1]
 
-        elif scipy_method == 'max':
-            statistics = df.map_partitions(lambda part: part.groupby(
-                                           np.digitize(part.x,
-                                                       bins=np.linspace(0, out_size, out_size)
-                                                       )
-                                           )['values'].max())
+        @dask.delayed
+        def get_block_statistic(block, mask_buffer, idxs_sorted, weights_sorted):
```
djhoese, line 221: Writing this here now before I forget: this should be moved to the global scope (a global function) or a staticmethod. Dask has trouble serializing local/hidden functions for non-threaded schedulers (distributed).

pyresample/bucket/__init__.py
```diff
@@ -209,27 +208,47 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
 
-        if scipy_method == 'min':
-            statistics = df.map_partitions(lambda part: part.groupby(
-                                           np.digitize(part.x,
-                                                       bins=np.linspace(0, out_size, out_size)
-                                                       )
-                                           )['values'].min())
+        # sort idxs and weights based on weights
+        if statistic_method == 'min':
+            order = np.argsort(weights)
```
djhoese, line 216: `weights` is a dask array that is based on the input `data`. By doing `np.argsort` here you are likely computing the whole data array here and then again later when you get the final output. Can this be moved into one of the delayed functions?

pyresample/bucket/__init__.py
```diff
@@ -209,27 +208,47 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
+        bins = np.linspace(0, out_size-1, out_size).astype('int')
+        chunk_size = len(bins)//int(1e3)
+        chunk_size = len(bins) if chunk_size == 0 else chunk_size
+
+        # preallocate buffers to avoid array creation in main loop
+        mask_buffer = np.empty((chunk_size, len(self.idxs) + 1), dtype=bool)
+        mask_buffer[:, -1] = True
+
+        # calculate the statistics in each block
+        blocks = da.from_array(bins, chunks=chunk_size).to_delayed()
```
djhoese, line 243: I don't think you need `.to_delayed` here. If it is a dask array then dask will figure out what it needs to do to pass it to a delayed function.

djhoese: Oh also you are iterating over `blocks` in the next block of code so you are creating a dask array and then immediately computing it so it can be iterated over, right? `for block in blocks`

pyresample/bucket/__init__.py
djhoese: Why would it lead to a different result in each run? Can you move the creation of `mask_buffer` into this function? Oh wait, I see the `for block in blocks` now. This isn't really what I had in mind. If you are already doing delayed functions then you should probably just put this whole computation into a delayed function. As I mentioned before this completely removes the chunked/parallel nature of dask, but if it works then that's at least a step in the right direction.

I had said that you need to separate the algorithm into the different components: the parts that operate on per-dask-chunk and the ones that need to operate on the whole array. This `mask_buffer` concept needs to operate on the whole array so the operations using it need to all be in the same function.

zxdawn:
@djhoese The purpose of Iterating the blocks is to deal with the memory error problem.

Now, I moved them into a delayed function and get the memory like this now:
```
    mask_buffer = (bins[:, None] <= idxs_sorted[None, :-1]) &\
numpy.core._exceptions.MemoryError: Unable to allocate 12.8 TiB for an array with shape (3750000, 3749999) and data type bool
```

djhoese:
What size data are you running this with (source size, target area size)? If the algorithm really is trying to take this much memory then it needs to be rewritten for full dask chunking. I'm sorry I don't have the time to help with this more. It sounds interesting and there is likely something that can be done...I just don't have the time.

zxdawn:
@djhoese The source and target area have the same size (ABI CONUS 2 km): 1500*2500 = 3750000.
I'm using the following code to resample the corrected GLM x, y to ABI area and get the min/max in the ABI area:
```
            corr_data = xr.DataArray(resampler_glm.get_min(scn_glm[channel].isel(time=i)),
                                     dims=['y', 'x'],
                                     coords=[abi_y, abi_x])
```

zxdawn:
@djhoese @pnuu Finally, I find the quickest method !!! Combing `np.digitize()` and `np.unique()` can avoid the memory error caused by the large 2D maks array.

djhoese:
So good news, I guess. I can reproduce the error after I updated all my packages. My guess is this is a change in dask since they had a release a couple days ago. No idea yet if this something I'm doing wrong or a bug in dask.

djhoese:
See https://github.com/pytroll/pyresample/pull/370 for test failures

zxdawn:
Finally, it almost passed the test. Could @gerritholl test your data? I believe it should accelerate #418 a lot!

pyresample/bucket/__init__.py
```diff
@@ -208,27 +237,18 @@ def _call_pandas_groupby_statistics(self, scipy_method, data, fill_value=None, s
...
-        df.columns = ['x', 'values']
-
-        if scipy_method == 'min':
-            statistics = df.map_partitions(lambda part: part.groupby(
-                                           np.digitize(part.x,
-                                                       bins=np.linspace(0, out_size, out_size)
-                                                       )
-                                           )['values'].min())
+        # sort idxs and weights
+        order = da.from_delayed(self._sort_weights(statistic_method, weights),
```
djhoese, line 241: I would move `_sort_weights` outside of the class. Dask is likely having to serialize the resampler object when this function is called. You might also want to look at setting `pure=True` when you create the delayed version of `_sort_weights`. I forget if that is the default for a delayed function.

gerritholl:
What still remains to be done for this PR?

gerritholl:
For a test script using the bucket resampler directly:

with `get_max`:

image

with `get_abs_max` (using #418):

image

pyresample/bucket/__init__.py
```diff
@@ -189,63 +223,40 @@ def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic):
...
         return statistics.reshape(self.target_area.shape)
 
     def get_min(self, data, fill_value=np.nan, skipna=True):
         """Calculate minimums for each bin with drop-in-a-bucket resampling.
 
-        .. warning::
+        .. note::
 
-            The slow :meth:`pandas.DataFrame.groupby` method is temporarily used here,
+            The :meth:`numpy.ufunc.reduceat` method is used here,
```
pnuu, line 259: It seems that `numpy.ufunc.reduceat` isn't used anymore, so the note can be removed.

pyresample/bucket/__init__.py
```diff
@@ -266,14 +277,14 @@ def get_min(self, data, fill_value=np.nan, skipna=True):
...
+        return self._call_bin_statistic('min', data, fill_value, skipna)
 
     def get_max(self, data, fill_value=np.nan, skipna=True):
         """Calculate maximums for each bin with drop-in-a-bucket resampling.
 
-        .. warning::
+        .. note::
 
-            The slow :meth:`pandas.DataFrame.groupby` method is temporarily used here,
+            The :meth:`numpy.ufunc.reduceat`  method is temporarily used here,
```
pnuu, line 287: It seems that `numpy.ufunc.reduceat` isn't used anymore, so the note can be removed.

pnuu:
Should this be converted to a proper PR, or is there still something that need to be addressed?

pyresample/bucket/__init__.py
gerritholl: Fixed.

gerritholl:
> Should this be converted to a proper PR, or is there still something that need to be addressed?

I thought there was, but I was wrong.  Marked as ready again.

pyresample/bucket/__init__.py
```diff
@@ -28,6 +29,39 @@
...
 
+@dask.delayed(pure=True)
+def _sort_weights(statistic_method, weights):
+    """Sort idxs and weights based on weights."""
+    if statistic_method == 'min':
+        order = np.argsort(weights)
+    elif statistic_method == 'max':
+        order = np.argsort(weights)[::-1]
+
+    return order
```
pnuu, line 40:
```suggestion
def _sort_weights(statistic_method, weights):
    """Sort idxs and weights based on weights.
    
    By default the method for sorting is `'min'`.
    """
    order = np.argsort(weights)
    if statistic_method == 'max':
        return order[::-1]
    return order
```

With this version undefined method won't crash in a funky way.

pyresample/bucket/__init__.py
```diff
@@ -28,6 +29,39 @@
...
+    """Get the statistic of each bin."""
+    # get where the `idxs_sorted` located in `bins`
+    binned_values = np.digitize(idxs_sorted, bins, right=True)
+
+    # mask value outside of bin
+    binned_values_masked = np.ma.masked_array(binned_values,
+                                              (idxs_sorted < bins.min()) | (idxs_sorted > bins.max()))
+
+    # get the first index and value in each bin
+    unique_bin, unique_idx = np.unique(binned_values_masked, return_index=True)
```
pnuu, line 54: These could be grouped together in `_find_unique_bins_and_indices()` function, and I think the comments could be removed.

pyresample/bucket/__init__.py
```diff
@@ -28,6 +29,39 @@
...
+
+    # create the full index array
+    weight_idx = np.full(len(bins), -1, dtype='int32')
+
+    # assign the valid index to array
+    weight_idx[unique_bin[~unique_bin.mask].data] = unique_idx[~unique_bin.mask]
+
+    return weights_sorted[weight_idx]  # last value of weigths_sorted always nan
```
pnuu, line 62: These could be grouped together in `_expand_bin_statistics()` function, and the comments removed.

pyresample/bucket/__init__.py
```diff
@@ -189,65 +223,37 @@ def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic):
...
-                                           )['values'].max())
-
-        # fill missed index
-        statistics = (statistics + pd.Series(np.zeros(out_size))).fillna(0)
+        # sort idxs and data
+        order = da.from_delayed(_sort_weights(statistic_method, data),
+                                shape=(len(data), ),
+                                dtype='int')
+        idxs_sorted = self.idxs[order]
+        data_sorted = np.append(data[order], np.nan)
```
pnuu, line 244: This could be a new method `_get_sorted_indices_and_data(self, statistic_method, data)`.

pyresample/bucket/__init__.py
```diff
@@ -189,65 +223,37 @@ def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic):
...
-        # Remove NaN values from the data when used as weights
-        weights = da.where(np.isnan(data), 0, data)
-
         # Rechunk indices to match the data chunking
-        if weights.chunks != self.idxs.chunks:
-            self.idxs = da.rechunk(self.idxs, weights.chunks)
+        if data.chunks != self.idxs.chunks:
+            self.idxs = da.rechunk(self.idxs, data.chunks)
 
         # Calculate the min of the data falling to each bin
```
pnuu, line 236: I think this obsolete(?) comment could be removed.

pnuu (commented):
Some suggestions inline.

pyresample/bucket/__init__.py
gerritholl: Done in 1aa2779.

pyresample/bucket/__init__.py
gerritholl: Done in 1aa2779.

pyresample/bucket/__init__.py
gerritholl: Done in 1aa2779.

pyresample/bucket/__init__.py
gerritholl: Removed in 1aa2779.

pyresample/bucket/__init__.py
```diff
@@ -28,6 +29,40 @@
 LOG = logging.getLogger(__name__)
 
 
+@dask.delayed(pure=True)
+def _sort_weights(statistic_method, weights):
+    """Sort idxs and weights based on weights.
+
```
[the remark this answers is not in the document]
gerritholl: Fixed in 1aa2779.

djhoese:
@gerritholl @zxdawn What is the current state of this PR as far as performance? Are there still some things you can't explain or you expect or want to be faster? Or do you think you've caught all the edges cases and it is as fast as it can reasonably be?

gerritholl:
I can't promise we've caught all edge cases and it's as fast as it can possibly be, but as far as I'm concerned we've made very good progress.

pyresample/bucket/__init__.py
```diff
@@ -189,65 +234,40 @@ def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic):
...
-                                           )['values'].min())
+        (idxs_sorted, data_sorted) = self._get_sorted_indices_and_data(statistic_method, data)
 
-        elif scipy_method == 'max':
-            statistics = df.map_partitions(lambda part: part.groupby(
-                                           np.digitize(part.x,
-                                                       bins=np.linspace(0, out_size, out_size)
-                                                       )
-                                           )['values'].max())
+        bins = np.linspace(0, out_size - 1, out_size).astype('int')
```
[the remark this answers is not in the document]
djhoese:
```suggestion
        bins = np.linspace(0, out_size - 1, out_size, dtype=np.int64)
```

djhoese (changes requested):
Looks pretty nice. Good job.

I had a couple small suggestions and then I have one big one that I would like tested out either in this PR or a promise that it will be done in a future PR. Right now the general flow of the dask arrays goes like this:

```
data -> delayed(_sort_weights) -> indexing and linspace -> delayed(_get_bin_statistics) -> reshape
```

To me there isn't much benefit to having that middle indexing and linspace section of code. The only benefit I see is if the indexing is much faster when done in parallel with dask, but since the indexes are coming from a delayed function (no real chunking) and it is then fed into the second delayed function right away. I think all this logic should be one giant delayed function. It'd be nice if it could be done with `map_blocks` but I don't see a way to do that. The only reason not to do this would be if combining all of these resulted in a large memory usage because all of the temporary arrays (`del idxs_sorted` might help with that).

Lastly, while looking at how `self.idxs` is generated, I think that entire function (`_get_indices`) should be turned into a map_blocks call. There are a lot of little intermediate operations there that are just adding more overhead to the dask graph. However, now that I say that I see that `y_idxs`, `x_idxs`, and `idxs` are all assigned to `self` in this method so maybe it isn't as pretty as I had hoped. Wait! No, `y_idxs` and `x_idxs` are never used outside this method, right? Having a `map_blocks` function that took `lons` and `lats` as inputs and returned the equivalent of `self.idxs` should clean up the dask graph a bit and I would hope improve memory usage and maybe execution time. Bare minimum the dask graph would look better.

gerritholl:
This is currently the dask graph:

br-max-dask-graph

pyresample/bucket/__init__.py
```diff
@@ -189,65 +234,40 @@ def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic):
...
+                                     shape=(len(bins),),
+                                     dtype=np.float64)
 
         return statistics.reshape(self.target_area.shape)
 
+    def _get_sorted_indices_and_data(self, statistic_method, data):
+        # sort idxs and data
+        order = da.from_delayed(_sort_weights(statistic_method, data),
+                                shape=(len(data), ),
+                                dtype='int')
```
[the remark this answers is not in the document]
gerritholl: Fixed in 1b21364

gerritholl:
> No, `y_idxs` and `x_idxs` are never used outside this method, right?

They're not used within this class, but they're exposed as public API, so removing them would affect backward compatibility.  @pnuu are `x_idxs` and `y_idxs` needed as public attributes?

pnuu:
No, I don't think `x_idxs` and `y_idxs` are needed outside of the resampler.

gerritholl:
Simplifying to only one `dask.delayed`-function as suggested by @djhoese speeds up a test `get_max()` call from 9 seconds with max 9 GB RAM to 5 seconds with max 4.5 GB RAM.  See 57fa302.

gerritholl:
I tried if bringing the `.reshape(...)` call inside the `@dask.delayed`-decorated function would resolve this, but it just causes another task (`_get_statistics`) to take over the role of dominating the calculation time.  I suppose I don't _quite_ understand what I'm doing here.

gerritholl:
In the spirit of incremental PRs that do one thing, I'm a bit hesitant to break open the `idxs`, `x_idxs`, and `y_idxs` attributes and `_get_indices()`.  `self.idxs` is used not only in `get_max()` and `get_min()`, but also in `get_count()` and `get_sum()`, and therefore by `get_average()`.  It would broaden the scope of the PR from speeding up `get_max()` and `get_min()` to refactoring the whole `BucketResampler` class.  Therefore, I would prefer to defer changing `_get_indices()` to another PR.

djhoese:
I'm fine with that.

PR #376, opened 2021-09-10
Update EWA ll2cr_static to handle swaths crossing the anti-meridian.

### LINKED ISSUE #374: Geographic EWA projection for swaths crossing the anti-meridian omits values for 90 ≤ longitude (degrees east) ≤ 180.

Input data swath crossing anti-meridian: [Available here](https://podaac-tools.jpl.nasa.gov/drive/files/allData/ghrsst/data/GDS2/L2P/MODIS_A/JPL/v2019.0/2021/229/20210817140001-JPL-L2P_GHRSST-SSTskin-MODIS_A-N-v02.0-fv01.0.nc), also [attached to this issue](https://github.com/pytroll/pyresample/files/7086053/20210817140001-JPL-L2P_GHRSST-SSTskin-MODIS_A-N-v02.0-fv01.0.nc.gz).

Note, this summarises code from various parts of a larger service - so is a little rough, particularly with handling the fill value.

```python
from pyresample import AreaDefinition, SwathDefinition
from pyresample.ewa import fornav, ll2cr
from pyresample.utils import check_and_wrap
from netCDF4 import Dataset
import numpy as np

INPUT_FILE_PATH = '20210817140001-JPL-L2P_GHRSST-SSTskin-MODIS_A-N-v02.0-fv01.0.nc'
TARGET_DIMENSIONS = (433, 7265)
TARGET_EXTENTS = (-179.992447, 31.104006, 179.997894, 52.569942)


with Dataset(INPUT_FILE_PATH, 'r') as input_dataset:
    # Get fill value
    unscaled_fill_value = variable.getncattr('_FillValue')
    scale_factor = variable.getncattr('scale_factor')
    add_offset = variable.getncattr('add_offset')
    variable_fill_value = (unscaled_fill_value * scale_factor) + add_offset

    variable_data = input_dataset['sea_surface_temperature'][0][:].filled(
        fill_value=variable_fill_value
    )
    input_swath = SwathDefinition(lats=wrapped_lats, lons=wrapped_lons)
    target_grid = AreaDefinition.from_extent('lat, lon', 'EPSG:4326',
                                             TARGET_DIMENSIONS,
                                             TARGET_EXTENTS)

    ewa_info = ll2cr(input_swath, target_grid)
    _, results = fornav(ewa_info[1], ewa_info[2], target_grid,
                        variable_data, maximum_weight_mode=False)

    np.nan_to_num(results, nan=variable_fill_value, copy=False)
```
#### Problem description

When taking an input swath that crosses the anti-meridian and reformatting it to a whole-Earth geographic grid, the EWA algorithm does not include data where 90 < longitude (degrees east) < 180. (We've confirmed that this behaviour is limited to  the EWA algorithm, other interpolation methods work nicely).

My colleagues and I have looked into this a bit, and think the issue may be in [`ll2cr_static`](https://github.com/pytroll/pyresample/blob/main/pyresample/ewa/_ll2cr.pyx#L255). There's a check for anti-meridian crossing. If the pixel is determined to cross the anti-meridian, the x coordinate (longitude in this case) is incremented by the projection circumference (360 degrees for geographic):

```python
elif proj_circum != 0 and abs(x_tmp - origin_x) >= (0.75 * proj_circum):
    # if x is more than 75% around the projection space, it is probably crossing the anti-meridian
    x_tmp += proj_circum
```

In our case, we have a whole-Earth output grid, so `origin_x = -180` and `proj_circum = 360.0`. The anti-meridian crossing condition for these values is satisfied for `longitude` ≥ 90 (degrees east). In the case of `longitude` = 90 degrees east, this shift means `longitude_incremented` = 450 degrees east. This is likely leading to the affected longitudes being considered outside of the target grid.

Our team was wondering if this could be mitigated by an additional check to the `elif` condition that ensures the x coordinate is less than the lefthand edge of the target grid: `x_tmp < origin_x`. Alternatively, maybe the projection could be checked to see if the target grid is geographic? (Or maybe a combination of both checks)

#### Expected Output

This is the input used in the sample code above. The output should resemble the input, but be geographically gridded to the target area:

#### Actual Result

The output below shows the non-fill pixels once the `sea_surface_temperature` has been projected. The output target area is essentially whole-Earth, but the pixels west of the anti-meridian are filled.

| Panoply georeferenced image | Gridded contour plot using target grid |
|---|---|
|  |  |

Tested under `pyresample` v1.17.0 and v1.20.0 in Python 3.7.

owenlittlejohns:
This PR removes an `elif` condition in the `ll2cr_static` method, which would identify when source pixels were at least three quarters of the circumference of the target projection away from the target grid origin x coordinate. Data that satisfied this condition were then incremented by the projection circumference.

For swaths that cross the anti-meridian, this meant data with longitudes greater than three quarters of the target projection circumference were being shifted out of the target grid, and therefore the data were being omitted.

 - [x] Closes #374 
 - [x] Tests added
 - [x] Tests passed

### DIFF
--- a/pyresample/test/test_ewa_ll2cr.py
+++ b/pyresample/test/test_ewa_ll2cr.py
@@ -50,6 +50,17 @@ static_lcc = {
     "proj4_definition": "+proj=lcc +a=6371200 +b=6371200 +lat_0=25 +lat_1=25 +lon_0=-95 +units=m +no_defs",
 }
 
+static_geo_whole_earth = {
+    "grid_name": "test_geo",
+    "origin_x": -180.0,
+    "origin_y": 40,
+    "width": 361,
+    "height": 181,
+    "cell_width": 1.0,
+    "cell_height": 1.0,
+    "proj4_definition": "+proj=longlat +datum=WGS84 +no_defs +type=crs",
+}
+
 
 class TestLL2CRStatic(unittest.TestCase):
     def test_lcc_basic1(self):
@@ -66,8 +77,31 @@ class TestLL2CRStatic(unittest.TestCase):
         w = grid_info["width"]
         h = grid_info["height"]
         points_in_grid = _ll2cr.ll2cr_static(lon_arr, lat_arr, fill_in, proj_str,
-                                                               cw, ch, w, h, ox, oy)
+                                             cw, ch, w, h, ox, oy)
         self.assertEqual(points_in_grid, lon_arr.size, "all these test points should fall in this grid")
+
+    def test_geo_antimeridian(self):
+        """ Ensure output for anti-meridian crossing input includes all points. """
+        from pyresample.ewa import _ll2cr
+        lon_arr = create_test_longitude(160.0, 200.0, (50, 100), dtype=np.float64)
+        lat_arr = create_test_latitude(40.0, 60.0, (50, 100), dtype=np.float64)
+
+        # wrap values in lon_arr so -180 ≤ longitude (degrees east) ≤ 180
+        lon_arr[lon_arr > 180] -= 360.0
+
+        grid_info = static_geo_whole_earth.copy()
+        fill_in = np.nan
+        proj_str = grid_info['proj4_definition']
+        cw = grid_info['cell_width']
+        ch = grid_info['cell_height']
+        ox = grid_info['origin_x']
+        oy = grid_info['origin_y']
+        w = grid_info['width']
+        h = grid_info['height']
+        points_in_grid = _ll2cr.ll2cr_static(lon_arr, lat_arr, fill_in, proj_str,
+                                             cw, ch, w, h, ox, oy)
+        self.assertEqual(points_in_grid, lon_arr.size,
+                         'all these test points should fall in this grid')
 
     def test_lcc_fail1(self):
         from pyresample.ewa import _ll2cr
--- a/pyresample/ewa/_ll2cr.pyx
+++ b/pyresample/ewa/_ll2cr.pyx
@@ -232,7 +232,6 @@ def ll2cr_static(numpy.ndarray[cr_dtype, ndim=2] lon_arr, numpy.ndarray[cr_dtype
     cdef tuple projected_tuple = p(lon_arr, lat_arr)
     cdef cr_dtype [:, ::1] rows_out = projected_tuple[1]
     cdef cr_dtype [:, ::1] cols_out = projected_tuple[0]
-    cdef double proj_circum = projection_circumference(p)
 
     # indexes
     cdef unsigned int row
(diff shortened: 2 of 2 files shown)


pyresample/test/test_ewa_ll2cr.py
```diff
@@ -66,9 +77,33 @@ def test_lcc_basic1(self):
...
 
+    def test_geo_antimeridian(self):
+        """ Ensure output for anti-meridian crossing input includes all points. """
+        from pyresample.ewa import _ll2cr
+        lon_arr = create_test_longitude(160.0, 200.0, (50, 100), dtype=np.float64)
+        lat_arr = create_test_latitude(40.0, 60.0, (50, 100), dtype=np.float64)
+
+        # wrap values in lon_arr so -180 ≤ longitude (degrees east) ≤ 180
+        lon_arr[lon_arr > 180] -= 360.0
+        print(lon_arr[0])
```
djhoese, line 91: Left over print statement. Could you remove this?

djhoese:
I keep going back to this PR trying to figure out why the static ll2cr had this logic in it. I'm pretty sure this is some feature I was using in my old Polar2Grid software that was, at the end of the day, going against what users should be doing *if* an area(grid) is defined properly. I think we had a lot of cases where users like those in Alaska had data going over the anti-meridian and wanted to use a single global (-180 to 180) Area to resample to. If the data went over the anti-meridian they didn't want one huge global area, they only wanted the small part of the image that had the contiguous data. However, this really only makes sense for the "dynamic" ll2cr grids, not the static ones so... :shrug:

pyresample/test/test_ewa_ll2cr.py
owenlittlejohns: D'oh! I've removed that now. Thanks.

owenlittlejohns:
Yeah - the Alaska use-case is exactly how we originally found this issue. Our service is basically doing what you outlined: making a big old global grid, that has a _lot_ of empty space. It'd be nice to have a more truncated grid, but on the other hand, it's also nice not having a longitude discontinuity in the middle of the data. Tricky stuff!

djhoese:
It should be possible (having tested it recently) to play with `+pm=180` in your PROJ.4 string definition. This basically makes the longitude space go from 0 to 360 instead of -180 to 180. With that you should be able to define your extents from something like 160 to 220 longitude. So while the data may cross the anti-meridian of the normal -180/180 space, it should be contiguous in this projection. The major downside though is that your geotiff (or other) image viewer has to support this PROJ parameter.

...or I guess you could make two separate images one of the "left" side and one for the "right" and then put both...oh discontinuities :facepalm: 

I'll plan on merging this tomorrow. I don't trust myself merging things late at night.

djhoese:
Thanks again for this. I'll see if I can make a bug fix release to include this some time today.

owenlittlejohns:
Just tried out v1.21.1. Works a treat. Thanks for the quick turnaround on the merge and deploy!

PR #379, opened 2021-09-24
Define base Resampler class

djhoese:
This PR depends on #369. This updates the base `Resampler` class so it serves more use cases, but most importantly it starts laying out the testing framework so that we can throw all sorts of cases at Pyresample 2.0 interfaces and know that things are going to work.

 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented


djhoese:
:facepalm: My isort PR contains the change to say "don't automatically fix things".

PR #391, opened 2021-11-05
Clarify SphPolygon.area documentation

### LINKED ISSUE #390: What units does SphPolygon.area return?
The units returned by `SphPolygon.area` are not documented and not obvious.

In this MCVE, I construct a quadrangle near `(0, 0)`  with an edge of 0.001 degrees.  At such a small area, we can approximate the Earth as flat, so I can compare the polygon area calculated with pyresample against a multiplication of the lengths of the edges of the quadrangle.  However, the results are very different.

```python
from pyproj import Geod
from numpy import deg2rad, array
from pyresample.spherical import SphPolygon
from math import tau
x = 0.001
pol = SphPolygon(deg2rad(array([[0., 0.], [0., x], [x, x], [x, 0.]])))
g = Geod(ellps="sphere")
print(pol.area()*1.5e8/(2*tau))
print((g.inv(0, 0, x, 0)[2] * g.inv(0, 0, 0, x)[2]) / 1e6)
```
#### Problem description

I don't know the units of the result.  The units are not documented at https://pyresample.readthedocs.io/en/latest/api/pyresample.html?highlight=SphPolygon#pyresample.spherical.SphPolygon.area.  It's clearly not km².  I thought it might be steradian, but multiplying by the surface of the Earth then dividing by 4π (or 2τ) doesn't yield a consistent result either.

#### Expected Output

I expect an output that I can interpret with documented units.  The surface area of the indicated polygon should be in the order of 0.012 km² (Euclidian approximation), but I can't find an explicable conversion factor between the two numbers.

0.003636094926640598
0.01236430006718607

Pyresample 1.22 on Python 3.9.7.

gerritholl:
Clarify the documentation for SphPolygon.area.

 - [x] Closes #390 
 - [x] Fully documented

### DIFF
--- a/pyresample/spherical.py
+++ b/pyresample/spherical.py
@@ -340,7 +340,7 @@ class SphPolygon:
 
         Args:
             vertices (np.ndarray): ndarray of shape ``[N, 2]`` with ``N``
-                points describing a polygon counterclockwise.  First column
+                points describing a polygon clockwise.  First column
                 describes longitudes, second column describes latitudes.  Units
                 should be in radians.
             radius (optional, number): Radius of spherical planet.
--- a/pyresample/spherical.py
+++ b/pyresample/spherical.py
@@ -321,15 +321,30 @@ class Arc(object):
         return None, None
 
 
-class SphPolygon(object):
+class SphPolygon:
     """Spherical polygon.
 
-    Vertices as a 2-column array of (col 1) lons and (col 2) lats is
-    given in radians. The inside of the polygon is defined by the
-    vertices being defined clockwise around it.
+    Represents a polygon on a spherical geoid.  Initialise with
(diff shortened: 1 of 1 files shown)


pyresample/spherical.py
```diff
@@ -321,15 +321,30 @@ def dist(args):
...
+    spherical geoid on which calculations occur.
+
     """
 
     def __init__(self, vertices, radius=1):
+        """Initialise SphPolygon object.
+
+        Args:
+            vertices (np.ndarray): ndarray of shape ``[N, 2]`` with ``N``
+                points describing a polygon counterclockwise.  First column
```
djhoese, line 343:
```suggestion
                points describing a polygon clockwise.  First column
```

djhoese: Or maybe "along a clockwise path"?

PR #395, opened 2021-11-11
Replace depracated Numpy dtypes

pnuu:
With a new production chain I noticed the following depraction warning in the logs:

```python
lib/python3.9/site-packages/pyresample/kd_tree.py:968: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  radius=self.radius_of_influence, dtype=np.int,
```

This PR replaces all the usages of `np.int`, `np.float` and `np.bool` with the preferred versions listed in the third column of the Release Notes for 1.20.0.

 - [x] Tests passed 

### DIFF
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -230,7 +230,7 @@ class BucketResampler(object):
         # fill missed index
         statistics = (statistics + pd.Series(np.zeros(out_size))).fillna(0)
 
-        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int)).ravel()
+        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(np.int64)).ravel()
 
         # TODO remove following line in favour of weights = data when dask histogram bug (issue #6935) is fixed
         statistics = self._mask_bins_with_nan_if_not_skipna(skipna, data, out_size, statistics)
@@ -346,7 +346,7 @@ class BucketResampler(object):
             data = da.where(data == fill_value, np.nan, data)
 
         sums = self.get_sum(data, skipna=skipna)
-        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(int))
+        counts = self.get_sum(np.logical_not(np.isnan(data)).astype(np.int64))
 
         average = sums / da.where(counts == 0, np.nan, counts)
         average = da.where(np.isnan(average), fill_value, average)
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -126,8 +126,8 @@ class BucketResampler(object):
         # Calculate array indices. Orient so that 0-meridian is pointing down.
         adef = self.target_area
         x_res, y_res = adef.resolution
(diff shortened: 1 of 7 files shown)


djhoese:
Do we just want to force this to int64 to ensure 64-bit precision?

djhoese:
I also think there is an np.float somewhere in `kd_tree.py` somewhere but when I switched it to float64 it started failing...I think.

pnuu:
I changed them to `np.int64` and `np.float64` to be explicit. And swapped `np.bool` to `np.bool_` to be consistent with the others.

Apparently `pytest-lazy-fixture` was missing from the `test_requires` as I couldn't run the tests locally. As the tests ran on CI, I assume it was explicitly installed there.

pnuu:
The tests don't pass locally on Python 3.9, which doesn't seem to be use in the CI at all!

pnuu:
Python 3.9 test runs added in https://github.com/pytroll/pyresample/pull/396

pyresample/bucket/__init__.py
```diff
@@ -126,8 +126,8 @@ def _get_indices(self):
         # Calculate array indices. Orient so that 0-meridian is pointing down.
         adef = self.target_area
         x_res, y_res = adef.resolution
-        x_idxs = da.floor((proj_x - adef.area_extent[0]) / x_res).astype(np.int)
-        y_idxs = da.floor((adef.area_extent[3] - proj_y) / y_res).astype(np.int)
+        x_idxs = da.floor((proj_x - adef.area_extent[0]) / x_res).astype(int)
+        y_idxs = da.floor((adef.area_extent[3] - proj_y) / y_res).astype(int)
```
djhoese, line 130: Do you want these to be `np.int64` too?

setup.py
```diff
@@ -46,7 +46,8 @@
                   'tests': test_requires}
 
 setup_requires = ['numpy>=1.10.0', 'cython']
-test_requires = ['rasterio', 'dask', 'xarray', 'cartopy>=0.20.0', 'pillow', 'matplotlib', 'scipy', 'zarr']
+test_requires = ['rasterio', 'dask', 'xarray', 'cartopy>=0.20.0', 'pillow', 'matplotlib', 'scipy', 'zarr',
+                 'pytest-lazy-fixture']
```
djhoese, line 50: Wow I'm really surprised I missed this. I thought I was careful. Thanks for fixing it.

djhoese (approved):
Small double check on one remaining `int`, otherwise LGTM

pyresample/bucket/__init__.py
pnuu: Yeah. Thanks, missed them in the previous change :man_facepalming:

PR #396, opened 2021-11-12
Add Python 3.9 to CI runs and use it for the experimental run

pnuu:
Python 3.9 wasn't used to run the unit tests. This'll add it, and swap the experimental test runs to it.

 - [x] Tests passed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -41,10 +41,10 @@ jobs:
       fail-fast: true
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
-        python-version: ["3.7", "3.8"]
+        python-version: ["3.7", "3.8", "3.9"]
         experimental: [false]
         include:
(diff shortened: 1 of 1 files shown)


djhoese:
@gerritholl had brought up that numpy will drop 3.7 support by the end of the year. Should we drop 3.7 testing now?

djhoese:
Thoughts on dropping Python 3.7 @gerritholl @mraspaud @pnuu ?

gerritholl:
> Thoughts on dropping Python 3.7 in the tests

Fine with me, although if we follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) strictly, then we should wait for 2021-12-26 to do so (or for the first release after this date).

djhoese:
Ok so maybe we just merge this?

gerritholl:
I think so.  Maybe by Christmas we can add Python 3.10 tests and drop Python 3.7 at the same time.

PR #400, opened 2021-11-29
Move bilinear import to avoid unnecessary warning

### LINKED ISSUE #375: Importing pyresample without having Xarray and/or zarray raises UserWarning

```python
import pyresample
```
```
test_resample_import.py:10: UserWarning: XArray and/or zarr not found, XArrayBilinearResampler won't be available.
  warnings.warn("XArray and/or zarr not found, XArrayBilinearResampler won't be available.")
```

#### Problem description

When importing pyresample without having XArray installed the above warning is thrown. I consciously have not installed XArray with this package since it is not on a major version yet. However this means that every time I run the pyresample import for my code I see this warning at the top of the stdout. If pyresample requires XArray to be importable should Xarray be included as a dependency (I noticed that it is included on the "xarray_bilinear" extras)?

A possible solution to this without requiring XArray as a dependency would be to include the ImportWarning category on the try/except block in the pyresample/bilinear/__init__.py file.
```
    warnings.warn("XArray and/or zarr not found, XArrayBilinearResampler won't be available.", category=ImportWarning)
```
This would still log the warning but only display it in Python Development Mode.

I think that there should be error handling when the user tries to use the XArrayBilinearResampler without having Xarray installed, but not on the import of the package. Since it is still unknown at import if the user will be using XArrayBilinearResampler or just NumpyBilinearResampler. 

I currently have a hotfix by surrounding the import statment with a catch_warnings() block,
```
with warnings.catch_warnings():
    warnings.simplefilter('ignore', UserWarning)
    import pyresample
```

I am currently testing this on python 3.8 with the current base requirements: setuptools>=3.2', 'pyproj>=2.2', 'configobj','pykdtree>=1.3.1', 'pyyaml', 'numpy>=1.10.0',

Thanks for your consideration!

yunjunz:
This PR moves the `bilinear` module import from the top of `image.py` to the function body of `ImageContainerBilinear.resample()`. 

`conda install -c conda-forge pyresample` does not install `xarray` and `zarr`. Thus, if one did not install `xarray` and `zarr` manually, running `import pyresample` will print out the following warning message:

```bash
(test) yunjunz:~>$ python -c "import pyresample"
/Users/yunjunz/tools/miniconda3/envs/test/lib/python3.10/site-packages/pyresample/bilinear/__init__.py:50: UserWarning: XArray and/or zarr not found, XArrayBilinearResampler won't be available.
  warnings.warn("XArray and/or zarr not found, XArrayBilinearResampler won't be available.")
```

With this change, the top-level `import pyresample`, which calls `from pyresample import image`, won't call the `bilinear` module, thus, avoid the warning msg.

 - [x] Closes #375  
 - [x] Tests passed 

### DIFF
--- a/pyresample/image.py
+++ b/pyresample/image.py
@@ -23,7 +23,7 @@ import warnings
 
 import numpy as np
 
-from pyresample import bilinear, geometry, grid, kd_tree
+from pyresample import geometry, grid, kd_tree
 
 
 class ImageContainer(object):
@@ -368,6 +368,10 @@ class ImageContainerBilinear(ImageContainer):
         image_container : object
             ImageContainerBilinear object of resampled geometry
         """
+        # import here, instead of the top of the script, to avoid xarray/zarr warning msg
+        # while import the top level pyresample module
+        from pyresample import bilinear
+
         image_data = self.image_data
 
         try:


djhoese:
Looks good. Thank you very much for clearing this up.

PR #404, opened 2021-12-07
Fix dask EWA code not creating unique dask task names for different target areas

djhoese:
I've always tested the new algorithm with only one target, but once it was thrown into Polar2Grid we quickly realized that multiple targets was a problem. Luckily because of my map_blocks PR last week I knew to look at the dask task names for the issue.

 - [x] Tests added 
 - [x] Tests passed 


mraspaud (approved):
LGTM

Issue #407, opened 2022-01-04
`SwathDef.aggregate` bugs and enhancements


ghiggi:
#### Problem description

`SwathDef.aggregate` currently works only if SwathDefinition is defined by passing lons/lats `xr.DataArray` with `lons` and `lats` dimensions and dask arrays. 

#### Code Sample

```python
import numpy as np
import dask.array as da
import xarray as xr
from pyresample.geometry import SwathDefinition

lons = np.arange(-179.75, 179.75 + 0.5, 0.5)
lats = np.arange(-89.75,  0.5,  0.5)
lons, lats = np.meshgrid(lons, lats)

# BUG: This does not work. It expects an xr.DataArray  
swath_def = SwathDefinition(lons=lons, lats=lats)
swath_def.aggregate(x=2, y=1)      # Downsample by factor of 2 along x

# BUG: This does not work. It expects a chunked dask array.
lons_da = xr.DataArray(lons, dims=("lats","lons"))
lats_da = xr.DataArray(lats, dims=("lats","lons"))
swath_def = SwathDefinition(lons=lons_da, lats=lats_da)
swath_def.aggregate(x=2, y=1)      # Downsample by factor of 2 along x
 
# This works to coarse the data
lons = da.from_array(lons)
lats = da.from_array(lats) 
lons_da = xr.DataArray(lons, dims=("lats","lons"))
lats_da = xr.DataArray(lats, dims=("lats","lons"))
swath_def = SwathDefinition(lons=lons_da, lats=lats_da)
swath_def.aggregate(x=2, y=1)      # Downsample by factor of 2 along x

```

#### Limitations & Consistency

`AreaDef.aggregate` enables to also upsampling (downscaling) the area by providing `x`, `y`  dimensions < 1 
```python
swath_def.aggregate(x=1/2, y=1/2)  # Split each pixel in 4 pixels
```
In contrast, `SwathDef.aggregate` is able to just coarse the grid by averaging geocentric x, y, z coordinates, as dask.array coarsen accepts only accepts only  `x`, `y` integers >= 1.

```python
swath_def.aggregate(x=1/2, y=1)   
swath_def.aggregate(x=1/2, y=1)   

```

#### Suggested improvements 
- [ ] Enable ` swath_def.aggregate`  to work also with lons/lats defined in numpy arrays 
- [ ] Add method `area_def.upsample`  (straighforward)
- [ ] Add method `swath_def.upsample` 
[...]

PR #413, opened 2022-01-18
Add initial methods for RW

ghiggi:
This PR aims to provide the methods to aggregate/downsample, upsample, extend and reduce AreaDefinition and SwathDefinition objects.
For SwathDefinition objects, it returns a new objects with the same inputs lats/lons type (numpy/dask/xr.DataArray).

 - [ ] Closes 
    - [ ] #407
    - [ ] #28   
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented 

Yet to be done/defined: 
- [ ] Do we define also a `downsample` method and add a deprecation warning to `aggregate`?
- [ ] Any reason to still use `pyproj.transform` instead of the more efficient `pyproj.Transformer` in `SwathDefinition._do_transform`?
- [ ] For `SwathDefinition.extend`, do we define which pyproj.Geod? `pyproj.Geod(ellps='sphere')`  or `pyproj.Geod(ellps='WGS84') `?
- [ ] `SwathDefinition.upsample` could be made dask-compatible but I need some hint on HOWTO do it correctly. See code ...
- [ ] `SwathDefinition.upsample` could also use instead `geotiepoints.GeoInterpolator` but I noticed a strange behavior. See next comment for a reproducible example ... 

PS: Is my first PR to a package, so please be patient and don't hesitate to provide a lot of feedback/reproaches :)

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -715,39 +715,310 @@ class SwathDefinition(CoordinateDefinition):
     def _do_transform(src, dst, lons, lats, alt):
         """Run pyproj.transform and stack the results."""
         x, y, z = transform(src, dst, lons, lats, alt)
-        return np.dstack((x, y, z))
+        return np.dstack((x, y, z))   
+    
+    # TODO: this is more efficient than pyproj.transform.  
+    # With pyproj > 3.1 it became thread safe: https://pyproj4.github.io/pyproj/stable/advanced_examples.html#multithreading
+    # @staticmethod
+    # def swath_do_transform(src, dst, lons, lats, alt):
+    #     """Run pyproj Transformer and stack the results."""
+    #     from pyproj import Transformer
+    #     transformer = Transformer.from_crs(src.crs, dst.crs)
+    #     x, y, z = transformer.transform(lons, lats, alt, radians=False)  
+    #     return np.dstack((x, y, z))
 
-    def aggregate(self, **dims):
-        """Aggregate the current swath definition by averaging.
-
-        For example, averaging over 2x2 windows:
-        `sd.aggregate(x=2, y=2)`
+    def aggregate(self, x=1, y=1, **kwargs):
+        """Downsample the swath definition by averaging the coordinates along x and y dimension.
+        
+        Builds upon xarray.DataArray.coarsen function. 
+        To downsample of a factor of 2, call swath_def.aggregate(x=2, y=2)
+        swath_def.aggregate(x=1, y=1) simply returns the current swath_def. 
+        By default, it raise a ValueError if the dimension size is not a multiple of the window size.
+        This can be changed by passing boundary="trim" or boundary="pad", but behaviour within pyresample is undefined.
+        See https://xarray.pydata.org/en/stable/generated/xarray.DataArray.coarsen.html for further details.
         """
+        import xarray as xr 
         import dask.array as da
         import pyproj
-
+        
+        # Check input validity 
+        x = int(x)
+        y = int(y)
+        if x < 1 or y < 1: 
+            raise ValueError('x and y arguments must be positive integers larger or equal to 1.')
+            
+        # Return SwathDefinition if nothing to aggregate 
+        if x==1 and y==1: 
+            return self 
+        
+        # Define geodetic and geocentric projection 
         geocent = pyproj.Proj(proj='geocent')
         latlong = pyproj.Proj(proj='latlong')
+        
+        # Get xr.DataArray with dask array
+        src_lats, src_lats_format = _convert_2D_array(self.lats, to='DataArray_Dask', dims=['lats','lons'])
+        src_lons, src_lons_format = _convert_2D_array(self.lons, to='DataArray_Dask', dims=['lats','lons'])
+    
+        # Conversion to Geocentric Cartesian (x,y,z) CRS
         res = da.map_blocks(self._do_transform, latlong, geocent,
-                            self.lons.data, self.lats.data,
-                            da.zeros_like(self.lons.data), new_axis=[2],
-                            chunks=(self.lons.chunks[0], self.lons.chunks[1], 3))
-        res = DataArray(res, dims=['y', 'x', 'coord'], coords=self.lons.coords)
-        res = res.coarsen(**dims).mean()
+                            src_lons.data, 
+                            src_lats.data,
+                            da.zeros_like(src_lons), # altitude
+                            new_axis=[2],
+                            chunks=(src_lons.chunks[0], src_lons.chunks[1], 3))
+        res = xr.DataArray(res, dims=['y', 'x', 'coord'], coords=src_lons.coords)
+        
+        # Aggregating 
+        res = res.coarsen(x=x, y=y, **kwargs).mean()
+        
+        # Back-conversion to geographic CRS 
         lonlatalt = da.map_blocks(self._do_transform, geocent, latlong,
-                                  res[:, :, 0].data, res[:, :, 1].data,
-                                  res[:, :, 2].data, new_axis=[2],
+                                  res[:, :, 0].data, # x
+                                  res[:, :, 1].data, # y
+                                  res[:, :, 2].data, # z
+                                  new_axis=[2],
                                   chunks=res.data.chunks)
-        lons = DataArray(lonlatalt[:, :, 0], dims=self.lons.dims,
-                         coords=res.coords, attrs=self.lons.attrs.copy())
-        lats = DataArray(lonlatalt[:, :, 1], dims=self.lons.dims,
-                         coords=res.coords, attrs=self.lons.attrs.copy())
-        try:
-            resolution = lons.attrs['resolution'] * ((dims.get('x', 1) + dims.get('y', 1)) / 2)
-            lons.attrs['resolution'] = resolution
-            lats.attrs['resolution'] = resolution
-        except KeyError:
-            pass
+        
+        # Back-conversion array as input format
+        lons, _ = _convert_2D_array(lonlatalt[:, :, 0], to=src_lons_format, dims=['lats','lons'])
+        lats, _ = _convert_2D_array(lonlatalt[:, :, 1], to=src_lats_format, dims=['lats','lons'])
+        
+        # Add additional info if the source array is a DataArray
+        if isinstance(self.lats, xr.DataArray) and isinstance(self.lons, xr.DataArray):
+            lats.coords = res.coords 
+            lats.attrs = self.lats.attrs.copy()
+            lons.coords = res.coords 
+            lons.attrs = self.lons.attrs.copy()
+            try:
+                resolution = lons.attrs['resolution'] * ((x + y) / 2)  
+                lons.attrs['resolution'] = resolution
+                lats.attrs['resolution'] = resolution
+            except KeyError:
+                pass
+            
+        # Return the downsampled swath definition 
+        return SwathDefinition(lons, lats)
+    
+    def upsample(self, x=1, y=1):
+        """Upsample the swath definition along x (along-track) and y (cross-track) dimensions.
+        
+        To upsample of a factor of 2 (each pixel splitted in 4 pixels), call swath_def.upsample(x=2, y=2).
+        swath_def.upsample(x=1, y=1) simply returns the current swath_def. 
+        """
+        # TODO: An alternative would be to use geotiepoints.geointerpolator.GeoInterpolator
+        # But I have some problem using it, see code snippet in the PR description.
+        import dask.array as da
+        import xarray as xr
+        import numpy as np
+        import pyproj
+        from xarray.plot.utils import _infer_interval_breaks
+        # https://github.com/pydata/xarray/blob/main/xarray/plot/utils.py#L784
+        
+        def _upsample_ranges_1D(x, factor=1):
+            ranges2D = np.linspace(x[:-1], x[1:], num=factor, endpoint=False, axis=1)
+            return np.concatenate((ranges2D.ravel(),[x[-1]]))
+        
+        def upsample_ranges_2D(x, factor=1, axis=0):
+            x = np.array(x)
+            if x.ndim not in [1,2]: 
+                raise ValueError("Expects 1D or 2D array.")
+            if not isinstance(axis, int): 
+                 raise TypeError("'axis' must be: 0 or 1 integer.")
+            if axis not in [0, 1]: 
+                raise ValueError("Expects 'axis' 0 or 1")
+            if not isinstance(factor, int): 
+                raise TypeError("'factor' must be an integer equal or larger than 1.")
+            if factor < 1: 
+                raise ValueError("'factor' must be an integer equal or larger than 1.")
+                
+            if x.ndim == 1:
+                return _upsample_ranges_1D(x, factor=factor)
+            else:
+                l_ranges = []
+                if axis==1:
+                    for i in range(x.shape[0]):
+                        l_ranges.append(_upsample_ranges_1D(x[i,:], factor=factor))
+                    return np.vstack(l_ranges)
+                else: # axis = 0 
+                    for i in range(x.shape[1]):
+                        l_ranges.append(_upsample_ranges_1D(x[:,i], factor=factor))
+                    return np.vstack(l_ranges).transpose()
+        
+        def _upsample_corners(corners, x_factor=1, y_factor=1):
+            new_breaks_xx = upsample_ranges_2D(corners, factor=x_factor, axis=1)
+            new_corners = upsample_ranges_2D(new_breaks_xx, factor=y_factor, axis=0)
+            return new_corners
+        
+        def _get_corners_from_centroids(centroids):
+            breaks_xx = _infer_interval_breaks(centroids, axis=1)
+            corners = _infer_interval_breaks(breaks_xx, axis=0)
+            return corners
+        
+        def _get_centroids_from_corners(corners):
+            centroids = (corners[1:, 1:] + corners[:-1, :-1]) / 2
+            return centroids
+        
+        # TODO: Decide if compute in memory or with dask 
+        def upsample_centroids(centroid_x, centroid_y, centroid_z, x_factor=1, y_factor=1): 
+            corners_x = _get_corners_from_centroids(centroid_x)
+            corners_y = _get_corners_from_centroids(centroid_y)
+            corners_z = _get_corners_from_centroids(centroid_z)
+            x_new_corners = _upsample_corners(corners_x, x_factor=x_factor, y_factor=y_factor)
+            y_new_corners = _upsample_corners(corners_y, x_factor=x_factor, y_factor=y_factor)
+            z_new_corners = _upsample_corners(corners_z, x_factor=x_factor, y_factor=y_factor)
+            x_new_centroids = _get_centroids_from_corners(x_new_corners)
+            y_new_centroids = _get_centroids_from_corners(y_new_corners)
+            z_new_centroids = _get_centroids_from_corners(z_new_corners)
+            return x_new_centroids, y_new_centroids, z_new_centroids
+        
+        def _upsample_centroid(centroid, x_factor=1, y_factor=1):
+            corners = _get_corners_from_centroids(centroid)
+            new_corners = _upsample_corners(corners, x_factor=x_factor, y_factor=y_factor)
+            new_centroids = _get_centroids_from_corners(new_corners)
+            return new_centroids
+    
+        # Define geodetic and geocentric projection 
+        geocent = pyproj.Proj(proj='geocent')
+        latlong = pyproj.Proj(proj='latlong')
+        
+        # Get xr.DataArray with dask array
+        src_lats, src_lats_format = _convert_2D_array(self.lats, to='DataArray_Dask', dims=['lats','lons'])
+        src_lons, src_lons_format = _convert_2D_array(self.lons, to='DataArray_Dask', dims=['lats','lons'])
+        
+        # Conversion to Geocentric Cartesian (x,y,z) CRS
+        res = da.map_blocks(self._do_transform, latlong, geocent,
+                            src_lons.data, 
+                            src_lats.data,
+                            da.zeros_like(src_lons), # altitude
+                            new_axis=[2],
+                            chunks=(src_lons.chunks[0], src_lons.chunks[1], 3))
+        res = xr.DataArray(res, dims=['y', 'x', 'xyz'])
+        
+        # Retrieve new centroids 
+        # TODO: make it dask compatible using _upsample_centroid_dask [HELP WANTED]
+        # res1 = da.apply_along_axis(_upsample_centroid_dask,
+        #                            2, 
+        #                            res.data,
+        #                            x, 
+        #                            y)
+        # res1 = xr.DataArray(res1, dims=['y', 'x', 'coord'], coords=src_lons.coords)
+        
+        res = np.stack(upsample_centroids(res[:,:,0].data, 
+                                          res[:,:,1].data,
+                                          res[:,:,2].data, x_factor=x, y_factor=y), axis=2)
+        new_centroids = xr.DataArray(da.from_array(res), dims=['y', 'x', 'xyz']) 
+        
+        # Back-conversion to geographic CRS 
+        lonlatalt = da.map_blocks(self._do_transform, geocent, latlong,
+                                  new_centroids[:, :, 0].data, # x
+                                  new_centroids[:, :, 1].data, # y
+                                  new_centroids[:, :, 2].data, # z
+                                  new_axis=[2],
+                                  chunks=new_centroids.data.chunks)
+        
+        # Back-conversion array as input format
+        lons, _ = _convert_2D_array(lonlatalt[:, :, 0], to=src_lons_format, dims=['lats','lons'])
+        lats, _ = _convert_2D_array(lonlatalt[:, :, 1], to=src_lats_format, dims=['lats','lons'])
+        
+        # Add additional info if the source array is a DataArray
+        if isinstance(self.lats, xr.DataArray) and isinstance(self.lons, xr.DataArray):
+            lats.coords = res.coords 
+            lats.attrs = self.lats.attrs.copy()
+            lons.coords = res.coords 
+            lons.attrs = self.lons.attrs.copy()
+            try:
+                resolution = lons.attrs['resolution'] * ((x + y) / 2)  
+                lons.attrs['resolution'] = resolution
+                lats.attrs['resolution'] = resolution
+            except KeyError:
+                pass
+            
+        # Return the downsampled swath definition 
+        return SwathDefinition(lons, lats)
+    
+    def extend(self, x=0, y=0):
+        """Extend the swath definition along x (along-track) and y (across-track) dimensions.
+        By default, it does not extend on any direction. 
+        To extend of n pixel  on both sides of the across-track direction, call swath_def.extend(x=0, y=2).
+        """
+        import xarray as xr 
+        # Check input validity 
+        x = int(x)
+        y = int(y)
+        if x < 0 or y < 0: 
+            raise ValueError('x and y arguments must be positive integers.')
+            
+        # Return SwathDefinition if nothing to extend 
+        if x==0 and y==0: 
+            return self 
+        
+        # Get lats/lons numpy arrays
+        src_lats, src_lats_format = _convert_2D_array(self.lats, to='numpy', dims=['lats','lons'])
+        src_lons, src_lons_format = _convert_2D_array(self.lons, to='numpy', dims=['lats','lons'])
+        
+        dst_lats = src_lats
+        dst_lons = src_lons
+    
+        # Extend on y direction (side0 and side2)
+        if y > 0:
+            list_side0 = (src_lons[1,:], src_lats[1,:], src_lons[0,:], src_lats[0,:])
+            list_side2 = (src_lons[-2,:], src_lats[-2,:], src_lons[-1,:], src_lats[-1,:])
+            extended_side0_lonlats = _get_extended_lonlats(*list_side0, npts=y)
+            extended_side2_lonlats = _get_extended_lonlats(*list_side2, npts=y)
+            dst_lats = np.concatenate((extended_side0_lonlats[1][::-1,:], dst_lats), axis=0)
+            dst_lats = np.concatenate((dst_lats, extended_side2_lonlats[1]), axis=0)
+            dst_lons = np.concatenate((extended_side0_lonlats[0][::-1,:], dst_lons), axis=0)
+            dst_lons = np.concatenate((dst_lons, extended_side2_lonlats[0]), axis=0)
+        
+        # Extend on x direction (side1 and side3)    
+        if x > 0: 
+           list_side1 = (dst_lons[:,-2], dst_lats[:,-2], dst_lons[:,-1], dst_lats[:,-1])
+           list_side3 = (dst_lons[:, 1], dst_lats[:, 1], dst_lons[:, 0], dst_lats[:, 0])
+           extended_side1_lonlats = _get_extended_lonlats(*list_side1, npts=x, transpose=False)
+           extended_side3_lonlats = _get_extended_lonlats(*list_side3, npts=x, transpose=False)
+           dst_lats = np.concatenate((dst_lats, extended_side1_lonlats[1]), axis=1)
+           dst_lats = np.concatenate((extended_side3_lonlats[1][:,::-1], dst_lats), axis=1)
+           dst_lons = np.concatenate((dst_lons, extended_side1_lonlats[0]), axis=1)
+           dst_lons = np.concatenate((extended_side3_lonlats[0][:, ::-1], dst_lons), axis=1)
+      
+        # Back-conversion array as input format
+        lons, _ = _convert_2D_array(dst_lons, to=src_lons_format, dims=['lats','lons'])
+        lats, _ = _convert_2D_array(dst_lats, to=src_lats_format, dims=['lats','lons'])
+        
+        # Add additional info if the source array is a DataArray
+        if isinstance(self.lats, xr.DataArray) and isinstance(self.lons, xr.DataArray):
+            lats.attrs = self.lats.attrs.copy()
+            lons.attrs = self.lons.attrs.copy()
+            
+        # Return the extended SwathDefinition 
+        return SwathDefinition(lons, lats)
+    
+    def reduce(self, x=0, y=0):
+        """Reduce the swath definition along x (along-track) and y (across-track) dimensions.
+        By default, it does not reduce on any direction. 
+        To reduce of n pixel on both sides of the across-track direction, call swath_def.reduce(x=0, y=2).
+        """
+        # Check input validity (ensure reduced area is at least 2x2)
+        height = self.lats.shape[0]
+        width = self.lats.shape[1]
+        x = int(x)
+        y = int(y)
+        if x < 0 or y < 0: 
+            raise ValueError('x and y arguments must be positive integers.')
+        if x >= np.floor(width/2):
+            max_x = int(np.floor(width/2)) - 1 
+            raise ValueError("You can at maximum reduce the along-track direction (x) of SwathDef by {} pixels on each side.".format(max_x))
+        if y >= np.floor(height/2):
+            max_y = int(np.floor(height/2)) - 1
+            raise ValueError("You can at maximum reduce the across-track direction (y) of SwathDef by {} pixels on each side.".format(max_y))
+        
+        # Return SwathDefinition if nothing to reduce 
+        if x==0 and y==0: 
+            return self 
+        
+        # Return the reduced SwathDefinition
+        lats = self.lats[slice(0+y,height-y), slice(0+x,width-x)]
+        lons = self.lons[slice(0+y,height-y), slice(0+x,width-x)]
         return SwathDefinition(lons, lats)
 
     def __hash__(self):
@@ -1426,7 +1697,91 @@ class AreaDefinition(_ProjectionDefinition):
         width = int(self.width / dims.get('x', 1))
         height = int(self.height / dims.get('y', 1))
         return self.copy(height=height, width=width)
-
+    
+    def areadef_upsample(self, x=1, y=1):
+        """Return an upsampled version of the area."""
+        width = int(self.width * x)
+        height = int(self.height * y)
+        return self.copy(height=height, width=width)
+    
+    def areadef_extend(self, x=0, y=0): 
+        """Extend AreaDef by x/y pixels on each side."""
+        if self.is_geostationary: 
+            raise NotImplementedError("'extend' method is not implemented for GEO AreaDefinition.")
+        # Check input validity 
+        x = int(x)
+        y = int(y)
+        if x < 0 or y < 0: 
+            raise ValueError('x and y arguments must be positive integers.')
+        
+        # Retrieve pixel and area info
+        new_width = self.width + 2*x
+        new_height = self.height + 2*y
+        pixel_size_x = self.pixel_size_x
+        pixel_size_y = self.pixel_size_y
+        # Extend area_extent (lower_left_x, lower_left_y, upper_right_x, upper_right_y)
+        area_extent = self._area_extent
+        new_area_extent = list(area_extent)
+        new_area_extent[0] = new_area_extent[0] - pixel_size_x*x
+        new_area_extent[1] = new_area_extent[1] - pixel_size_y*y
+        new_area_extent[2] = new_area_extent[2] + pixel_size_x*x
+        new_area_extent[3] = new_area_extent[3] + pixel_size_y*y
+        # Define new AreaDefinition 
+        projection = self.crs_wkt  
+        area_def = AreaDefinition(self.area_id, self.description, self.proj_id,
+                                  projection=projection, 
+                                  width=new_width,
+                                  height=new_height,
+                                  area_extent=new_area_extent, 
+                                  rotation=self.rotation,
+                                  nprocs=self.nprocs, 
+                                  dtype=self.dtype)
+            
+        return area_def
+  
+    def areadef_reduce(self, x=0, y=0):
+        """Reduce AreaDef by x/y pixels on each side."""
+        if self.is_geostationary: 
+            raise NotImplementedError("'reduce' method is not implemented for GEO AreaDefinition.")
+        # Check input validity (ensure reduced area is at least 2x2)
+        width = self.width
+        height = self.height 
+        x = int(x)
+        y = int(y)
+        if x < 0 or y < 0: 
+            raise ValueError('x and y arguments must be positive integers.')
+        if x >= np.floor(width/2):
+            max_x = int(np.floor(width/2)) - 1 
+            raise ValueError("You can at maximum reduce width (x) of AreaDef by {} pixels on each side.".format(max_x))
+        if y >= np.floor(height/2):
+            max_y = int(np.floor(height/2)) - 1
+            raise ValueError("You can at maximum reduce height (y) of AreaDef by {} pixels on each side.".format(max_y))
+            
+        # Retrieve pixel and area info 
+        new_width = self.width + 2*x
+        new_height = self.height + 2*y
+        pixel_size_x = self.pixel_size_x
+        pixel_size_y = self.pixel_size_y
+        area_extent = self._area_extent
+        # Extend area_extent (lower_left_x, lower_left_y, upper_right_x, upper_right_y)
+        new_area_extent = list(area_extent)
+        new_area_extent[0] = new_area_extent[0] + pixel_size_x*x
+        new_area_extent[1] = new_area_extent[1] + pixel_size_y*y
+        new_area_extent[2] = new_area_extent[2] - pixel_size_x*x
+        new_area_extent[3] = new_area_extent[3] - pixel_size_y*y
+        # Define new AreaDefinition 
+        projection = self.crs_wkt  
+        area_def = AreaDefinition(self.area_id, self.description, self.proj_id,
+                                  projection=projection, 
+                                  width=new_width,
+                                  height=new_height,
+                                  area_extent=new_area_extent, 
+                                  rotation=self.rotation,
+                                  nprocs=self.nprocs, 
+                                  dtype=self.dtype)
+            
+        return area_def
+    
     @property
     def resolution(self):
         """Return area resolution in X and Y direction."""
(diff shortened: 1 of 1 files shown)


djhoese:
pre-commit.ci autofix

ghiggi:
This paragraph is to discuss the possibility of using `geotiepoints.GeoInterpolator` within `SwathDefinition.upsample`.
I might have misunderstood the usage of the GeoInterpolator, but I can't retrieve the correct upsampled centroids:

```
import numpy as np 
from geotiepoints.geointerpolator import GeoInterpolator

# Data Example 1 
x = 2
y = 2
src_lons = np.array([[5,9],
                     [5,9],
                     [5,9],
                     [5,9]])
src_lats = np.array([[6,6],
                     [4,4],
                     [0,0],
                     [-2,-2]])

# Data Example 2
lons = np.arange(-179.5, 179.5 + 0.5, 0.5)
lats = np.arange(-89.5,  0.5,  0.5)
src_lons, src_lats = np.meshgrid(lons, lats)
x = 2
y = 2

#--------------------------------
# Workflow 
height = src_lons.shape[0]
width = src_lons.shape[1]
   
tie_cols = np.arange(0, width*2, 2) 
fine_cols = np.arange(-1 + 1/x, width*2-1, 1/x*2) 

fine_rows = np.arange(-1 + 1/y, height*2-1, 1/y*2) 
tie_rows = np.arange(0, height*2, 2) 
   
# Print ties 
print(fine_rows)
print(tie_rows)

print(fine_cols)
print(tie_cols)

#--------------------------------
interpolator = GeoInterpolator((src_lons, src_lats),
                               (tie_rows, tie_cols),
                               (fine_rows, fine_cols),
                               2, 2)
interpolator.fill_borders("y", "x")  
lons_up, lats_up = interpolator.interpolate()
print("With fill borders:")
print(np.round(lats_up[:,0],2))  # stop at lats 0.5  instead of -1.5 (ex1) and -8.78 instead of 0.5 (ex2)
     
interpolator = GeoInterpolator((src_lons, src_lats),
                               (tie_rows, tie_cols),
                               (fine_rows, fine_cols),
                               1, 1) # 2,2, does not work with ex1
lons_up, lats_up = interpolator.interpolate()
print("")
print("Without fill borders:")
print(np.round(lats_up[:,0],2))  # With ex2, not using fill_borders enable to reach at least lat 0, but  do not estimate the borders
 

```

ghiggi:
The code below documents an uncommon SwathDefinition edge case where `swath_def.upsample(2,2).aggregate(2,2)` does not return the same source swath_def (only close to the poles).

```
import numpy as np
from pyresample.geometry import SwathDefinition
np.set_printoptions(suppress=True)

lons = np.arange(-179.75, 179.75 + 0.5, 0.5)
lats = np.arange(-89.75,  0.5,  0.5)
lons, lats = np.meshgrid(lons, lats)
swath_def = SwathDefinition(lons=lons, lats=lats)

### Precision error (?) between upsampling and aggregating 
swath_def_rev = swath_def.upsample(y=2, x=2).aggregate(x=2, y=2) 
np.allclose(swath_def_rev.lons, swath_def.lons, atol=1e-01)
np.allclose(swath_def_rev.lats, swath_def.lats, atol=1e-03)
np.isclose(swath_def_rev.lons, swath_def.lons, atol=1e-08)
bool_arr = np.isclose(swath_def_rev.lons, swath_def.lons, atol=1e-08)
swath_def.lats[~bool_arr.any(axis=1),0]  # problematic latitudes 
np.sum(bool_arr)/np.size(bool_arr)*100
```

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+
+        # Define geodetic and geocentric projection
         geocent = pyproj.Proj(proj='geocent')
         latlong = pyproj.Proj(proj='latlong')
+
+        # Get xr.DataArray with dask array
+        src_lats, src_lats_format = _convert_2D_array(self.lats, to='DataArray_Dask', dims=['y', 'x'])
+        src_lons, src_lons_format = _convert_2D_array(self.lons, to='DataArray_Dask', dims=['y', 'x'])
+
+        # Conversion to Geocentric Cartesian (x,y,z) CRS
```
mraspaud, line 763: I'd like these operations be performed with decorators to make the code clearer. So conversion to DataArray (and back) would we one decorator, and transformation to cartesian (and back) would be the other.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+            lats = lats.assign_coords(res.coords)
+            lats.attrs = self.lats.attrs.copy()
+            lons = lons.assign_coords(res.coords)
+            lons.attrs = self.lons.attrs.copy()
+            try:
+                resolution = lons.attrs['resolution'] * ((x + y) / 2)
+                lons.attrs['resolution'] = resolution
+                lats.attrs['resolution'] = resolution
+            except KeyError:
+                pass
```
mraspaud, line 798: This could be extracted as a private method for cleanliness.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+
+        # Return the downsampled swath definition
+        return SwathDefinition(lons, lats)
+
+    def upsample(self, x=1, y=1):
+        """Upsample the swath definition along x (along-track) and y (cross-track) dimensions.
```
mraspaud, line 804: Please don't use `along-track` and `accross-track`. Pyresample is use for many data type that are not satellite data. Prefer `lines` and `columns`.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+
+        To upsample of a factor of 2 (each pixel splitted in 4 pixels), call swath_def.upsample(x=2, y=2).
```
mraspaud, line 806:
```suggestion
        To upsample of a factor of 2 (each pixel splitted in 2x2 pixels), call swath_def.upsample(x=2, y=2).
```

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+        swath_def.upsample(x=1, y=1) simply returns the current swath_def.
+        """
+        # TODO: An alternative would be to use geotiepoints.geointerpolator.GeoInterpolator
```
mraspaud, line 809: Indeed, this should be using geotiepoints to avoid functionality duplication.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+                pass
+
+        # Return the downsampled swath definition
+        return SwathDefinition(lons, lats)
+
+    def extend(self, x=0, y=0):
+        """Extend the swath definition along x (along-track) and y (across-track) dimensions.
+
+        By default, it does not extend on any direction.
+        To extend of n pixel  on both sides of the across-track direction, call swath_def.extend(x=0, y=2).
```
mraspaud, line 912: what if we want to extend just on one side? Could we use another syntax to make this more flexible, eg `sd.extend(x=[-2, -1, width + 1.5])`?

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+        """
+        import xarray as xr
+
+        # Check input validity
+        x = int(x)
+        y = int(y)
+        if x < 0 or y < 0:
+            raise ValueError('x and y arguments must be positive integers.')
```
mraspaud, line 920: Please make the error message more informative, like `SwathDefinition.extend needs positive numbers of pixels.`

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+        # Extend on x direction (side1 and side3)
+        if x > 0:
+            list_side1 = (dst_lons[:, -2], dst_lats[:, -2], dst_lons[:, -1], dst_lats[:, -1])
+            list_side3 = (dst_lons[:, 1], dst_lats[:, 1], dst_lons[:, 0], dst_lats[:, 0])
+            extended_side1_lonlats = _get_extended_lonlats(*list_side1, npts=x, transpose=False)
+            extended_side3_lonlats = _get_extended_lonlats(*list_side3, npts=x, transpose=False)
+            dst_lats = np.concatenate((dst_lats, extended_side1_lonlats[1]), axis=1)
+            dst_lats = np.concatenate((extended_side3_lonlats[1][:, ::-1], dst_lats), axis=1)
+            dst_lons = np.concatenate((dst_lons, extended_side1_lonlats[0]), axis=1)
+            dst_lons = np.concatenate((extended_side3_lonlats[0][:, ::-1], dst_lons), axis=1)
```
mraspaud, line 953: I think there is some extrapolation capabilities in python_geotiepoints, could we use these to avoid duplication?

pyresample/geometry.py
```diff
@@ -892,6 +1137,239 @@ def compute_optimal_bb_area(self, proj_dict=None):
...
+    da = xr.DataArray(
+        data=arr,
+        dims=("y", "x"),
+        coords={"y": src_ties_y, "x": src_ties_x}
+    )
+    da_interp = da.interp({"y": dst_ties_y, "x": dst_ties_x}, method="linear")
+    return da_interp.data
+
+
+def _get_extended_lonlats(lon_start, lat_start, lon_end, lat_end, npts, transpose=True):
```
mraspaud, line 1347: Would it be more efficient/array-friendly to use vector rotation like in SLERP (https://en.wikipedia.org/wiki/Slerp). I think in particular slerp could be used for extrapolation. (We have an implementation of slerp in pygac here: https://github.com/pytroll/pygac/blob/main/pygac/slerp.py (which actually should be in geotiepoints if it was to be used outside pygac too))

pyresample/geometry.py
```diff
@@ -892,6 +1137,239 @@ def compute_optimal_bb_area(self, proj_dict=None):
...
+    elif isinstance(arr, xr.DataArray) and isinstance(arr.data, da.Array):
+        if to.lower() == 'numpy':
+            dst_arr = arr.data.compute()
+        elif to.lower() == 'dask':
+            dst_arr = arr.data
+        elif to.lower() == 'dataarray_numpy':
+            dst_arr = arr.compute()
+        else:  # to.lower() == 'dataarray_dask':
+            dst_arr = arr
+        return dst_arr, 'DataArray_Dask'
```
mraspaud, line 1224: Each of these blocks should be its own function.

pyresample/test/test_geometry.py
```diff
@@ -1872,26 +1872,357 @@ def test_compute_optimal_bb(self):
         assert_np_dict_allclose(res.proj_dict, proj_dict)
         self.assertEqual(res.shape, (6, 3))
 
-    def test_aggregation(self):
+    def test_downsampling(self):
```
mraspaud, line 1875: These test functions are way too long and have a lot of duplicated code. Please refactor these, and try to stick with the principle that one test function should assert one thing.

pyresample/geometry.py
```diff
@@ -892,6 +1137,239 @@ def compute_optimal_bb_area(self, proj_dict=None):
...
+
+def _get_extended_lonlats(lon_start, lat_start, lon_end, lat_end, npts, transpose=True):
+    """Utils employed by SwathDefinition.extend.
+
+    It extrapolate npts following the forward azimuth with an interdistance
+    equal to the distance between the starting point and the end point.
+    """
+    import pyproj
+    geod = pyproj.Geod(ellps='sphere')  # TODO: sphere or WGS84?
+    #  geod = pyproj.Geod(ellps='WGS84') # sphere
```
mraspaud, line 1355: Why not have this as a kwarg to let the user choose?

pyresample/geometry.py
```diff
@@ -1423,10 +1901,119 @@ def copy(self, **override_kwargs):
...
+            raise ValueError('x and y arguments must be positive integers larger or equal to 1.')
+        # Upsample
+        width = int(self.width * x)
+        height = int(self.height * y)
+        return self.copy(height=height, width=width)
+
+    def extend(self, x=0, y=0):
+        """Extend AreaDef by x/y pixels on each side."""
+        if self.is_geostationary:
+            raise NotImplementedError("'extend' method is not implemented for GEO AreaDefinition.")
```
mraspaud, line 1940: why can't the area extents of a geos area def be extended?

pyresample/geometry.py
```diff
@@ -1423,10 +1901,119 @@ def copy(self, **override_kwargs):
...
+                                  width=new_width,
+                                  height=new_height,
+                                  area_extent=new_area_extent,
+                                  rotation=self.rotation,
+                                  nprocs=self.nprocs,
+                                  dtype=self.dtype)
+
+        return area_def
+
+    def reduce(self, x=0, y=0):
```
mraspaud, line 1973: I think this is already implemented with the more pythonic `new_area = area[slice(-2, 2), slice(10, -4)]`

pyresample/test/test_geometry.py
```diff
@@ -2562,6 +2893,195 @@ def test_aggregate(self):
...
+        #                                    proj_id='seviri',
+        #                                    projection={'proj': 'geos',
+        #                                                'lon_0': 0.0,
+        #                                                'a': 6378169.00,
+        #                                                'b': 6356583.80,
+        #                                                'h': 35785831.00,
+        #                                                'units': 'm'},
+        #                                    width=123, height=123,
+        #                                    area_extent=[5500000, 5500000, -5500000, -5500000])
+        # self.assertRaises(NotImplementedError, area_geo.upsample, 2, 2)
```
mraspaud, line 2991: Please remove all comments that are not relevant.

mraspaud (changes requested):
Thank you very much for this PR. I can see that a lot of work went into this, I am grateful for that.

In general, I think there is a lot of duplication of code or functionality that could be avoided, and be careful about not breaking backwards compatibility. Pyresample is used by many, so we have to be very careful when deprecating functions for example.

Some more detailed comments inline.

Thanks again for the hard work!

mraspaud:
Also, this review took me more than 2 hours, in the future please consider making more incremental changes in each PR :sweat_smile:

pyresample/geometry.py
```diff
@@ -1423,10 +1901,119 @@ def copy(self, **override_kwargs):
...
+    def downsample(self, x=1, y=1):
+        """Return a downsampled version of the area."""
+        # Check input validity
+        x = int(x)
+        y = int(y)
+        if x == 1 and y == 1:
+            return self
+        if x < 1 or y < 1:
+            raise ValueError('x and y arguments must be positive integers larger or equal to 1.')
+        # Downsample
```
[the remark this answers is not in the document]
ghiggi: Not sure what do you mean by splitting a pixel of 1.5. 
The idea of this function is to split each pixel in `x` x `y` other pixels so you would require integers.

pyresample/geometry.py
```diff
@@ -1423,10 +1901,119 @@ def copy(self, **override_kwargs):
 
     def aggregate(self, **dims):
         """Return an aggregated version of the area."""
+        warnings.warn("'aggregate' is deprecated, use 'downsample' or 'upsample' instead.", PendingDeprecationWarning)
```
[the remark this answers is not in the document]
ghiggi: No. It was just to have method name consistency with SwathDefinition. 
If the input arguments are -1<x<1 the method currently fails later on when calling AreaDef methods ... 
`downsample `is the opposite of `upsample `.
The opposite of `aggregate` would be `disaggregate` ... thinking to image processing is more straightforward up/down sampling in my opinion. I remove the DeprecationWarning?

pyresample/geometry.py
```diff
@@ -892,6 +1137,239 @@ def compute_optimal_bb_area(self, proj_dict=None):
...
+            dst_arr = arr.compute()
+        else:  # to.lower() == 'dataarray_dask':
+            dst_arr = arr
+        return dst_arr, 'DataArray_Dask'
+
+    else:
+        raise NotImplementedError
+
+
+def _linspace1D_between_values(arr, num=0):
```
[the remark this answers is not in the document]
ghiggi: I will remove it. _linspace2D_between_values is used by `SwathDefinition.upsample` !

pyresample/geometry.py
ghiggi: Should I define these functions outside `geometry.py`? In `utils/array.py` maybe?

pyresample/geometry.py
```diff
@@ -892,6 +1137,239 @@ def compute_optimal_bb_area(self, proj_dict=None):
...
+
+    Useful to return swath lons, lats in the same original format after processing.
+
+    Parameters
+    ----------
+    arr : (np.ndarray, da.Array, xr.DataArray)
+        The 2D array to be converted to another array format.
+    to : TYPE
+        The desired array output format.
+        Accepted formats are: ['Numpy','Dask', 'DataArray_Numpy','DataArray_Dask']
```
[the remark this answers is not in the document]
ghiggi: I don't see how to specify "types" to infer if a xr.DataArray "contains" a numpy or dask array ...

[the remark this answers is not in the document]
mraspaud: not sure either, isn't there something like that when we use type annotations?

pyresample/geometry.py
mraspaud: Yes, that definitely doesn't belong here. Problably this whole function should be moved there then.

(Not in the scope of this PR) But I'm wondering, in the grand scheme of things, if it should be preferable to have subclasses for each type of container, like `_SwathDefinition_numpy`, `_SwathDefinition_dask`, etc

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: Ah, indeed, but since I though upsample should be rewritten to use geotiepoints, I didn't check it thoroughly, sorry.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+            raise ValueError("""You can at maximum reduce the across-track direction (y)
+                             of SwathDef by {} pixels on each side.""".format(max_y))
+
+        # Return SwathDefinition if nothing to reduce
+        if x == 0 and y == 0:
+            return self
+
+        # Return the reduced SwathDefinition
+        lats = self.lats[slice(0 + y, height - y), slice(0 + x, width - x)]
+        lons = self.lons[slice(0 + y, height - y), slice(0 + x, width - x)]
```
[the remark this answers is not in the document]
ghiggi: I added it because is the opposite of `extend`. It appeared to me would have been nice having boths.
If one want to extend a swath, then upsample and then reduce it a bit, to me the code 
`swath_def.extend(y=20, x=0).upsample(x=5,y=5).reduce(y=2, x=0)` looks more readable than 
`swath_def.extend(y=20, x=0).upsample(x=5,y=5)[slice(2, -2),:]` .

pyresample/geometry.py
ghiggi: I couldn't find the code. Could you point me to it?

mraspaud: https://github.com/pytroll/python-geotiepoints/blob/main/geotiepoints/interpolator.py#L110

pyresample/geometry.py
ghiggi: Mmh ok to enable to specify the extension on each side. 
Do we enable to provide x and y arguments as integer or list [x_top, x_bottom] ?
Or do we define x_top, y_right, x_bottom, y_left ? Or side0, side1, side2, side3?

pyresample/geometry.py
ghiggi: I wasn't able to use geotiepoints to do that. I would need help to do that. 
Look at the code snippet in the PR ... I get strange results. 
This implementation btw is dask-friendly compared to geotiepoints ... 
It can even made more efficient (see TODOs in the code)

pyresample/geometry.py
ghiggi: You mean what? All the If content? It changes between upsample and aggregate. With upsample I can not upsample eventual coords currently ...

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: I beg to differ :smile: , especially since `reduce` means something totally different in functional programming terms ( https://docs.python.org/3/library/functools.html#functools.reduce ) and that I don't find it intuitive at all that the function shrinks the data on both sides at once. I can only imagine the number of WTFs per minute we will get when reading code using this in the future :joy: . But maybe it's just me? ping @djhoese @pnuu

pyresample/geometry.py
ghiggi: I think I see what do you mean, but not sure how to do it properly. Here I just reused the preexisting template of `Swath.Definition.aggregate `

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+        This can be changed by passing boundary="trim" or boundary="pad", but behaviour within pyresample is undefined.
+        See https://xarray.pydata.org/en/stable/generated/xarray.DataArray.coarsen.html for further details.
         """
         import dask.array as da
         import pyproj
+        import xarray as xr
+
+        # Check input validity
+        x = int(x)
+        y = int(y)
```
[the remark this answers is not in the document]
ghiggi: I don't see what coarsening with float means. Would be an interpolation instead ...

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+    # TODO: this is more efficient than pyproj.transform.
+    # With pyproj > 3.1 it became thread safe:
+    # https://pyproj4.github.io/pyproj/stable/advanced_examples.html#multithreading
+    # @staticmethod
+    # def swath_do_transform(src, dst, lons, lats, alt):
+    #     """Run pyproj Transformer and stack the results."""
+    #     from pyproj import Transformer
+    #     transformer = Transformer.from_crs(src.crs, dst.crs)
+    #     x, y, z = transformer.transform(lons, lats, alt, radians=False)
+    #     return np.dstack((x, y, z))
```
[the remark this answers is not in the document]
ghiggi: Waiting your approval to do it. It might raise problems with pyresample versions with pyproj < 3.1?

pyresample/geometry.py
mraspaud: yes, I see that aggregate was the base for this, however now with the added array conversion it becomes clear that these conversions (array and cartesian) clutter the code and should be refactored to make the code more readable.
Using decorators, this function could be reduced to something like:
```python
@convert_arrays(...)
@convert_to_xyz(...)
def downsample(...):
    return res.coarsen(...)
```

pyresample/geometry.py
```diff
@@ -1423,10 +1901,119 @@ def copy(self, **override_kwargs):
 
     def aggregate(self, **dims):
         """Return an aggregated version of the area."""
+        warnings.warn("'aggregate' is deprecated, use 'downsample' or 'upsample' instead.", PendingDeprecationWarning)
         width = int(self.width / dims.get('x', 1))
         height = int(self.height / dims.get('y', 1))
         return self.copy(height=height, width=width)
 
+    def downsample(self, x=1, y=1):
```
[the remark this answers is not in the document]
ghiggi: Yes I duplicated the code for back-compatibility purposes. Current aggregate fails under some conditions. See above

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: can we have both, like this implementation and a fallback if pyproj 3.1 isn't available?

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: then we should improve `aggregate` :smile:

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: If aggregate is breaking stuff it should definitely be fixed. I'm not against having upsample/downsample, but if aggregate and downsample are synonyms, one should call the other.

pyresample/test/test_geometry.py
ghiggi: Can you elaborate a bit more? I am new to tests. 
Should I remove all testing of different array types? 
The other tests were necessary to not reduce coverage ...

pyresample/geometry.py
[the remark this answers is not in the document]
djhoese: I agree with Martin on this (shocking, I know). I think the heavy existing usage of `reduce` in functional programming makes this an unfortunate naming. Dask has a `reduce` function, but it resembles the functional approach where you provide it functions to call on portions of the array. I think someone using this method would assume something similar or at least pre-built options for how the array should be reduced/aggregated (first pixel in a slice, last pixel, mean of the pixels, etc). I'm also not sure why the `height - ` is needed, wouldn't a negative index work just as well?

pyresample/geometry.py
mraspaud: I will have a look and try to help asap. It could be that geotiepoints is not working optimally, so we should improve it to make it work here I think.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+        height = self.lats.shape[0]
+        width = self.lats.shape[1]
+        x = int(x)
+        y = int(y)
+        if x < 0 or y < 0:
+            raise ValueError('x and y arguments must be positive integers.')
+        if x >= np.floor(width / 2):
+            max_x = int(np.floor(width / 2)) - 1
+            raise ValueError("""You can at maximum reduce the along-track direction (x)
+                             of SwathDef by {} pixels on each side.""".format(max_x))
```
djhoese, line 983: Correct me if I'm wrong, but doesn't using triple quotes retain the whitespace you used for indentation? So the message would end up having a ton of whitespace in the middle when it is reported to the user.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+
+        # Check input validity
+        x = int(x)
+        y = int(y)
+        if x < 1 or y < 1:
+            raise ValueError('x and y arguments must be positive integers larger or equal to 1.')
 
+        # Return SwathDefinition if nothing to downsample
+        if x == 1 and y == 1:
+            return self
```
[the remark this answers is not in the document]
ghiggi: For sure improve performance because avoid unnecessary array and coords transformations !

ghiggi:
I plan to do the changes tomorrow afternoon @djhoese @mraspaud . Should I resolve the conservation comments after having resolved your concerns?

mraspaud:
Better leave them open if it's not trivial, so we can have another look after it's fixed and approve.

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: ok, so what does `coarsen` say when floats are passed? If it raises an error, maybe we don't need to the checking/casting ourselves?

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: True

pyresample/geometry.py
mraspaud: I mean everything after `# Add additional info if the source array is a DataArray`. I just mean moving that part to a separate method and call it from here.

pyresample/geometry.py
mraspaud: indeed, we could have
`def extend(self, top=[], bottom=[], left=[], right=[])`

pyresample/geometry.py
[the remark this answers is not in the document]
mraspaud: I'm thinking if I have an AreaDefinition that is 1500x1500 pixels and I want to downsample it to 1000x1000 pixels.

pyresample/test/test_geometry.py
mraspaud: you don't need to remove anything, everything you test here is valid of course. What I mean is that it's easier to read and debug if you keep small tests. So this test function here should be split in many smaller test functions, for example `test_downsampling_keeps_ndarrays` that would just do `assert isinstance(res_np.lats, np.ndarray)`, then another function called eg `test_downsampling_keeps_DataArray` with `assert isinstance(res_xr.lats, xr.DataArray)`, etc.
All the code that will be duplicated between the smaller test should be put in a common `setup` method, so you might need to create more test cases.
I'm advising that each test should just test one behaviour of what function you are testing.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
         x, y, z = transform(src, dst, lons, lats, alt)
         return np.dstack((x, y, z))
 
-    def aggregate(self, **dims):
```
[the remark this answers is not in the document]
ghiggi: It actually already assumed dimension names to be x and y at lines 
https://github.com/pytroll/pyresample/blob/main/pyresample/geometry.py#L735
`res = DataArray(res, dims=['y', 'x', 'coord'], coords=self.lons.coords)`
https://github.com/pytroll/pyresample/blob/main/pyresample/geometry.py#L746 
`resolution = lons.attrs['resolution'] * ((dims.get('x', 1) + dims.get('y', 1)) / 2)`

I however added back the aggregate method

pyresample/geometry.py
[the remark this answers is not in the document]
ghiggi: Done

pyresample/geometry.py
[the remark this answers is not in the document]
ghiggi: `coarsen` raises:

- `TypeError 'float' object cannot be interpreted as an integer` when passing a positive float 
- `ValueError: window must be > 0. Given -1 for dimension x` when passing a negative integer
- ` ValueError: window must be > 0. Given 0 for dimension x` when passing 0

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
         """
         import dask.array as da
         import pyproj
+        import xarray as xr
+
+        # Check input validity
+        x = int(x)
+        y = int(y)
+        if x < 1 or y < 1:
+            raise ValueError('x and y arguments must be positive integers larger or equal to 1.')
```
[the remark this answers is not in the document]
ghiggi: Done

pyresample/geometry.py
ghiggi: Mmh, this would imply that:
- **we force array and coordinate transformation also when downsample aggregation factors are 1** because AFAIK decorators would apply before the downsample function code. There is a way to avoid that? 
- we force `@convert_to_xyz(...)` to expects xr.DataArray? And raise an error if not DataArray? Or decorate  `convert_to_xyz` with `convert_arrays(to='DataArray_Dask')` decorator?

If you think this should be done @mraspaud, I can do it this afternoon. Let me know ;)

pyresample/geometry.py
ghiggi: Done

pyresample/geometry.py
ghiggi: Done

pyresample/geometry.py
ghiggi: I spent 3 hours yesterday again trying to use geotiepoints for this use case, but I failed. 
I think the code I added can be made dask-friendly very easily. Have a look at the TODOs  in the code.

pyresample/geometry.py
```diff
@@ -717,37 +717,282 @@ def _do_transform(src, dst, lons, lats, alt):
...
+                lons.attrs['resolution'] = resolution
+                lats.attrs['resolution'] = resolution
+            except KeyError:
+                pass
+
+        # Return the downsampled swath definition
+        return SwathDefinition(lons, lats)
+
+    def extend(self, x=0, y=0):
+        """Extend the swath definition along x (along-track) and y (across-track) dimensions.
```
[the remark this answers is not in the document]
ghiggi: Done

pyresample/geometry.py
ghiggi: Done

pyresample/geometry.py
ghiggi: Done

pyresample/geometry.py
ghiggi: I noticed really strange behaviours when using fill_borders. Look at the code snippet in the PR

pyresample/geometry.py
ghiggi: Solved

pyresample/geometry.py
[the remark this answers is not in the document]
ghiggi: I changed the name to `shrink`.
I used the `slice(..., height -y)` pattern instead of `slice(..., -y) `  to avoid obtaining `slice(..., -0)` when `x=0`.
I changed the input arguments to `left=0, right=0, bottom=0, top=0` to follow the new `extend` convention

pyresample/geometry.py
ghiggi: Moved to utils/array.py and each block separated in its own function.
I wonder what name convention is better: 
- `dataarray_dask` or `xarray_dask` 
- `dataarray_numpy` or `xarray_numpy`

@mraspaud @djhoese  any preference?

pyresample/geometry.py
ghiggi: Done

pyresample/geometry.py
ghiggi: We could leave that to my upcoming PR of spherical.py refactoring if you agree ...

pyresample/geometry.py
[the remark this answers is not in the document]
ghiggi: Fixed

pyresample/geometry.py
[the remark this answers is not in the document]
ghiggi: Done.

pyresample/geometry.py
```diff
@@ -1423,10 +1901,119 @@ def copy(self, **override_kwargs):
...
+        x = int(x)
+        y = int(y)
+        if x == 1 and y == 1:
+            return self
+        if x < 1 or y < 1:
+            raise ValueError('x and y arguments must be positive integers larger or equal to 1.')
+        # Upsample
+        width = int(self.width * x)
+        height = int(self.height * y)
+        return self.copy(height=height, width=width)
```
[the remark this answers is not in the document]
ghiggi: Done.

pyresample/geometry.py
ghiggi: Lons/lats arrays of geostationary area are a 2D matrix with a disk surrounded by nan or inf. 
By modifying width and height, we don't get the desired extended/shrunk area, but we just add/drop row/ column of NaN/Inf.

ghiggi: I also replaced `x` and `y` arguments with `left, right, bottom, top` arguments to be consistent with `SwathDefinition` methods

ghiggi: Also, `left, right, bottom, top` make not lot of sense with geos disc area ...

pyresample/test/test_geometry.py
ghiggi: I tried to follow your suggestions. Have a look at the `TestSwathDefinitionDownsampling` class in `test_geometry` and tell me if that is what you expect.
I wait for your feedbacks ... and then I will adapt also the rest of the tests. 
Is not clear to me if I need to create a test class for each method in the end ...

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
-        `sd.aggregate(x=2, y=2)`
+        """Return an aggregated version of the area."""
+        warnings.warn("'aggregate' is deprecated, use 'downsample' instead.", PendingDeprecationWarning)
+        return self.downsample(x=dims.get('x', 1), y=dims.get('y', 1))
+
+    def downsample(self, x=1, y=1, **kwargs):
+        """Downsample the SwathDefinition along x (columns) and y (lines) dimensions.
+
+        Builds upon xarray.DataArray.coarsen averaging function.
+        To downsample of a factor of 2, call swath_def.downsample(x=2, y=2)
```
djhoese, line 742:
```suggestion
        To downsample of a factor of 2, call ``swath_def.downsample(x=2, y=2)``.
```

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
+        swath_def.downsample(x=1, y=1) simply returns the current swath_def.
```
djhoese, line 743:
```suggestion
        ``swath_def.downsample(x=1, y=1)`` simply returns the current swath_def.
```

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
     def aggregate(self, **dims):
-        """Aggregate the current swath definition by averaging.
-
-        For example, averaging over 2x2 windows:
-        `sd.aggregate(x=2, y=2)`
+        """Return an aggregated version of the area."""
+        warnings.warn("'aggregate' is deprecated, use 'downsample' instead.", PendingDeprecationWarning)
+        return self.downsample(x=dims.get('x', 1), y=dims.get('y', 1))
+
+    def downsample(self, x=1, y=1, **kwargs):
```
djhoese, line 738: Could you add a google-style docstring and explain what `kwargs` is?

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
-        res = res.coarsen(**dims).mean()
+                            src_lons.data,
+                            src_lats.data,
+                            da.zeros_like(src_lons),  # altitude
+                            new_axis=[2],
+                            chunks=(src_lons.chunks[0], src_lons.chunks[1], 3))
+        res = xr.DataArray(res, dims=['y', 'x', 'xyz'], coords=src_lons.coords)
+
+        # Aggregating
+        res = res.coarsen(x=x, y=y, **kwargs).mean()
```
djhoese, line 780: This should probably wait for a separate PR, but it'd be nice if `mean` was a keyword argument that the user could control.

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
+                pass
+
+        # Return the downsampled swath definition
         return SwathDefinition(lons, lats)
 
+    def extend(self, left=0, right=0, bottom=0, top=0):
+        """Extend the SwathDefinition of n pixels on specific boundary sides.
+
+        By default, it does not extend on any side.
+        """
```
djhoese, line 922: You may have had this discussion with Martin in previous comment sections, but what is this method used for? What is the real world use case?

djhoese:
```suggestion
    def extend(self, left=0, right=0, bottom=0, top=0):
        """Extend the SwathDefinition by n pixels on specific boundary sides.

        By default, it does not extend on any side.
        """
```

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
+        y_max_shrink = height - 2
+        if (left + right) > x_max_shrink:
+            raise ValueError("SwathDefinition.shrink can drop maximum {} pixels "
+                             "along the x direction.".format(x_max_shrink))
+        if (top + bottom) > y_max_shrink:
+            raise ValueError("SwathDefinition.shrink can drop maximum {} pixels "
+                             "along the y direction.".format(y_max_shrink))
+
+        # Return the shrinked SwathDefinition
+        return self[slice(top, height - bottom), slice(left, width - right)]
```
djhoese, line 1011: I think I mentioned this in a past comment, but why can't negative indexes be used here?

I don't see much utility here over just "normal" slicing. If `extend` has a real world use case I'd almost like to consider calling it something like `extrapolate` and getting rid of `shrink`.

pyresample/utils/array.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Useful to return swath lons, lats in the same original format after processing.
+
+    Parameters
+    ----------
+    arr : (np.ndarray, da.Array, xr.DataArray)
+        The 2D array to be converted to another array format.
+    to : TYPE
+        The desired array output format.
+        Accepted formats are: ['Numpy','Dask', 'DataArray_Numpy','DataArray_Dask']
```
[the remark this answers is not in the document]
djhoese: To add to this and clarify a little, you could use actually class objects for this (`np.ndarray`, `da.Array`, `xr.DataArray`, etc). Or even better (at least that's my current feeling) do the split functions like I mentioned and now that I realize you return the input type, you could instead return the function needed to reverse the conversion.

Or...you could do something like a decorator on the methods that replaces your use of these functions so that the internal implementation of the methods only deals with the object types they want. All the conversion complexity is kept outside of the internals/implementation. I do realize that there are some edge cases where a conversion would be done when nothing was really modified in the data, but if done correctly that shouldn't be too big of a deal.

These are just ideas. I can definitely be convinced otherwise, but the new string names for these things feels like the wrong approach.

djhoese (changes requested):
I reviewed what I could of the main code. I skipped over the tests as I see that they are pretty long and I had larger scale questions/comments on the main code. I think you have some really good ideas in this PR. I'm not sure I understand everything you're going for, but that's where I asked questions. I'm a little worried about performance in the different array cases and error handling/logging when a user doesn't have xarray or dask installed but the method requires it.

Part of me thinks the upsample/downsample stuff should be put in a separate module as a generic geocentric (or geographic?) upsample/downsample pair of functions. The methods in SwathDefinition could then be simple calls to these functions and keep the amount of code in `geometry.py` down a bit (it is already way too long before this PR).

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
+        """Perform pyproj transformation and stack the results.
+
+        If using pyproj >= 3.1, it employs thread-safe pyproj.transformer.Transformer.
+        If using pyproj < 3.1, it employs pyproj.transform.
+
+        Docs: https://pyproj4.github.io/pyproj/stable/advanced_examples.html#multithreading
+        """
+        if float(pyproj.__version__[0:3]) >= 3.1:
+            from pyproj import Transformer
+            transformer = Transformer.from_crs(src.crs, dst.crs)
```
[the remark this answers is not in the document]
mraspaud: yes, for our use cases, `always_xy=True` is a must

pyresample/utils/array.py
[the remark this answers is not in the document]
ghiggi: OK I will take some time to think of your suggestions and try implementing it. 
Just to be sure @djhoese: 
- You mean to pass in the function argument the class object (`np.ndarray, da.Array, xr.DataArray`) we want instead of a string-acronym right? 
- Should I add a `chunked/lazy=True/False` argument to decide whether to return an xr.DataArray with dask or numpy array? Or default to a xr.DataArray with numpy?

pyresample/geometry.py
ghiggi: If `bottom=0` or `right=0`, `slice(top, -0)` or `slice(left, -0)` lead to empty arrays.

pyresample/geometry.py
ghiggi: I am currently implementing a pipeline for collocating and remapping GEO imagery on LEO sensor swath.
It's very useful to actually have GEO imagery outside the original LEO swath for downstream image processing/ML applications.

As an example, having a LEO sensor with 5 km footprint resolution, I first extend the LEO SwathDefinition of N pixels on both sides along-track, then I upsample the SwathDefinition by a factor of 5 to match the 1km resolution of GEO imagery.  --> `swathdef.extend(top=N, bottom=N).upsample(x=5,y=5) `

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
+                            src_lats.data,
+                            da.zeros_like(src_lons),  # altitude
+                            new_axis=[2],
+                            chunks=(src_lons.chunks[0], src_lons.chunks[1], 3))
+        res = xr.DataArray(res, dims=['y', 'x', 'xyz'])
+
+        # Retrieve new centroids
+        # TODO: make it dask compatible using _upsample_centroid_dask [HELP WANTED]
+        # res1 = da.apply_along_axis(_upsample_centroid_dask,
+        #                            2,
```
ghiggi, line 882: @djhoese have a look at this portion of code when you will have half an hour.
I am not sure what is the best way to use dask here.

pyresample/geometry.py
djhoese: Well dang. Ok...`None` might work but eh whatever.

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
-        """Run pyproj.transform and stack the results."""
-        x, y, z = transform(src, dst, lons, lats, alt)
+        """Perform pyproj transformation and stack the results.
+
+        If using pyproj >= 3.1, it employs thread-safe pyproj.transformer.Transformer.
+        If using pyproj < 3.1, it employs pyproj.transform.
+
+        Docs: https://pyproj4.github.io/pyproj/stable/advanced_examples.html#multithreading
+        """
+        if float(pyproj.__version__[0:3]) >= 3.1:
```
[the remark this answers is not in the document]
mraspaud: What if pyproj gets to version 3.10? in floats, that's 3.1...
@djhoese I could live with that too.

pyresample/geometry.py
```diff
@@ -713,43 +715,301 @@ def copy(self):
...
+            except KeyError:
+                pass
+
+        # Return the downsampled swath definition
+        return SwathDefinition(lons, lats)
+
+    def upsample(self, x=1, y=1):
+        """Upsample the SwathDefinition along x (columns) and y (lines) dimensions.
+
+        To upsample of a factor of 2 (each pixel splitted in 2x2 pixels),
```
djhoese, line 813:
```suggestion
        To upsample of a factor of 2 (each pixel split into 2x2 pixels),
```

pyresample/geometry.py
djhoese: I do not fully understand the logic here, but I understand the overall goal. To summarize we have an overall array of (M rows, N columns) and we want to produce a final array with a shape of (M * y, N * x), right? To do this we need convert from lon/lat to x/y/z geocentric, do some math, then convert back x/y/z to lon/lat. In the "do some math" portion, do we need to the values of all the pixels next to the current pixel? Or can we operate on groups of `y` by `x` pixels?

Given what I've learned about number of tasks in a dask graph, I think it would be best to do this entire thing in a giant map_blocks call. By "entire thing" I mean lon/lat -> x/y/z, the math, x/y/z -> lon/lat. If you require to work on groups of `y` by `x` pixels then you could make sure that the input chunks are divisible by those numbers (2x2 -> 1024x1024 would be ok). We don't want to make really small chunks. If you need overlap between the computations (in the case where each pixel needs to know about all the values around it) then `map_overlap` may be something to look into: https://docs.dask.org/en/stable/array-overlap.html

pyresample/geometry.py
ghiggi: I stupidly mentioned that I am extending along-track while I am extending the SwathDefinition cross-track ;)

I essentially want to have neighbor information around the swath and upscale it to: 
1. not lose potential fine-scale information available from GEO data (i.e. overshooting tops, inhomogeneity/non-uniform beam filling degree,...). 
2. and if I apply a series of convolutions (i.e. with CNN), the image at the borders gets shrunk when not applying padding, so this is another reason to extend the swath cross-track so that after a series of convolutions/pooling operations I can be aligned with original LEO SwathDefinition.

djhoese: Does the ML code handle the geolocation of the data at all? Or are you telling it (in some way) that the top-left 2x2 pixels of the resampled GEO pixels represent pixel [0, 0] of the LEO data? Or I guess in this extended case the extrapolated pixels of the GEO data don't have any corresponding LEO pixel.

If the ML has a concept of the location of the data, then why not simply slice the GEO data to the bounds of the LEO data instead of resampling?

Note: I'm not trying to argue against this `extend` method I just really want to understand how it is used so I can argue for/against changes in the future.

pyresample/geometry.py
ghiggi: I can do also this.
There is a small set of possible functions that can be called from an [xarray.core.rolling.DataArrayCoarsen](https://xarray.pydata.org/en/stable/generated/xarray.core.rolling.DataArrayCoarsen.html#xarray.core.rolling.DataArrayCoarsen) object. I can add the default argument` fun="mean"`  and then do:
```
res = res.coarsen(x=x, y=y, **kwargs)
func = getattr(res, fun')
res = func(res)
```
Should I check for valid `fun` values in `set("min", "max", "sum", "std", "var", "count", "mean", "median") `? Or defer this to the xarray function?

djhoese: I'd say add an explicit keyword argument in the method signature for `coarsen_func` or something like that and document in the docstring to look at the xarray docs (link to the exact page if possible, intersphinx may allow you to do something like:

```
See the :doc:`Xarray coarsen documentation <xarray:coarsen-docs-page>` for more information.
```

And I think you could leave the error handling up to the `getattr`. You could shorten this to:

```
coarsen_res = res.coarsen(...)
res = getattr(coarsen_res, coarsen_func)()
```

This assumes the coarsen function doesn't take additional kwargs.

PR #414, opened 2022-01-26
Fix gradient search for single band data

pnuu:
Helping out a colleague, I noticed that gradient search resampler doesn't work if a single-channel dataset (shape `(1, y, x)`) are used.

The error message was:
```python
  File "/home/lahtinep/bin/test_akun_resamplays.py", line 40, in <module>
    main()
  File "/home/lahtinep/bin/test_akun_resamplays.py", line 31, in main
    scn2 = scn.resample(suomi, resampler='gradient_search', fill_value=0)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/scene.py", line 906, in resample
    self._resampled_scene(new_scn, destination, resampler=resampler,
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/scene.py", line 822, in _resampled_scene
    res = resample_dataset(dataset, destination_area, **kwargs)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/resample.py", line 1395, in resample_dataset
    new_data = resample(source_area, dataset, destination_area, fill_value=fill_value, **kwargs)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/satpy/satpy/resample.py", line 1358, in resample
    res = resampler_instance.resample(data, **kwargs)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/pyresample/pyresample/resampler.py", line 126, in resample
    return self.compute(data, cache_id=cache_id, **kwargs)
  File "/home/lahtinep/Software/pytroll/pytroll_packages/pyresample/pyresample/gradient/__init__.py", line 274, in compute
    res = xr.DataArray(res, dims=data_dims, coords=coords)
  File "/home/lahtinep/mambaforge/envs/pytroll/lib/python3.9/site-packages/xarray/core/dataarray.py", line 406, in __init__
    coords, dims = _infer_coords_and_dims(data.shape, coords, dims)
  File "/home/lahtinep/mambaforge/envs/pytroll/lib/python3.9/site-packages/xarray/core/dataarray.py", line 102, in _infer_coords_and_dims
    raise ValueError(
ValueError: coords is not dict-like, but it has 3 items, which does not match the 2 dimensions of the data
```

This PR fixes this by only applying `.squeeze()` in the case of result being of higher dimensionality than the `dims`/`coords` would indicate.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/gradient/__init__.py
+++ b/pyresample/gradient/__init__.py
@@ -271,6 +271,9 @@ class GradientSearchResampler(BaseResampler):
 
         if fill_value is not None:
             res = da.where(np.isnan(res), fill_value, res)
+        if res.ndim > len(data_dims):
+            res = res.squeeze()
+
         res = xr.DataArray(res, dims=data_dims, coords=coords)
 
         return res
@@ -400,6 +403,6 @@ def _concatenate_chunks(chunks):
         prev_y = y
     res.append(da.concatenate(col, axis=1))
 
-    res = da.concatenate(res, axis=2).squeeze()
+    res = da.concatenate(res, axis=2)
 
     return res
--- a/pyresample/test/test_gradient.py
+++ b/pyresample/test/test_gradient.py
@@ -228,6 +228,16 @@ class TestGradientResampler(unittest.TestCase):
         assert np.allclose(res[1, :, :], 2.0)
         assert np.allclose(res[2, :, :], 3.0)
 
+    def test_resample_area_to_area_3d_single_channel(self):
+        """Resample area to area, 3d with only a single band."""
+        data = xr.DataArray(da.ones((1, ) + self.src_area.shape,
+                                    dtype=np.float64),
+                            dims=['bands', 'y', 'x'])
+        res = self.resampler.compute(
+            data, method='bil').compute(scheduler='single-threaded')
+        assert res.shape == (1, ) + self.dst_area.shape
+        assert np.allclose(res[0, :, :], 1.0)
+
     def test_resample_swath_to_area_2d(self):
         """Resample swath to area, 2d."""
         data = xr.DataArray(da.ones(self.src_swath.shape, dtype=np.float64),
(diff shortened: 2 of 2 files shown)


djhoese:
Does this error suggest that gradient search can't handle 3D+ inputs at all?

pnuu:
I think it will handle N-D just fine, but in 2-D case with additioal `'bands'` dimension the output array was `.squeeze()`'d and `dims`/`coords` didn't anymore match the array shape after resampling.

djhoese:
Sorry I'm a little slow today. Doesn't this mean that the input data should have the `bands` dimension removed before being resampled so then the `res` output array is only 2D?

pnuu:
That's an alternative, yes. But if the input has a shape `(num_bands, y, x)` I'd expect the output also to be `(num_bands, y2, x2)` instead of `(y2, x2)` for some cases (`num_bands = 1`).

djhoese:
I guess what I'm trying to say is why isn't the low level function getting a 2D array and then the higher level code is reconstructing a 3D array if needed?

pnuu:
So first squeezing and then then re-expanding if necessary? Nope. This "squeeze if necessary" is a one step way for the same result. If the input data already have `.ndims = 2` (no `'bands'`) then also the output will be that way.

djhoese:
What I'm saying is if this resampling already handles multi-band images then how/why is your needed? If this method is only supposed to handle 2D arrays and the caller is supposed to handle the split/expand of a multi-band image (ex. RGB) then this change makes more sense but I'm still wondering why it is getting a 3D array as input if the caller is supposed to give it a 2D array.

djhoese:
So what happens when you give an RGB to this resampler?

pnuu:
It works exactly like it did before.

djhoese:
I'm not suggesting your change breaks this. I'm suggesting that this fix isn't fixing the whole problem. If an RGB is given to this resampler, does it fail? If so, then the longer term fix (alternative to this current fix) would be to split the arrays, resample, and rejoin. If it doesn't fail, then there is something weird going on with dimensions here that whatever logic is supposed to be handling non-geo dimensions is not handling them correctly.

pnuu:
No, RGB doesn't fail, as it has both `dims` and `coords` for all the dimensions, and the `.ndims` of the array doesn't change, as `.squeeze()` isn't removing the 3-band dimension like it would do for an array based on `L` mode `(1, y, x)` data. In the `L` case there would be an extra `'bands'` value in the `.coords` and `.dims` that doesn't exist in the output array because it was erroneously `.squeeze()`'d away. The way it is handled in this PR is consistent in retaining the shape, dimensions and coordinates of the input data.

djhoese (approved):

djhoese:
Feel free to merge or wait for @mraspaud review.

mraspaud (approved):
LGTM, but it would be nice to know why :)

pnuu:
Xarray, correctly, doesn't add 3D coordinate/dimension labels to a 2D dataset. The input being originally a 3D representation of 2D data with dimensions `(bands, y, x)` where `bands = ['L']`, and the original code squeezing the single-value `'bands'` dimension away from the array, the dimensionalities of the data and `dims` didn't match anymore.

PR #415, opened 2022-01-31
Update AreaDefinition equality to use pyproj CRS

djhoese:
Pyproj has issued a warning for a long time now to tell the user to avoid using PROJ.4 strings and dictionaries as they are considered inaccurate as far as completely defining a projection (where WKT is now preferred). This PR updates one usage of this in `AreaDefinition.__eq__` so it compares the pyproj CRS objects directly instead of the PROJ.4 strings. This change does make equality a lot more strict, but probably in a good way. It now fails if your ellipsoid was defined in different ways or if axis order is different. This is very common with EPSG inputs.

This PR updates a few things to improve on this in the tests. It also includes a backwards incompatible change to the rasterio utilities to bring the argument name inline with AreaDefinition and be more accurate (`projection` instead of `proj_dict`).

Note: With these changes I am able to generate an ABI true_color composite in Satpy with no PROJ warnings.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1817,7 +1817,7 @@ class AreaDefinition(_ProjectionDefinition):
     def __eq__(self, other):
         """Test for equality."""
         try:
-            return ((self.proj_str == other.proj_str) and
+            return ((self.crs == other.crs) and
                     (self.shape == other.shape) and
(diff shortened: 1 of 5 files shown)


pnuu (approved):
LGTM

PR #418, opened 2022-02-04
Implement get_abs_max on BucketResampler

### LINKED ISSUE #417: Add get_abs_max (and get_abs_min) to BucketResampler
## Feature request

**Is your feature request related to a problem? Please describe.**

For the parallax correction https://github.com/pytroll/satpy/pull/1904, I could use a `.get_abs_max()` method on the `BucketResampler` class.

In the first step, the parallax correction algorithm calculates the parallax error for each pixel.  This results in a Δlat and Δlon.  It then calculates, for each pixel, where that pixel is supposed to be.  This leads to multiple lat/lon being assigned to a particular pixel: if a cloud at the Northern Hemisphere is shifted toward the equator (for a geostationary satellite), it may end up in a pixel that was previously clear-sky or had a lower cloud.  The same pixel may then contain a Δlat of -0.01 (pixel originated 0.01° to the north) and -0.05 (pixel originated 0.05° to the north).  The correct answer for a parallax correction should be to select the value -0.05°, because the larger shift corresponds to the higher cloud.  Likewise, in the Southern Hemisphere, 0.05° should win over 0.01°.  Therefore, I need a way to select a pixel with the largest absolute shift.

The real situation could be even more complex, because satellites are not always located at (0°, 0°).  For IODC, a pixel at 20° longitude may shift east and one at 60° longitude may shift west.  But I think I can handle such a shift client-side.  The absolute maximum is harder to implement client-side, because it is critically important that the sign is retained.  Retaining the sign through resampling when passing the absolute value of the Δ is tricky and may be inefficient or erroneous.

**Describe the solution you'd like**

I would like that the `BucketResampler` class grows a method `.get_abs_max()`.

For consistency, it should probably also have `.get_abs_min()`, even though I don't need it myself.

**Describe any changes to existing user workflow**

N/A.

**Additional context**

I have tried to simulate this by resampling both the absolute value and its sign, but resampling the sign has its own challenges.

gerritholl:
Implement get_abs_max on the BucketResampler class.

 - [x] Closes #417 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -226,6 +226,18 @@ class BucketResampler(object):
                                                        bins=np.linspace(0, out_size, out_size)
                                                        )
                                            )['values'].max())
+        elif scipy_method == "absmin":
+            def get_abs_min(part):
+                gb = part.groupby(np.digitize(part.x, bins=np.linspace(0, out_size, out_size)))
+                return gb.apply(lambda p: p.loc[abs(p["values"]).argmin()])
+            statistics = df.map_partitions(get_abs_min)
+        elif scipy_method == "absmax":
+            def get_abs_max(part):
+                gb = part.groupby(np.digitize(part.x, bins=np.linspace(0, out_size, out_size)))
+                return gb.apply(lambda p: p.loc[abs(p["values"]).argmax()])
+            statistics = df.map_partitions(get_abs_max)
+        else:
+            raise ValueError(f"Invalid method: {scipy_method:s}")
 
         # fill missed index
         statistics = (statistics + pd.Series(np.zeros(out_size))).fillna(0)
@@ -295,6 +307,64 @@ class BucketResampler(object):
         """
         LOG.info("Get max of values in each location")
         return self._call_pandas_groupby_statistics('max', data, fill_value, skipna)
+
+    def get_abs_max(self, data, fill_value=np.nan, skipna=True):
+        """Calculate absolute maximums for each bin with drop-in-a-bucket resampling.
+
+        .. warning::
+
+            The slow :meth:`pandas.DataFrame.groupby` method is temporarily used here,
+            as the `dask_groupby <https://github.com/dcherian/dask_groupby>`_ is still under development.
+
+        Parameters
+        ----------
+        data : Numpy or Dask array
+            Data to be binned.
+        skipna : boolean (optional)
+                If True, skips NaN values for the maximum calculation
+                (similarly to Numpy's `nanmax`). Buckets containing only NaN are set to zero.
+                If False, sets the bucket to NaN if one or more NaN values are present in the bucket
+                (similarly to Numpy's `max`).
+                In both cases, empty buckets are set to 0.
+                Default: True
+
+        Returns
+        -------
+        data : Numpy or Dask array
+            Bin-wise maximums in the target grid
+        """
+        LOG.debug("Get abs max of values in each location")
+        return self._call_pandas_groupby_statistics('absmax', data, fill_value, skipna)
+
+
+    def get_abs_min(self, data, fill_value=np.nan, skipna=True):
+        """Calculate absolute minimums for each bin with drop-in-a-bucket resampling.
+
+        .. warning::
+
+            The slow :meth:`pandas.DataFrame.groupby` method is temporarily used here,
+            as the `dask_groupby <https://github.com/dcherian/dask_groupby>`_ is still under development.
+
+        Parameters
+        ----------
+        data : Numpy or Dask array
+            Data to be binned.
+        skipna : boolean (optional)
+                If True, skips NaN values for the maximum calculation
+                (similarly to Numpy's `nanmax`). Buckets containing only NaN are set to zero.
+                If False, sets the bucket to NaN if one or more NaN values are present in the bucket
+                (similarly to Numpy's `max`).
+                In both cases, empty buckets are set to 0.
+                Default: True
+
+        Returns
+        -------
+        data : Numpy or Dask array
+            Bin-wise maximums in the target grid
+        """
+        LOG.debug("Get abs min of values in each location")
+        return self._call_pandas_groupby_statistics('absmin', data, fill_value, skipna)
+
 
     def get_count(self):
         """Count the number of occurrences for each bin using drop-in-a-bucket resampling.
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -329,6 +329,11 @@ class BucketResampler(object):
         """
         max_ = self.get_max(data, fill_value=fill_value, skipna=skipna)
         min_ = self.get_min(data, fill_value=fill_value, skipna=skipna)
+        return self._get_abs_max_from_min_max(min_, max_)
+
+    @staticmethod
+    def _get_abs_max_from_min_max(min_, max_):
+        """From array of min and array of max, get array of abs max."""
         return da.where(-min_ > max_, min_, max_)
 
     def get_count(self):
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -226,8 +226,6 @@ class BucketResampler(object):
                                                        bins=np.linspace(0, out_size, out_size)
                                                        )
                                            )['values'].max())
-        else:
-            raise ValueError(f"Invalid method: {scipy_method:s}")
 
         # fill missed index
         statistics = (statistics + pd.Series(np.zeros(out_size))).fillna(0)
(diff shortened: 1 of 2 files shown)


gerritholl:
I don't understand why this fails.  It works in practice.

gerritholl:
It also seems to be very slow ☹

Computing `br.get_abs_max()` for a full disc image takes 20.7 seconds on my machine.  Even just constructing the dask graph (calling `br.get_abs_max()` without computing) takes 24.9 seconds, such that in total it takes around 45 seconds to get the absolute maximum for a full disc image… not great.

djhoese:
I think this was @zxdawn's problem in the other PR. It is hard to get this stuff to perform well.

zxdawn:
@djhoese Yes, #368. That's the slow process of calculating the min/max of group results. Maybe @dcherian has some ideas? He's developing [flox](https://github.com/dcherian/flox) which supports dask for groupby().

gerritholl:
I'm using it in the parallax correction (https://github.com/pytroll/satpy/pull/1904), which, as it stands now using `.get_abs_max()`, is too slow to be used operationally (but I still need to check if that has other causes the `.get_abs_max()` being slow).

zxdawn:
@gerritholl  Here's the [question](https://stackoverflow.com/questions/67882615/fastest-way-to-get-the-minimum-value-of-data-array-in-another-paired-bin-array) I asked on StackOverflow. The root should be the slow process of pandas groupby(). You can test the time it costs.

gerritholl:
Thanks @dcherian for the reply!  I don't have a minimal example yet.  In my non-minimal real use case, I have 3712×3712 ≅ 13.7M bins.  99% of the bins have exactly one value.  About 0.5% have 2 values and less than 0.1% have 3 or more values, with the largest number of values in the image being 5.  Slightly less than 0.5% have 0 values.  The chunks are much larger than the bins.  I have no time dimension.  My aim is to get the largest absolute value in each bin, or nan for bins without any values.

I suspect that my current approach is widely suboptimal, regardless of using dask or not.  For the 99% I should just keep the single value, and only the ones with 2 or more should need any calculations.

I should certainly look at flox.  Unfortunately, I'm neither very well versed in dask nor in pandas, so I'm not sure yet how to approach optimising my implementation (and at the same time keep it flexible and generic enough for it to fit in pyresample).

dcherian:
Thanks @gerritholl what are your chunk sizes then?

>  My aim is to get the largest absolute value in each bin, or nan for bins without any values.

It sounds like this is mostly reindexing and a tiny amount of groupby-reduction. Are the lat, lon for the input image stored as numpy arrays i.e. do you know ahead of time which x,y pixels belong to which bin? 

I'm guessing these are dask arrays (based on the documentation for bucketing). In that case, any general solution (like the one in flox) will effectively rechunk to one chunk along x,y for the output grid. Can you start with that chunking scheme instead?

> I should certainly look at flox. 

It's not worth it right now, grouping by multiple dask variables has not been implemented yet.

It would help if you provided some example code, I can take a look next week but no promises. I'd need the array to group, arrays that we are grouping by(lat,lon i'm guessing) and the output grid. It's OK if the stuff is being read from a file.

gerritholl:
Each of the arrays for data, lat, and lon have shape `(3712, 3712)` and chunksize `(1024, 1024)`, all with dtype `float64`.

I really appreciate your offer for assistance!  I will work on some example code.

gerritholl:
Likely related to the slowness is the observation that the unit test passes only with `CustomScheduler(max_computes=3)`, not with 2 or 1.

gerritholl:
> Likely related to the slowness is the observation that the unit test passes only with `CustomScheduler(max_computes=3)`, not with 2 or 1.

I suppose that ideally, they should pass with `CustomScheduler(max_computes=0)`, but that's not the case for `.get_max()` or `.get_min()` either.

Seeing that the slowness is probably a consequence of `.get_max()` and `.get_min()` being slow (they share almost all code between them), which may be to the premature calculations, is this something I should fix here or may I defer this to another PR (such as #368)?

djhoese:
@zxdawn do you have time to give a formal review of this PR?

zxdawn:
@djhoese sorry, I'm on the vacation in the Iceland and will back on 10th April. If I find any free time, I will check it.

djhoese:
> I'm on the vacation

Get off of GitHub!

zxdawn:
@gerritholl Just one small comment: When I see *absolute max*, the *maximum value of absolute values* comes to my mind. Actually, your function gets the original value which has the largest absolute value. Maybe it's better to clarify it in the function?

gerritholl:
> @gerritholl Just one small comment: When I see _absolute max_, the _maximum value of absolute values_ comes to my mind. Actually, your function gets the original value which has the largest absolute value. Maybe it's better to clarify it in the function?

Right!  Clarified in the method docstring.

gerritholl:
The latest tests pass if and only if the latest version of https://github.com/pytroll/pyresample/pull/368 is merged first.

pyresample/test/test_bucket.py
```diff
@@ -198,15 +199,15 @@ def test_get_min(self):
...
         self.assertEqual(np.count_nonzero(result == 2), 1)
-        # test single entry average
+        # test single entry minimum
         self.assertEqual(np.count_nonzero(result == 5), 1)
         # test that minimum of bucket with only nan is nan, and empty buckets are nan
         self.assertEqual(np.count_nonzero(~np.isnan(result)), 2)
 
     def _get_max_result(self, data, **kwargs):
-        """Compute the bucket average with kwargs and check that no dask computation is performed."""
+        """Compute the bucket max with kwargs and check that no dask computation is performed."""
```
sfinkens, line 210: Thanks for fixing these typos!

pyresample/bucket/__init__.py
```diff
@@ -296,6 +296,38 @@ def get_max(self, data, fill_value=np.nan, skipna=True):
...
+        Parameters
+        ----------
+        data : Numpy or Dask array
+            Data to be binned.
+        skipna : boolean (optional)
+                If True, skips NaN values for the maximum calculation
+                (similarly to Numpy's `nanmax`). Buckets containing only NaN are set to zero.
+                If False, sets the bucket to NaN if one or more NaN values are present in the bucket
+                (similarly to Numpy's `max`).
+                In both cases, empty buckets are set to 0.
```
sfinkens, line 319: Looking at `test_get_abs_max` it seems that empty buckets and all-nan-buckets are set to `fill_value`. Or am I misunderstanding something?

pyresample/test/test_bucket.py
```diff
@@ -216,13 +217,31 @@ def test_get_max(self):
...
-        # test multiple entries average
+        # test multiple entries maximum
         self.assertEqual(np.count_nonzero(result == 11), 1)
-        # test single entry average
+        # test single entry maximum
         self.assertEqual(np.count_nonzero(result == 5), 1)
         # test that minimum of bucket with only nan is nan, and empty buckets are nan
         self.assertEqual(np.count_nonzero(~np.isnan(result)), 2)
 
+    def test_get_abs_max(self):
```
sfinkens, line 227: Instead of counting how many elements equal -11/5 etc, would it be possible to compare the entire result array with the expectation?

pyresample/bucket/__init__.py
```diff
@@ -296,6 +296,38 @@ def get_max(self, data, fill_value=np.nan, skipna=True):
...
+                Default: True
+
+        Returns
+        -------
+        data : Numpy or Dask array
+            Bin-wise maximums in the target grid
+        """
+        max_ = self.get_max(data, fill_value=fill_value, skipna=skipna)
+        min_ = self.get_min(data, fill_value=fill_value, skipna=skipna)
+        return da.where(-min_ > max_, min_, max_)
```
sfinkens, line 329: I had to think a moment to understand what the `da.where` call is doing. What do you think of moving this to a small method with a self-describing name such as `get_abs_max_from_min_max`? This would also equalize the level of abstraction in this method.

pyresample/bucket/__init__.py
sfinkens: Also, the `fill_value` argument is missing in the docstring.

sfinkens (commented):
Very useful, thanks a lot for adding this! Just a few questions.

pyresample/bucket/__init__.py
gerritholl: Added fill_value in cc377d6fca85c25a447a5410fea24d0536cbb55d.

pyresample/bucket/__init__.py
gerritholl: Done in 70b24c1.

pyresample/test/test_bucket.py
sfinkens: :smile: Ok I see

pnuu (approved):
LGTM, just needs #368 to be merged so the tests can complete.

djhoese (approved):
I assume in the future the `get_max` and `get_min` logic could be combined and allow for more optimized performance on this, but for now I think this looks pretty good.

PR #431, opened 2022-05-06
Add 'antimeridian_mode' to DynamicAreaDefinition

### LINKED ISSUE #428: Add more flexible antimeridian handling to DynamicAreaDefinition
#### Problem description

I'm attempting to replace some old code (Polar2Grid) that would take the equivalent of a DynamicAreaDefinition and produce a static AreaDefinition. In my old code I would "help" the user if they tried to "freeze" the dynamic area with data that crossed the anti-meridian of the projection by modifying the projection and extents so the final result covered the data, not the whole projection space and not a partial of the data. Right now this behavior is not possible with DynamicAreaDefinition.freeze.

I'd like to add a keyword argument to control this logic and call it something like `antimeridian_mode`. This kwarg could be set to 1 of 3 options (4th value `None` as default). See Expected Output below for examples of what these look like.

1. modify_extents/None: This is what is currently done. The X-axis extents are modified to be the left and right of the data where the right-most side has 360 added to it so xmin->xmax is increasing.
2. modify_projection: Includes modify_extents operations but also does `+pm=180` on the PROJ.4 dict. This results in one complete image of the data for simple cases that only stretch the anti-meridian.
3. global_extents: When it is detected that data goes over the poles or antimeridian, the extents are forced to -180/180 longitude.

Thoughts? Concerns?

#### Expected Output

1. modify_extents:

image

2. modify_projection:

image

4. global_extents:

image

#### Alternatives/Other Info

I'm not sure if the `modify_extents` image is cut off and still shows the full image size because of the EWA resampling I'm using or just because that's the way it is supposed to be. ...nope, "nearest" does it too.

I could see using the term "fit" somewhere in the possible keyword values, like "fit_extents" instead of "modify_extents".

djhoese:
Closes #428. See #428 for more details about the possible cases here.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


pyresample/geometry.py
```diff
@@ -1105,10 +1105,16 @@ def _compute_bound_centers_dask(self, proj_dict, lons, lats):
...
             xarr_pos = da.where(xarr >= 0, xarr, np.nan)
             xarr_neg = da.where(xarr < 0, xarr, np.nan)
-            xmin = np.nanmin(xarr_pos)
-            xmax = np.nanmax(xarr_neg) + 360
-            xmin, xmax = da.compute(xmin, xmax)
-        return xmin, ymin, xmax, ymax
+            new_xmin = np.nanmin(xarr_pos)
+            new_xmax = np.nanmax(xarr_neg) + 360
+            if antimeridian_mode == "global_extents":
+                new_xmin, new_xmax = (-180.0, 180.0)
```
djhoese, line 1111: @pnuu @mraspaud I'm looking for help on this case. In this method I'm supposed to return the bounds of the centers of pixels, but -180/180 are the outer extents of the pixels. The only way to know what the centers would be would be to handle the resolution but all the resolution handling is happened in another method after the call to this method.

I could use `None` or something to indicate to the later method "use the maximum extents possible", but that seems pretty error prone. I'm also realizing that I assume in this method that the input geographic (`+proj=latlong`) projection doesn't already have a `+pm=180` which would make the hardcoded -180/180 incorrect.

So my question is, how can I freeze the dynamic area definition to the global extents of the projection without making the code ugly by passing `resolution` down to this method?

lobsiger: @djhoese TBH I do not fully understand what you are resolving here. But it seems to me that this might be related to the PyCoast horizontal scratch problem of cylindrical projections when crossing the dateline that I just solved with PR59. Does PyCoast after the changes you propose here see different dynamic area definition boundaries than what it gets to see now?

djhoese: Pycoast should never see a DynamicAreaDefinition object directly. This code doesn't change if you receive a valid area or not, but rather which of the potential valid areas you get. Do you have a specific Satpy use case that wasn't working?

lobsiger: I had no specific case because TBH I have never tested a dynamic area def that crosses the datelline before. So I was just wondering whether my recent fix of the scratch problem in PyCoast would a) work at all and b) be possibly impaired by what you do here. I add two images of Metop-B that show at least a) with and without my fix to work as expected now. 
Metop-Bsunny-20220510-DAY-2035-natural_with_night_fog-omerc_bb_1
Metop-Bsunny-20220510-DAY-2035-natural_with_night_fog-omerc_bb_2

djhoese: I can't think of a reason why this DynamicAreaDefinition stuff would cause issues in pycoast. The result of using a DynamicAreaDefinition is a regular AreaDefinition. So the result of my work here won't specifically be the problem (the results are all valid areas), but it could reveal that something in pycoast still doesn't work in cases where AreaDefinitions:

1. Have a right-most extent >180 (this is what is currently done).
2. Have a projection with a prime meridian at 180. This is still valid PROJ, so pycoast shouldn't care too much unless it is doing `area_def.crs.is_geographic` and assuming that the prime meridian is 0.

I think those are the only special cases, but those are possible to create from a user perspective with or without this PR.

lobsiger: In my above images PyCoast returns lon_min = 163.5 and lon_max = -154.5. Not sure what happens if lon_max > 180.

djhoese: Sometimes it depends exactly how pycoast is determining those values. Pycoast shouldn't assume much about the area and should be converting X/Y coordinates to normal lon/lat values. In these of these new areas these will also be lon/lat with a prime meridian at 180 instead of metered X/Y projection coordinates so...it should be OK. :crossed_fingers:

lobsiger: To determine whether a (coasts, borders, rivers, ..) Polygon or Line for some projection and area_extent should be drawn or is 'irrelevant' (and potentially harmful !) PyCoast uses the GSHHG bounding boxes. These are strictly within [-180, 180].

pnuu: To the original question: aren't all extents the extreme edges of the pixels/data, not centers?

pyresample/geometry.py
```diff
@@ -1074,11 +1131,16 @@ def _compute_bound_centers_numpy(self, proj_dict, lons, lats):
...
             # cross anti-meridian of projection
-            xmin = np.nanmin(xarr[xarr >= 0])
-            xmax = np.nanmax(xarr[xarr < 0]) + 360
-        return xmin, ymin, xmax, ymax
+            if antimeridian_mode == "global_extents":
+                xmin, xmax = (None, None)
+            else:
+                xmin = np.nanmin(xarr[xarr >= 0])
+                xmax = np.nanmax(xarr[xarr < 0]) + 360
+            if antimeridian_mode == "modify_projection":
```
pnuu, line 1139: Would `shift_prime_median` be more specific?

pyresample/geometry.py
```diff
@@ -1074,11 +1131,16 @@ def _compute_bound_centers_numpy(self, proj_dict, lons, lats):
         y_is_pole = (ymax >= 90 - epsilon) or (ymin <= -90 + epsilon)
         if proj4.crs.is_geographic and x_passes_antimeridian and not y_is_pole:
             # cross anti-meridian of projection
-            xmin = np.nanmin(xarr[xarr >= 0])
-            xmax = np.nanmax(xarr[xarr < 0]) + 360
-        return xmin, ymin, xmax, ymax
+            if antimeridian_mode == "global_extents":
```
pnuu, line 1134: This is the only option that doesn't have an imperative mode, so should it be `use_global_extents`?

pnuu (changes requested):
Couple of inline comments. In general I guess this LGTM, but there should be some documentation at least listing the options.

pyresample/geometry.py
djhoese: @pnuu I've lost track of this conversation, but yes extents are the edges of the pixels, centers are the centers of the pixels. The method being changed here assumes that the lon/lat coordinates contained within the SwathDefinition are the centers of the pixels. I've updated the method to return `None` for the X corner coordinates if they should be considered to be the furthest/full extent of the projection.

pyresample/geometry.py
djhoese: Hm interesting idea. I think I chose "projection" in case other projections in the future may need different types of modifications. For example, if we were trying to support someone using `+x` and `+y` parameters in their projection definition, I'm not sure the `+pm` would be the best thing to modify.

pyresample/geometry.py
```diff
@@ -982,26 +984,56 @@ def pixel_size_y(self):
...
+        Args:
+            corners:
+                4-element sequence representing the outer corners of the
+                region. Note that corners represents the center of pixels,
+                while area_extent represents the edge of pixels. The four
+                values are (xmin_corner, ymin_corner, xmax_corner, ymax_corner).
+                If the x corners are ``None`` then the full extent (area of use)
+                of the projection will be used if defined. A RuntimeError will
+                be raised if the PROJ library does not know the area of use for
+                the projection and it is not a geographic lon/lat projection.
```
gerritholl, line 1005: This is slightly ambiguous.  It can mean that a RuntimeError will be raised if the PROJ library does not know the area of use, or it can mean that a RuntimeError will be raised if th PROJ library does not know the area of use AND the user passed ``None`` as the x corners.

pyresample/geometry.py
```diff
@@ -982,26 +984,56 @@ def pixel_size_y(self):
...
+                Spatial resolution in projection units (typically meters or
+                degrees). If not specified then shape must be provided.
+                If a scalar then it is treated as the x and y resolution. If
+                a tuple then x resolution is the first element, y is the
+                second.
+            shape:
+                Number of pixels in the area as a 2-element tuple. The first
+                is number of rows, the second number of columns.
+            projection:
+                PROJ.4 definition string, dictionary, or pyproj CRS object
```
gerritholl, line 1016: Can it also be an int, interpreted as an EPSG code?  That works in other contexts.

gerritholl (approved):
Just a few clarification questions on the documentation.  Looks good.

pyresample/geometry.py
djhoese: It is the AND case since the area of use is only used if the corners are `None`. I will try to clarify that.

pyresample/geometry.py
djhoese: Hhhhmmm, probably. I guess I should add `int` and then just say in the docs "it'll accept anything pyproj accepts". In the code it actually converts it to a dict if/when it needs to.

pyresample/geometry.py
```diff
@@ -1026,6 +1087,31 @@ def freeze(self, lonslats=None, resolution=None, shape=None, proj_info=None):
...
+                the effect of putting the data on a continuous coordinate
+                system. However, this means that comparing data resampled to
+                this resulting area and an area not over the anti-meridian
+                would be more difficult.
+            * global_extents: Ignore the bounds of the data and use -180/180
+                degrees as the west and east bounds of the data. This will
+                generate a large output area, but with the benefit of keeping
+                the data on the original projection. Note that some resampling
+                methods may produce artifacts when resampling on the edge of
+                the area (the anti-meridian).
```
djhoese, line 1114: @mraspaud @pnuu @gerritholl I think the last remaining question is what are the best names for these parameters. In another comment Panu pointed out that `global_extents` is the only one that doesn't have an imperative verb at the front. He also suggested using "shift_prime_meridian" instead of "modify_projection", but my worry with that is that in the future modifying the projection to support the data may not only include shifting the prime meridian.

Thoughts? Opinions?

gerritholl: `make_extent_global` or `globalize_extent` would be an imperative form of `global_extents`.  But it's not actually global, is it?  It covers all longitudes, but must not necessarily cover all latitudes.  It anyway should affect only the longitude extent, not the latitude extent, shouldn't it?  The docstring in lines 1091 and 1092 mentions lon/lat, but this should be only about lon?

About `modify_projection`: it would seem that it's not accurate to speak of a projection at all here.  When we describe data in lat/lon then the data are unprojected, or are they?

pyresample/geometry.py
```diff
@@ -1026,6 +1087,31 @@ def freeze(self, lonslats=None, resolution=None, shape=None, proj_info=None):
           the shape of the resulting area.
         proj_info:
           complementing parameters to the projection info.
+        antimeridian_mode:
+            How to handle lon/lat data crossing the anti-meridian of the
+            projection. This currently only effects lon/lat geographic
+            projections and data cases not covering the north or south pole.
```
gerritholl, line 1093:
```suggestion
            projections and data cases not covering the prime meridian, so
            it should not cover the north or south pole.
            
```

pyresample/geometry.py
gerritholl: I would replace `projection` by `CRS` and change `modify_projection` to `modify_CRS`.  I can see how a future case of, for example, data covering a pole may modify lat/lons by putting the pole somewhere different, yet retain spherical coordinates therefore still not being technically a projection (but rather a confusion ☺).

pyresample/geometry.py
```diff
@@ -1026,6 +1087,31 @@ def freeze(self, lonslats=None, resolution=None, shape=None, proj_info=None):
           the shape of the resulting area.
         proj_info:
           complementing parameters to the projection info.
+        antimeridian_mode:
+            How to handle lon/lat data crossing the anti-meridian of the
+            projection. This currently only effects lon/lat geographic
```
gerritholl, line 1092:
```suggestion
            projection. This currently only affects lon/lat geographic
```

pyresample/geometry.py
gerritholl: The poles are a coordinate singularity, so lat/lon coordinates are simply not suitable around the poles.

pnuu: I like `modify_crs` (lower case, to be more pythonic). And indeed, I was only thinking about the current case of adjusting the prime meridian.

Similarly `global_extents` would be future proof for any projection type, although what @gerritholl sait about the actual globality is a valid point. Some (hasty!) suggestions for names for this: `use_full_width(_extents)`, `wrap_x_extents`, `cover_all_meridians`.

djhoese: The thing about the "global_extents" not being global is a little odd. Right now it is global in the longitudinal dimension, but not necessarily in latitude. I would hope users wouldn't be too confused that a parameter dealing with the anti-meridian would only effect meridians and not parallels. We could call it `expand_extents` which can give the general impression of "we're going to increase the range of the extents to fit more than just the data".

pyresample/geometry.py
```diff
@@ -1012,7 +1047,33 @@ def compute_domain(self, corners, resolution=None, shape=None):
...
+        aou = self._get_crs_area_of_use(projection)
+        if shape is not None:
+            width = shape[1]
+            x_resolution = (aou.east - aou.west) / width
+            corners[0] = aou.west + x_resolution / 2.0
+            corners[2] = aou.east - x_resolution / 2.0
+        else:
+            x_resolution = resolution[0]
+            corners[0] = aou.west + x_resolution / 2.0
+            corners[2] = aou.east - x_resolution / 2.0
```
mraspaud, line 1063: This looks quite similar to the `if shape` part in `compute_domain`. Could it be factorized, maybe by converting to CRS earlier and setting an area of use if there isn't already one?

pyresample/geometry.py
```diff
@@ -1026,6 +1087,31 @@ def freeze(self, lonslats=None, resolution=None, shape=None, proj_info=None):
           the shape of the resulting area.
         proj_info:
           complementing parameters to the projection info.
+        antimeridian_mode:
+            How to handle lon/lat data crossing the anti-meridian of the
+            projection. This currently only effects lon/lat geographic
+            projections and data cases not covering the north or south pole.
+            The possible options are:
+
+            * modify_extents: Set the X bounds to the edges of the data, but
```
mraspaud, line 1096: Should these parameter names have quotes `"` around them?

pyresample/geometry.py
```diff
@@ -1105,10 +1195,16 @@ def _compute_bound_centers_dask(self, proj_dict, lons, lats):
             # cross anti-meridian of projection
             xarr_pos = da.where(xarr >= 0, xarr, np.nan)
             xarr_neg = da.where(xarr < 0, xarr, np.nan)
-            xmin = np.nanmin(xarr_pos)
-            xmax = np.nanmax(xarr_neg) + 360
-            xmin, xmax = da.compute(xmin, xmax)
-        return xmin, ymin, xmax, ymax
+            new_xmin = np.nanmin(xarr_pos)
+            new_xmax = np.nanmax(xarr_neg) + 360
```
mraspaud, line 1199: This should be placed into the next `else` section.

pyresample/geometry.py
```diff
@@ -1047,20 +1133,19 @@ def freeze(self, lonslats=None, resolution=None, shape=None, proj_info=None):
...
         lons, lats = self._extract_lons_lats(lonslats)
         if hasattr(lons, 'compute'):
-            return self._compute_bound_centers_dask(proj_dict, lons, lats)
-        return self._compute_bound_centers_numpy(proj_dict, lons, lats)
+            return self._compute_bound_centers_dask(proj_dict, lons, lats, antimeridian_mode)
+        return self._compute_bound_centers_numpy(proj_dict, lons, lats, antimeridian_mode)
 
-    def _compute_bound_centers_numpy(self, proj_dict, lons, lats):
-        # TODO: Do more dask-friendly things here
+    def _compute_bound_centers_numpy(self, proj_dict, lons, lats, antimeridian_mode):
```
mraspaud, line 1148: This method and `_compute_bound_centers_dask` have a lot of duplicated code. Could they be refactored to minimise the code duplication?

mraspaud (changes requested):
The logic seems sound, great work! I'd just like to get rid of some code duplication

pyresample/geometry.py
djhoese: I'm not sure I understand the intention with your comment here. The prime meridian isn't really considered when deciding whether or not this keyword is applied.

pyresample/geometry.py
djhoese: I thought about this for a long time when I originally wrote it and the only other option that I liked was to restructure `compute_domain` so it computes `extents` earlier (in which case area of use could be used directly), but I felt like it made the logic more complex and I think you end up with more code. I'm not seeing an easy way to do this.

pyresample/geometry.py
djhoese: Sure.

pyresample/geometry.py
mraspaud: `ignore_extents` instead of `global_extents`? and `modify_crs` sounds good.

pyresample/geometry.py
djhoese: Done. Let me know what you think.

pyresample/geometry.py
djhoese: I spent some time trying to clean this usage up and I just can't find anything that I'm happy with. It can be restructured into multiple if/elif blocks and it cleans up the number of calculations to get the final extent, but makes the logic for everything else really confusing. You end up with duplicate code that if you were to extract it out into shared methods they'd be like one line. I think the current way of having `corners` get "filled in" with the projection extents results in the cleanest overall code.

pyresample/geometry.py
djhoese: Why "ignore"? What is being ignored? Oh you mean ignoring the extents of the data...eh I'm not sure I like that. I'll change to `modify_crs` for that one, but still not sure on extents.

pyresample/geometry.py
```diff
@@ -1140,79 +1140,61 @@ def freeze(self, lonslats=None, resolution=None, shape=None, proj_info=None,
...
         return lons, lats
 
+    def _compute_new_x_corners_for_antimeridian(self, xarr, antimeridian_mode):
+        if antimeridian_mode == "global_extents":
+            xmin, xmax = (None, None)
+        else:
+            if hasattr(xarr, "compute"):
+                xmin, xmax = self._wrap_x_corners_dask(xarr)
+            else:
+                xmin, xmax = self._wrap_x_corners_numpy(xarr)
```
mraspaud, line 1182: How about:

```suggestion
            wrapped_array = xarr % 360
            xmin = np.nanmin(wrapped_array)
            xmax = np.nanmax(wrapped_array)
            if hasattr(wapped_array, "compute"):
                xmin, xmax = da.compute(xmin, xmax)
```

I think it's more concise, maybe at the cost of slightly worse performance?

djhoese: Done.

mraspaud (approved):
LGTM, thanks for the hard work!

pyresample/geometry.py
gerritholl: From my understanding, the problem with data covering the pole stems from all longitudes converging there, so data covers the pole spans the full range of possible longitudes.  There can also be data that stem all longitudes without covering the pole, if we have a satellite in an orbit that's neither polar nor geosynchronous.  For example, the TRMM orbit in this image: https://www.researchgate.net/figure/One-TRMM-swath-intersecting-path-of-Hurricane-Irene-in-2005_fig3_221206961. 

I thought the situation here is that we can handle the antimeridian as long as we have a limited span of longitudes.  More specifically, I understood that we should not have data at ±180° longitude compared to the antimeridian, otherwise we're just recreating the problem we were trying to solve.  Therefore, we can handle the antimeridian as long as data are not covering the prime meridian, correct?

(also, "effects" should be "affects" here)

djhoese: Try refreshing your page, the "affects" change should be there already.

As for the rest of that, at least for what is being checked currently in the code, the 3 things being checked are:

1. Is the CRS geographic (lon/lat)?
2. Are the X min/max greater than 355 degrees apart?
3. Is the Y point beyond +/- -89.9 degrees?

If all of these are true then we enter into the antimeridian handling. So to modify the docstring this way would be incorrect as far as what is actually being checked. If this was saying when `antimeridian_mode` should be expected to work then I think this would be more accurate, but right now it says when it takes effect (affect?). I don't think there is an easy way to check if the data crosses the prime meridian and the antimeridian without adding more calculations. You are correct that data that is already spanning all longitudes but not the poles would result in the antimeridian handling being triggered. By default it would result in a new definition from 0 to 360 longitude. That behavior existed before this PR, but I'm not sure how best to fix it either.

gerritholl: OK, it's complicated :)  Never mind my suggested change then.

djhoese: Thanks for letting me take the easy road :wink: 

I think this could be cleared up in the future, but it is also a special case that may need some real world examples to be the most accurate.

pnuu (approved):
LGTM

PR #435, opened 2022-06-21
Fix SwathDefinition causing unnecessary dask computes when used as a dict key

djhoese:
While debugging some performance issues in my own software that uses pyresample I noticed that using a SwathDefinition in a dict can sometimes causes computation of the underlying dask arrays. This was unknown behavior to me but turns out to be expected. When a dict key hash matches one that already exists, Python's dict will do an `__eq__` check between the two keys to make sure they are actually the same object as it is possible for two different objects to produce the same hash. In our case, SwathDefinition hashes are based on the underlying dask arrays (in the case that dask arrays are used) to avoid hashing them as numpy arrays. In the cases where `__eq__` is called the entire dask array for lon and lat for both swath definitions are computed when compared to each other.

Side note: While working on this I found out that numpy-based arrays were actually being computed as dask arrays. Probably an unnecessary overhead.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -139,3 +139,5 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.pypi_password }}
+          skip_existing: true
+
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -154,15 +154,20 @@ class BaseDefinition:
         if isinstance(other_lons, DataArray) and np.ndarray is not DataArray:
             other_lons = other_lons.data
             other_lats = other_lats.data
+
+        arrs_to_comp = (self_lons, self_lats, other_lons, other_lats)
+        all_dask_arrays = da is not None and all(isinstance(x, da.Array) for x in arrs_to_comp)
+        if all_dask_arrays:
+            # Optimization: We assume that if two dask arrays have the same task name
+            # they are equivalent. This allows for using geometry objects in dict keys
+            # without computing the dask arrays underneath.
+            return self_lons.name == other_lons.name and self_lats.name == other_lats.name
+
         try:
-            from dask.array import allclose
-        except ImportError:
-            from numpy import allclose
-        try:
-            lons_close = allclose(self_lons, other_lons, atol=1e-6, rtol=5e-9, equal_nan=True)
+            lons_close = np.allclose(self_lons, other_lons, atol=1e-6, rtol=5e-9, equal_nan=True)
             if not lons_close:
                 return False
-            lats_close = allclose(self_lats, other_lats, atol=1e-6, rtol=5e-9, equal_nan=True)
+            lats_close = np.allclose(self_lats, other_lats, atol=1e-6, rtol=5e-9, equal_nan=True)
             return lats_close
         except (AttributeError, ValueError):
             return False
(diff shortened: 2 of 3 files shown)


pnuu (approved):
LGTM

.github/workflows/deploy.yaml
```diff
@@ -139,3 +139,5 @@ jobs:
         with:
           user: __token__
           password: ${{ secrets.pypi_password }}
+          skip_existing: true
+
```
gerritholl, line 143: I don't know what this does, so I'll trust you that it's useful and intended :)

pyresample/geometry.py
```diff
@@ -146,25 +146,30 @@ def __eq__(self, other):
...
-        if self_lons is other_lons and self_lats is other_lats:
-            return True
         if isinstance(self_lons, DataArray) and np.ndarray is not DataArray:
             self_lons = self_lons.data
             self_lats = self_lats.data
         if isinstance(other_lons, DataArray) and np.ndarray is not DataArray:
             other_lons = other_lons.data
             other_lats = other_lats.data
+        if self_lons is other_lons and self_lats is other_lats:
+            return True
```
gerritholl, line 156: Why was this block moved from before to after the `isinstance` checks?

pyresample/test/test_geometry.py
```diff
@@ -31,6 +31,7 @@
 from pyresample import geo_filter, geometry, parse_area_file
 from pyresample.geometry import (
     IncompatibleAreas,
+    SwathDefinition,
```
gerritholl, line 34: Why is the import added here?  Don't we usually have imports from the module to be tested inside the test functions/methods, such as we (redundantly?) have in line 2933?

gerritholl (approved):
Just a few non-critical questions from my side, but looks good to me.

pyresample/test/test_geometry.py
djhoese: I would say we "used" to do that, but I haven't found it necessary since using pytest. I believe the main argument for doing the imports inside the test is that `unittest` would fail completely if an import failed. I'm not sure pytest has that problem. Regardless I'm not sure I agree with this practice anymore as it produces a lot of redundant code with no benefit. If the import fails the tests fail in one way or another and I'm ok with that.

pyresample/geometry.py
djhoese: Ah because now it checks if the dask arrays are the same instance. Before it would only check if the DataArrays were the same. This is more likely to catch more cases (ex. two DataArrays with the same dask array underneath).

.github/workflows/deploy.yaml
djhoese: It is intended, but not a huge problem in pyresample given the way our CI is set up. I noticed it while working on another pytroll package. Basically, if PyPI has a specific package already and you try to upload it again the CI job would fail. In any case where this would be a problem we would likely be uploading a new version of a wheel (ex. new version of Python without a new pyresample release, just new wheel build for that version). We don't want the upload to fail just because there are existing wheels/files.

mraspaud (approved):
LGTM

PR #437, opened 2022-06-24
Add GIL releasing to EWA cython code

djhoese:
I realized while debugging performance that none of the EWA cython code was releasing the GIL when it could (it is almost all non-python C++ code). ~This PR also refactors some code that couldn't be cleaned up with the older original versions of cython that were used when it was first developed.~ (nevermind the cleaner code won't work)

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/ewa/_fornav.pyx
+++ b/pyresample/ewa/_fornav.pyx
@@ -275,10 +275,10 @@ def fornav_wrapper(numpy.ndarray[cr_dtype, ndim=2, mode='c'] cols_array,
     if not all(output_array.dtype == out_type for output_array in output_arrays):
         raise ValueError("Input arrays must all be of the same data type")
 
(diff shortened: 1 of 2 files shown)


djhoese:
I tested this at the same time as adding the `nogil` context manager to some python-geotiepoints code I was testing so I'm not sure the exact difference of only this change. The main overall idea can be seen in these resource plots from dask's profilers when generating all of Polar2Grid's default products (true_color + false_color + a ton of regular bands from the reader). Before the nogil changes:

image

And with the nogil changes:

image

I'm not sure this is a perfect comparison as I may have changed some other things too, but the main point is the CPU line in the graph. In both plots you can see that memory generally follows the same pattern, but in the first CPU goes up and down almost never reaching the full 800% (8 dask workers). In the second plot CPU never dips below 600% once it starts going.

PR #438, opened 2022-06-30
Fix using cached LUTs in bilinear resampler

pnuu:
This PR fixes the bilinear resampling when the resampling LUTs have been cached.

The bug were introduced in #324 where a check was added to make sure the data array dimensions are in correct order. The test was done against an index array that only exists during the initial run, and a better comparison is against the shape of the source geo definition.

 - [ ] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/pyresample/bilinear/xarr.py
+++ b/pyresample/bilinear/xarr.py
@@ -164,7 +164,7 @@ class XArrayBilinearResampler(BilinearBase):
         def from_delayed(delayeds, shp):
             return [da.from_delayed(d, shp, np.float32) for d in delayeds]
 
-        data = _check_data_shape(data, self._valid_input_index)
+        data = _check_data_shape(data, self._source_geo_def.shape)
         if data.ndim == 2:
             shp = self.bilinear_s.shape
         else:
@@ -264,10 +264,10 @@ def _get_valid_input_index(source_geo_def,
     return valid_input_index, source_lons, source_lats
 
 
-def _check_data_shape(data, input_idxs):
+def _check_data_shape(data, input_xy_shape):
     """Check data shape and adjust if necessary."""
     # Handle multiple datasets
-    if data.ndim > 2 and data.shape[0] * data.shape[1] == input_idxs.shape[0]:
+    if data.ndim > 2 and data.shape[0] * data.shape[1] == input_xy_shape[0]:
         # Move the "channel" dimension first
         data = da.moveaxis(data, -1, 0)
 


pnuu:
In addition to Coveralls acting up, I can't get the tests to fail. I tried adding

```python
_ = new_resampler.resample(self.data1)
```

to https://github.com/pytroll/pyresample/blob/main/pyresample/test/test_bilinear.py#L1198

As far as I understand, this test pretty much does what Satpy is doing, so it should be failing, but passes with the current `main`.

pnuu:
Also `res = new_resampler.get_sample_from_bil_info(self.data1)` works. This is what Satpy is actually doing, while `.resample()` calls the `.get_bil_info()` method first.

djhoese:
Yeah I would have expected adding `get_sample_from_bil_info` to the line in the test you specified to fail. My best guesses are:

1. You are accidentally running a different version locally than you think you are (not `main` which should break).
2. Something in satpy and/or the tests and/or the resampler itself is setting some attribute on the class itself so it "persists" when you make the new resampler in the test.

pnuu:
I'm fairly certain it's the correct version because my `import ipdb; ipdb.set_trace()` in the test method is caught.

I modified the test so that it saves to a persistent location, then ran again with modification that skipped the saving. Still works fine.

djhoese:
I have to finish my docstring updates for my antimeridian PR but then I'll switch branches and test on main.

pnuu:
I'm starting my holiday now, so I'll get back to this in August at the earliest.

djhoese:
Ok, I've tracked down why the real world case fails and the test doesn't. It is this line of code:

https://github.com/pytroll/pyresample/blob/e531bd9c8bc42213ee128a09842797a1346dee06/pyresample/bilinear/xarr.py#L113

From what I understand this `if` block is only entered if there aren't neighbors for every output pixel (the source data doesn't fully cover the target area). In the test the target def being used is only 4x4 pixels and so `res.size` is equal to the size of the target area (16). So I think the tests need to be expanded for other coverages. I saw that there is another test area that doesn't overlap the input data, but I'm not sure that is "good enough". I'll try a quick test.

djhoese:
Hhhmmm even the outside target area doesn't hit this condition. I'm not sure if it is a smart idea for me to dive further into this until you're back @pnuu. Let me know if you have ideas and I can try them out before then.

pnuu:
Damn, had forgotten this. I'll see if I have time early next week to have a look.

pnuu:
I tested with the target area that doesn't cover the overlap the input area at all with the `main` branch code, and it works. And also enters this branch:

https://github.com/pytroll/pyresample/blob/e531bd9c8bc42213ee128a09842797a1346dee06/pyresample/bilinear/xarr.py#L113

Same with a partially overlapping target area (added with the latest commit to this PR) that gets a sample using pre-calculated LUTs.

djhoese:
Looks like the test failure is an HTTP hiccup in the deploy test (no big deal) and a new failure in the EWA resampling. I guess I know what I'm working on this morning.

pyresample/ewa/ewa.py
```diff
@@ -149,7 +149,7 @@ def fornav(cols, rows, area_def, data_in,
     """
     if isinstance(data_in, (tuple, list)):
         # we can only support one data type per call at this time
-        assert(in_arr.dtype == data_in[0].dtype for in_arr in data_in[1:])
+        assert (in_arr.dtype == data_in[0].dtype for in_arr in data_in[1:])
```
djhoese, line 152:
```suggestion
        for in_arr in data_in[1:]:
           if in_arr.dtype != data_in[0].dtype:
               raise ValueError("All input arrays must be the same dtype")
```

djhoese: I think we're supposed to avoid asserts in non-test code so I figured since you were updating it anyway why don't I just remove the assert.

pyresample/ewa/ewa.py
```diff
@@ -149,7 +149,9 @@ def fornav(cols, rows, area_def, data_in,
     """
     if isinstance(data_in, (tuple, list)):
         # we can only support one data type per call at this time
-        assert(in_arr.dtype == data_in[0].dtype for in_arr in data_in[1:])
+        for in_arr in data_in[1:]:
+           if in_arr.dtype != data_in[0].dtype:
+               raise ValueError("All input arrays must be the same dtype")
```
djhoese, line 154:
```suggestion
        for in_arr in data_in[1:]:
            if in_arr.dtype != data_in[0].dtype:
                raise ValueError("All input arrays must be the same dtype")
```

That's what I get for using github's web editor.

djhoese:
I'm a little scared about the EWA failure. I didn't change anything and it seems to pass now. I guess I'll deal with it next time...

djhoese (approved):
LGTM. Feel free to merge.

pyresample/ewa/ewa.py
pnuu: Oh yeah, I was just fixing the emerging new flake8 warnings. Thanks!

pnuu:
> I'm a little scared about the EWA failure. I didn't change anything and it seems to pass now. I guess I'll deal with it next time...

I was a bit surprised about the failure in the first place, as only thing I changed was the flake8 complaint. But yeah, merging.

PR #447, opened 2022-08-01
Fix handling of lon/lat coordinates on CRS with prime meridian != 0

djhoese:
For a long time we have used a shortcut when transforming lon/lat area coordinates to generic "lon/lat degrees" by checking if the CRS of that area is "geographic" and returning the original input coordinates. This is really only correct in the cases where the CRSes are the same. This shortcut was causing issues in nearest (and other) resamplers where if you resampled to an AreaDefinition with a shifted prime meridian (ex. at 180/-180). In these cases the resampling was using the non-shifted lon/lat coordinates which usually meant you got a black image as a result of resampling. This PR removes this shortcut.

By removing this shortcut this means that the utility/custom Proj class in `pyresample/_spatial_mp.py` is no longer needed as it doesn't actually do anything any more. I will remove this utility class shortly.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/_spatial_mp.py
+++ b/pyresample/_spatial_mp.py
@@ -119,8 +119,6 @@ class Proj(BaseProj):
     def __call__(self, data1, data2, inverse=False, radians=False,
                  errcheck=False, nprocs=1):
         """Transform coordinates to coordinate system except for geographic coordinate systems."""
-        if self.crs.is_geographic:
-            return data1, data2
         return super(Proj, self).__call__(data1, data2, inverse=inverse,
                                           radians=radians, errcheck=errcheck)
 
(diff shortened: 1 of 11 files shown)


gerritholl (approved):
Looks good to me.

djhoese:
Ok, now this PR is done. There were a couple inconvenient usages of the Proj utility class as a stand-in for Proj_MP or Proj where `nprocs` was being passed to either one even though `Proj` doesn't actually support that kwarg and doesn't do anything with it (just blindly accepts it). I'm not a fan of that usage. If we want a class to act as both and only do multiprocessing stuff when nprocs is greater than 1, then we should do that, not have a class that hides keyword usage or not usage.

djhoese:
Also, I ran satpy's tests against this PR and it was fine.

djhoese:
Whoa, unexpected, but this seems to fix nearest's (and possibly bilinear @pnuu ) handling of AreaDefinitions that start on the positive longitudes (0 to 180) and cross the antimeridian going 180-360.

image

Previously, it would just cut this image off at 180.

PR #450, opened 2022-08-05
Area definition html representation for Jupyter notebooks

BENR0:
This adds `_repr_html_` to the `AreaDefinition` class. In Jupyter notebooks therefore instead of a string a nicely formatted text is show together with a map overview (cartopy plot with borders and coastlines) as can be seen in the example below. The map by default is folded right now to save space (for the screenshot I unfolded it).

pyresample_area_html_repr

The way this is setup is heavily influenced by how xarray implements the `_repr_html_`. Basically I used the same directory/ file structure. I also copied some code (the html with the svg icons, the function to read the static data) which I did not attribute yet.

While functionally ready some considerations before this get merged:
 - Apart from some refactoring of the plotting function (removing unecessary code and comments) I was thinking about maybe moving the plotting function into the class?
 - Only the `AreaDefinition` has a html representation as of now. It would be nice if something similar could be done for `SwathDefinition` to make the look and feel more homogeneous.

 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
You may want to try installing the pre-commit hooks. It looks like you've made the style checkers very angry :wink: 

Other than the style issues, what happens if matplotlib and/or cartopy are not installed? These are not hard requirements of pyresample so we should still show something useful if they are not installed.

djhoese (changes requested):
In addition to my previous comment, I think you need to make sure that all of these new non-python files are in the MANIFEST.in so they are included in the distributed package. Lastly, would it be possible to just include the `.css` and `.html` files as strings in the new `formatting_html.py` module? And can we prefix `formatting_html.py` with an underscore so users are tempted to look into it?

This is really cool. Thanks for putting in the time to get it working.

BENR0:
@djhoese Yeah I already saw it :-D. I think most of the complaints are in the plotting which I need to refactor a little anyway.

mraspaud:
@BENR0 what's the status of this? also, will this affect performance of the notebooks?

BENR0:
Yes I think that should be possible in general. I already did some work in that direction and also towards having a representation for `SwathDefinitions`. It's not done yet though. If you want to include  in a release before PCW I could try and refactor what I have to a satisfactory level to be included. Otherwise working on this as well as on the html representation for Scenes is on my list for PCW :-).

djhoese:
Oh then lets leave this as a task for you for the PCW with the goal of merging it by the end of the week.

BENR0:
Current status:
- html representations for Area as well as SwathDefinition is implemented
- For the time being the SwathDefinition will not get an overview plot due to computing the bounding box lon/lats is currently slow if full resolution lon/lat arrays area used (which for example is the default in modis readers due to interpolation). Tests with only tiepoint arrays suggest that it would be fast enough.
- If xarray or cartopy are not installed html representation is still generated but only for the parts which do not rely on those packages. For the map overview and the lon/lat array display in the SwathDefinition case a note is displayed that further rich display can be enabled when those packages are installed.
- There are no performance implications in Scene loading or dataset display since the html representation is only generated when areas are directly called inside a notebook.

pnuu:
Pre-commit is complaining about missing docstrings:
```
pyresample/static/__init__.py:1:1: D104 Missing docstring in public package
pyresample/static/html/__init__.py:1:1: D104 Missing docstring in public package
pyresample/static/css/__init__.py:1:1: D104 Missing docstring in public package
```

pnuu:
The pre-commit jobs is showing that some docstrings are missing.

mraspaud (commented):
This looks good I think. I see that a lot of code is untested, could the pytest plugin `nbmake` be used? eg https://semaphoreci.com/blog/test-jupyter-notebooks-with-pytest-and-nbmake

mraspaud:
Waiting for tests here :)

BENR0:
Ok, so I added some tests for the "cartopy not installed" case. If xarray is not installed I now opted to print the numpy arrays, so there was no need for some extra message for the user. This should work as expected now. There are some things to be noted/ todos:

- `_repr_html_` for the `SwathDefinition` has no Area overview because plotting is to expensive currently in too many cases. This is something I will add again once the G-Ring information is added to the `SwathDefinition`
- There is some improvement possible designwise and codewise for the display of the numpy arrays. I want to refactor that a little later on to make it more similar to the display of the xarray case.
- I am probably going to move the css and the icons inline which will get rid of the extra directories.
- Resolution calculation in the numpy array case is very crude and should be improved/discussed. This might also be something that is superfluous if this is added as metadata/attribute to the `SwathDefinition`.

djhoese:
What's the current state of this PR? Should it wait for the next PCW? Or can I plan to include it in the next minor release (1.28.0)?

BENR0:
@djhoese This can be merged. The overview for the `SwathDefinition` is still missing due to the performance problem. But Martin and I talked about it during PCW and agreed that this is good to go as is and that I will add that at a later point when the problem is solved together with some more refactoring (possibly in conjunction with https://github.com/pytroll/satpy/pull/2167 which should also be good to go and https://github.com/pytroll/satpy/pull/2171).

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,356 @@
...
+    ("pyresample.static.html", "icons_svg_inline.html"),
+    ("pyresample.static.css", "style.css"),
+)
+
+
+@lru_cache(None)
+def _load_static_files():
+    """Lazily load the resource files into memory the first time they are needed."""
+    return [
+        read_binary(package, resource).decode("utf-8")
```
djhoese, line 52: Any idea if `read_binary` suffers from the same performance issues that `pkg_resources` utilities do when used in an environment with a lot of packages installed?

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,356 @@
...
+    # icon_name should be defined in pyresample/static/html/icon-svg-inline.html
+    return (
+        "<svg class='icon pyresample-{0}'>"
+        "<use xlink:href='#{0}'>"
+        "</use>"
+        "</svg>".format(icon_name)
+    )
+
+
+def plot_area_def(area, feature_res="110m", fmt="svg"):
```
djhoese, line 67: Shouldn't this, as a utility function, default to showing the matplotlib figure? Eh I guess it is in the `_formatting_html.py` module so I suppose it is fine.

pyresample/area_config.py
```diff
@@ -849,3 +850,37 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+            if i == 0:
+                area_rep = area_repr(area, include_header=False, include_static_files=False)
+            else:
+                area_rep = area_repr(area, include_header=False)
+
+            content = "\n".join([x.rjust(len(x) + 5) for x in area_rep.split("\n")])
+            area_list.append(template.format(area_name=aname, n="", header_title_length=len(aname),
+                                             content=content))
+        else:
+            pass
```
djhoese, line 884: How about swapping the `if` logic and doing `if not `hasattr(area, "_repr_html_"): continue` and then dedenting the main area repr portion of this?

pyresample/_formatting_html.py
BENR0: Good question. I don't really know. This is probably something that I would move inline in a refactoring step similar to https://github.com/pytroll/satpy/pull/2171. That would also get rid of the directories which I think are not very nice.

pyresample/_formatting_html.py
BENR0: That's still a valid point. I will mark that for the refactoring also.

pyresample/area_config.py
BENR0: Yes, very nice I will change that :-).

pyresample/area_config.py
```diff
@@ -849,3 +850,37 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+
+    for i, (aname, params) in enumerate(_read_yaml_area_file_content(area_file).items()):
+        area = _create_area_def_from_dict(aname, params)
+        if not hasattr(area, "_repr_html_"):
+            continue
+
+        if i == 0:
+            area_rep = area_repr(area, include_header=False, include_static_files=False)
+        else:
+            area_rep = area_repr(area, include_header=False)
```
djhoese, line 880: So static files are included for every row except the first? Or is this logic inverted from what it should be? Also, doesn't the `i == 0` check fail if we were given some swath definitions or other non-AreaDefinition (no `_repr_html_` method) geometries? I guess that's not technically possible right now (an area file with non-areas), but it could be in the future.

djhoese (commented):
pre-commit wants a docstring on all of the `__init__.py` files in the static directory. Could you add those (3 total)?

pyresample/area_config.py
BENR0: Yes you are right about the static files logic. Totally missed that one. Fixed it.

About `i == 0` check: this could be a possible break point if non-areas could be possible in an area file in the future. This is based purely on what is in the area file right now. I don't really know what you have in mind for non-areas so I am not sure what's the smartest way to go here.

BENR0:
> pre-commit wants a docstring on all of the `__init__.py` files in the static directory. Could you add those (3 total)?

Done

mraspaud (commented):
Looks good. I'm just a bit worried by all the codecov notes about untested parts of the code. Maybe adding a simple/quick test for the SwathDefinition will make them go away? I'm aware the swath part is not optimized yet, but if the code is present it has to be tested.

pyresample/area_config.py
djhoese: I'm just thinking if we allowed users to make a SwathDefinition from YAML that loads binary lon/lat array files (an example off the top of my head) that this current logic could break. Given the logic is simple and saying "is this the first area we're printing", maybe the logic could be something not based on the index of the area in the list?

BENR0:
@mraspaud

BENR0:
@mraspaud @mraspaud I added some more tests. They are pretty rudimentary and just check if the generated html string contains tags specific to the feature being tested.

While working on this I stumbled about the resolution calculation for the `SwathDefinition` in the numpy case again There are two points: First (as mentioned earlier in the discussion) this is very basic and probably not very precise and therefore should be refactored (ideas welcome, maybe there is something that can be reused already?). Second in the xarray case it is possible that the representation breaks because currently the assumption is from a `Satpy` view were the Latitude and Longitude `DataArrays` in `SwathDefinitions` get assigned attributes like `name`, `resolution` and `units`. So in the case were a user creates a `SwathDefinition` with `xarray` lats/lons without those attributes the representation throws an error. These attributes could be made mandatory if `SwathDefinitions` are created with `DataArrays` which is a pretty hard constraint I guess or the representation could do more checks and in case these attributes are missing fall back to calculating these as in the numpy case which is "inline" with point one needing some more work. Let me know what you think.

pyresample/test/test_geometry/test_swath.py
```diff
@@ -34,8 +34,8 @@ def _gen_swath_def_xarray_dask(create_test_swath):
     lons, lats = _gen_swath_lons_lats()
     lons_dask = da.from_array(lons)
     lats_dask = da.from_array(lats)
-    lons_xr = xr.DataArray(lons_dask, dims=('y', 'x'))
-    lats_xr = xr.DataArray(lats_dask, dims=('y', 'x'))
+    lons_xr = xr.DataArray(lons_dask, dims=('y', 'x'), attrs={'name': 'Longitude', 'resolution': 500, 'units': 'm'})
+    lats_xr = xr.DataArray(lats_dask, dims=('y', 'x'), attrs={'name': 'Latitude', 'resolution': 500, 'units': 'm'})
```
djhoese, line 38: I'm a little scared about this. The resolution and units I feel like are not guaranteed even in Satpy...eh I guess they should be. I don't know if I'm scared enough (that this breaks checks that are needed) to ask you to change it. I guess leave it for now.

pyresample/area_config.py
```diff
@@ -849,3 +850,37 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+
+    for i, (aname, params) in enumerate(_read_yaml_area_file_content(area_file).items()):
+        area = _create_area_def_from_dict(aname, params)
+        if not hasattr(area, "_repr_html_"):
+            continue
+
+        if i == 0:
+            area_rep = area_repr(area, include_header=False)
+        else:
+            area_rep = area_repr(area, include_header=False, include_static_files=False)
```
djhoese, line 880: Not tested, but I think this makes sense and removes a lot of code:

```suggestion
    for aname, params in _read_yaml_area_file_content(area_file).items():
        area = _create_area_def_from_dict(aname, params)
        if not hasattr(area, "_repr_html_"):
            continue

        area_rep = area_repr(area, include_header=False, include_static_files=not bool(area_list))
```

djhoese (changes requested):
I had an idea for the area list for loop that might clean things up if you agree. Otherwise, do you have screenshots of what the area definition and the swath definition print outs look like (both for numpy and xarray cases?). Does the swath print out change if dask arrays are used (by themselves or inside the xarray DataArray)?

I'm not sure I understood what you were talking about with the resolution issue for swaths.

pyresample/test/test_geometry/test_swath.py
BENR0: Yes I agree. As far as I see this does not break things right now. But I could change it back and write extra Swath generation functions for the repr tests. Actually while adding this made me think again about the situation I described above.

So what I meant was that currently in the xarray case the representation relies on the name, resolution and units attributes being set in the `DataArray`. If that is not the case, for example a user just provides  plain `DataArrays` with lons/lats, the representation currently throws an error.
So either these can be made mandatory, which is a hard constraint for the `SwathDefinition` I guess or the representation gets some more checks to set/calculate these. I hope this makes it clearer.
In the numpy case this is already done now including a very rough calculation of the resolution (https://github.com/BENR0/pyresample/blob/67f9df2862a2527a8142a28f65ff436f65127d38/pyresample/_formatting_html.py#L243-L249).

pyresample/area_config.py
BENR0: I will try that.

BENR0:
Currently the pure `dask.array` and `numpy` case print outs look the same. I think the reason that if `dask.arrays` are input into the `SwathDefinition` the printout still shows them as numpy arrays is the following lines https://github.com/pytroll/pyresample/blob/12892b875221846f0fbc7f68775dbdadd8b68584/pyresample/geometry.py#L802-L804. Anyway this is how it looks like (fixed the resolution decimals by rounding, will push it later):
swath_da_array

The `xarray` case looks like this:
swath_xarray

The `AreaDefinition` looks like this (the map portion is folded initially):
area_def

pyresample/test/test_geometry/test_swath.py
djhoese: FYI the `resolution` in the DataArray is the nominal resolution in the `DataID` and doesn't necessarily match/reflect the current state of the DataArray. After resampling (which is not common for a target swath definition) the resolution could be completely different.

djhoese:
What does the `SSP` mean on the resolution? Single standard parallel?

This looks pretty good to me. In the future we may want to have an optional WKT display and we may want to handle name/ID and description being empty since pyresample 2.0 won't use them by default (it considers them metadata).

BENR0:
SSP = Sub Satellite Point

pyresample/area_config.py
djhoese: I think if this can be added/changed, then I'm good with the rest of the PR given there is refactoring planned.

BENR0: I changed it and this works.

pyresample/_formatting_html.py
BENR0: Done

djhoese:
Looks like there are merge conflicts (and somehow appveyor got run again...wtf). Can you fix up the merge conflicts and see how the rest of CI feels about it?

djhoese:
Sorry for being so late on this. I was a little concerned by how large the coverage loss was. I checked and it looks like the `generate_area_def_rst_list` function isn't covered. Would it be possible to add a test for it?

BENR0:
I was afraid one of you would notice and ask for a test :-D. I always find it difficult to write tests for something like this but I will have a look at it and come up with something. I am not sure if I will be able to do that before my holiday next week though (I have a long train ride I will try to use but no guarantees).

djhoese:
Where are we at with this?

BENR0:
@djhoese sorry for the delay. I added a test now. I am only checking if `area_repr` is called because actual checking of the output seems kind of difficult especially due to the plots being included as svg string. The line when an area in the file does not have a `_repr_html_` attribute does not get checked. I think this is only the case for stacked and dynamic areas which can not be dumped to yaml anyway right? I can't remember the exact cases why this was needed I can just remember some of the areas throwing errors when I tested this way back. Anyway therefore the test is not really complete I guess.

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,356 @@
...
+
+import numpy as np
+
+import pyresample.geometry as geom
+
+try:
+    import cartopy
+    cart = True
+except ModuleNotFoundError:
+    cart = False
```
djhoese, line 32: This is a personal preference, but you could avoid the extra boolean variable (here and for the xarray case) and do `cartopy = None` in the except case.

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,356 @@
...
+
+try:
+    import xarray as xr
+    from xarray.core.formatting_html import _obj_repr, datavar_section
```
djhoese, line 36: Do we have to use a private function (`_obj_repr`)? How does it differ from `repr(xarr_obj)`?

djhoese (changes requested):
A couple suggestions and questions, but otherwise coveralls is still saying the SwathDefinition branch of the formatting is not covered. That and the PNG output option. I see mention of swath definitions in your test so any idea why those branches aren't being covered?

pyresample/_formatting_html.py
BENR0: Unfortunately yes. `repr(xarr_obj)` produces the string representation which is also used as a fallback/ in the non notebook case. I could just "copy" the function into pyresample and make it non private?

pyresample/_formatting_html.py
BENR0: Do you want me to change it?

BENR0:
> A couple suggestions and questions, but otherwise coveralls is still saying the SwathDefinition branch of the formatting is not covered. That and the PNG output option. I see mention of swath definitions in your test so any idea why those branches aren't being covered?

I think the `SwathDefinition` tests just test the html code part and not the overview plot that's why that branch in the plotting function is not covered. I will try and add tests for the plotting function itself or update the test to also cover the `SwathDefinition` branch. The PNG output option could be removed I just left it there as a feature but it's not used in the html representation. Should I remove it then or add a test (even though I am not sure how write a check for that yet :-D).

pyresample/_formatting_html.py
djhoese: And your usage differs from the public `array_repr` and `dataset_repr` in that same `formatting_html` module in xarray (these functions use `_obj_repr`)?

djhoese:
If the PNG stuff isn't used in the final output then I guess I'm ok with it being removed. Although...it might be nice to export an area definition to a PNG. A test wouldn't be too bad as long as you just checked that it was saving and maybe check that it isn't all black. I guess it is up to you what is best for the future. Or maybe @mraspaud has an opinion.

pyresample/_formatting_html.py
djhoese: Sure. It's just me being nitpicky.

BENR0:
The problem with the testing of the plotting function is that I actually don't save the plot to disc but generate the string representation to include it in the html later on. I guess I could check if `savefig` was called  which would be somehow equivalent to checking that something gets saved. I have no idea how to check if the output is all black or not.
/

pyresample/_formatting_html.py
djhoese: Very interesting. Yeah I see what you mean. I was going to say maybe we copy that functionality here, but then noticed it is loading the static icon SVGs so that gets awkward too. I guess that function is small too but it feels weird to copy them.

@mraspaud @pnuu any opinions on this for what "feels" right from the above screenshots? The bottom has so much extra "cruft" for things that aren't used, but maybe that is OK since we are just completely depending on upstream xarray's "private" functions. Or... @BENR0 what about a PR to xarray (in the long term) that let's you exclude empty sections? That would still include the Attributes but maybe that's a good thing? I could see it being beneficial to users to see the extra metadata hanging around the swath definition.

djhoese:
Given that we're basically depending on another program to output the PNG representation I am OK with just knowing that it isn't empty (0 size).

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,361 @@
...
+    """Create html for attribute section based on SwathDefinition.
+
+    Args:
+        area (:class:`~pyresample.geometry.SwathDefinition`): Swath definition.
+
+    Returns:
+        str: String of html.
+
+    Todo:
+        - Improve resolution estimation from lat/lon arrays. Maybe use CoordinateDefinition.geocentric_resolution?
```
djhoese, line 242: Better resolution and boundary information is something I'd say is involved with ghiggi's PRs (not tagging just to avoid spam from the rest of the conversation) related to boundaries. I think swath definitions having a more accurate resolution will have to come later. Something I hope comes to Satpy and pyresample in the future where things like VIIRS and MODIS readers add g-ring/polygon information to the SwathDefinitions and along with that more specific estimates of limb and nadir resolutions. It is added in some readers as metadata to the lons/lats DataArrays but definitely isn't required.

djhoese:
Ah I see you added specific tests that check savefig.

pyresample/_formatting_html.py
BENR0: Yes I think we already talked about that somewhere. Just put it here as a reminder that this could use some improvement because it is not very obvious from the code.

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,361 @@
+#!/usr/bin/env python
+#
+# Copyright (C) 2010-2020 Pyresample developers
+#
```
djhoese, line 4:
```suggestion
# Copyright (C) 2023 Pyresample developers
#
```

pyresample/_formatting_html.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+STATIC_FILES = (
+    ("pyresample.static.html", "icons_svg_inline.html"),
+    ("pyresample.static.css", "style.css"),
+)
+
+
+@lru_cache(None)
+def _load_static_files():
```
djhoese, line 49: Is there any reason for this not to be 1?

pyresample/area_config.py
```diff
@@ -858,3 +859,34 @@ def convert_def_to_yaml(def_area_file, yaml_area_file):
...
+
+
+def generate_area_def_rst_list(area_file):
+    """Create rst list of available area definitions with overview plot.
+
+    Args:
+        area_file (str): Path to area yaml file.
+
+    Returns:
+        str: rst list formatted string.
```
djhoese, line 871: For future reference I think I'd like pyresample to migrate to type annotations and no type information in the docstring.

djhoese (approved):
A couple more comments, but otherwise I approve of this as long as @pnuu and @mraspaud are OK with that comment about importing private functions from xarray.

pyresample/_formatting_html.py
mraspaud: Using a private function/method from another libraries feels wrong I have to say. I agree that the bottom is maybe too verbose, but having the dimension size explicitly here I think is nice actually.

mraspaud (approved):
LGTM. I think the import of the private function from xarray is indeed fragile, as there is no guaranty of that function's future. However, is seems to be something that is helpful, and I would thus suggest opening an issue in xarray to make this function public instead.

mraspaud (approved):
LGTM!

mraspaud:
pre-commit.ci autofix

BENR0:
Opened an issue about the private `_obj_repr` function here: https://github.com/pydata/xarray/issues/8551

PR #454, opened 2022-09-16
Fix import warning in bilinear resampler to mention dask

### LINKED ISSUE #453: Import Error using XArrayBilinearResampler missing failed import of dask

```python
from pyresample import bilinear
```
#### Problem description

Importing bilinear without the required libraries raises the following warning:

```
UserWarning: XArray and/or zarr not found, XArrayBilinearResampler won't be available.
```

This occurs when xarray and zarr are installed (and confirmed with `conda list`).

The actual problem is missing `dask`, but this is missing from the warning, and so is hard to diagnose (without going into the pyresample files themselves).

Installing `dask` fixes the issue.

Would it be possible to change the warning message to something like:
```
UserWarning: XArray, dask and/or zarr not found, XArrayBilinearResampler won't be available.
```
Thanks!

Python 3.9.13 (Anaconda)
OS: Ubuntu

benjaminesse:
Include dask in list of potential missing imports

 - [ ] Closes #453

### DIFF
--- a/pyresample/bilinear/__init__.py
+++ b/pyresample/bilinear/__init__.py
@@ -47,7 +47,7 @@ try:
         XArrayResamplerBilinear,
     )
 except ImportError:
-    warnings.warn("XArray and/or zarr not found, XArrayBilinearResampler won't be available.")
+    warnings.warn("XArray, dask, and/or zarr not found, XArrayBilinearResampler won't be available.")
     XArrayBilinearResampler = None
     CACHE_INDICES = None
     XArrayResamplerBilinear = None


djhoese (approved):
Thanks! This looks good to me.

For future reference for all of the many pull requests you're going to make for pyresample :wink:, I would create a new branch for each pull request (based on `main`). Otherwise, it can become difficult to keep the git history straight between pull requests and pytroll's upstream repository.

Thanks again. I'll merge it when the tests pass.

djhoese:
Closes #453

PR #461, opened 2022-10-13
Reconstruct CRS from WKT only

### LINKED ISSUE #460: Reconstruct CRS from WKT only
## Feature Request

**Is your feature request related to a problem? Please describe.**

For some areas, `area.crs.to_wkt()` returns a dictionary that contains only the `crs_wkt` key and nothing else.  I don't know if this is CF-conform, but this is the responsibility of pyproj.  For some discussion, see the comments on https://github.com/pytroll/satpy/issues/2226.  

For example:

```
In [63]: ar = create_area_def("test", 4087, resolution=1000, width=100, height=100, center=[0, 0])

In [64]: ar.crs.to_cf()
Out[64]: {'crs_wkt': 'PROJCRS["WGS 84 / World Equidistant Cylindrical",BASEGEOGCRS["WGS 84",ENSEMBLE["World Geodetic System 1984 ensemble",MEMBER["World Geodetic System 1984 (Transit)"],MEMBER["World Geodetic System 1984 (G730)"],MEMBER["World Geodetic System 1984 (G873)"],MEMBER["World Geodetic System 1984 (G1150)"],MEMBER["World Geodetic System 1984 (G1674)"],MEMBER["World Geodetic System 1984 (G1762)"],MEMBER["World Geodetic System 1984 (G2139)"],ELLIPSOID["WGS 84",6378137,298.257223563,LENGTHUNIT["metre",1]],ENSEMBLEACCURACY[2.0]],PRIMEM["Greenwich",0,ANGLEUNIT["degree",0.0174532925199433]],ID["EPSG",4326]],CONVERSION["World Equidistant Cylindrical",METHOD["Equidistant Cylindrical",ID["EPSG",1028]],PARAMETER["Latitude of 1st standard parallel",0,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8823]],PARAMETER["Longitude of natural origin",0,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8802]],PARAMETER["False easting",0,LENGTHUNIT["metre",1],ID["EPSG",8806]],PARAMETER["False northing",0,LENGTHUNIT["metre",1],ID["EPSG",8807]]],CS[Cartesian,2],AXIS["easting (X)",east,ORDER[1],LENGTHUNIT["metre",1]],AXIS["northing (Y)",north,ORDER[2],LENGTHUNIT["metre",1]],USAGE[SCOPE["Graticule coordinates expressed in simple Cartesian form."],AREA["World."],BBOX[-90,-180,90,180]],ID["EPSG",4087]]'}
```

Currently, pyresample cannot construct an AreaDefinition using only this CRS and the corresponding coordinates.  Therefore, Satpy's `satpy_cf_nc` reader cannot read the files written by its `cf` writer.  See https://github.com/pytroll/satpy/issues/2226 for the Satpy perspective.  

The NetCDF header may look like:

```
netcdf GOES-16-abi-20221006120020-20221006120952 {
dimensions:
        y = 499 ;
        x = 1000 ;
variables:
        int64 wcm40km ;
                wcm40km:crs_wkt = "PROJCRS[\"WGS 84 / World Equidistant Cylindrical\",BASEGEOGCRS[\"WGS 84\",ENSEMBLE[\"World Geodetic System 1984 ensemble\",MEMBER[\"World Geodetic System 1984 (Transit)\"],MEMBER[\"World Geodetic System 1984 (G730)\"],MEMBER[\"World Geodetic System 1984 (G873)\"],MEMBER[\"World Geodetic System 1984 (G1150)\"],MEMBER[\"World Geodetic System 1984 (G1674)\"],MEMBER[\"World Geodetic System 1984 (G1762)\"],MEMBER[\"World Geodetic System 1984 (G2139)\"],ELLIPSOID[\"WGS 84\",6378137,298.257223563,LENGTHUNIT[\"metre\",1]],ENSEMBLEACCURACY[2.0]],PRIMEM[\"Greenwich\",0,ANGLEUNIT[\"degree\",0.0174532925199433]],ID[\"EPSG\",4326]],CONVERSION[\"World Equidistant Cylindrical\",METHOD[\"Equidistant Cylindrical\",ID[\"EPSG\",1028]],PARAMETER[\"Latitude of 1st standard parallel\",0,ANGLEUNIT[\"degree\",0.0174532925199433],ID[\"EPSG\",8823]],PARAMETER[\"Longitude of natural origin\",0,ANGLEUNIT[\"degree\",0.0174532925199433],ID[\"EPSG\",8802]],PARAMETER[\"False easting\",0,LENGTHUNIT[\"metre\",1],ID[\"EPSG\",8806]],PARAMETER[\"False northing\",0,LENGTHUNIT[\"metre\",1],ID[\"EPSG\",8807]]],CS[Cartesian,2],AXIS[\"easting (X)\",east,ORDER[1],LENGTHUNIT[\"metre\",1]],AXIS[\"northing (Y)\",north,ORDER[2],LENGTHUNIT[\"metre\",1]],USAGE[SCOPE[\"Graticule coordinates expressed in simple Cartesian form.\"],AREA[\"World.\"],BBOX[-90,-180,90,180]],ID[\"EPSG\",4087]]" ;
                wcm40km:long_name = "wcm40km" ;
        double y(y) ;
                y:standard_name = "projection_y_coordinate" ;
                y:units = "m" ;
        double x(x) ;
                x:standard_name = "projection_x_coordinate" ;
                x:units = "m" ;
```

where trying to load the area will fail with

```
  File "/data/gholl/checkouts/satpy/satpy/readers/satpy_cf_nc.py", line 304, in get_area_def
    area = AreaDefinition.from_cf(self.filename)
  File "/data/gholl/mambaforge/envs/py310/lib/python3.10/site-packages/pyresample/geometry.py", line 1790, in from_cf
    return load_cf_area(cf_file, variable=variable, y=y, x=x)[0]
  File "/data/gholl/mambaforge/envs/py310/lib/python3.10/site-packages/pyresample/utils/cf.py", line 474, in load_cf_area
    raise ValueError("Found no AreaDefinitions in this netCDF/CF file.")
ValueError: Found no AreaDefinitions in this netCDF/CF file.
```

**Describe the solution you'd like**

I would like that, if the grid mapping variable contains only the attribute `crs_wkt` and no other information, pyresample should do its best to construct an area using the CRS WKT with other available information.

**Describe any changes to existing user workflow**

There shouldn't be any.

**Additional context**

My current workaround involves writing all lat/lons and using a `SwathDefinition`.  This is inefficient, in particular since I'm then resampling back to the area I had when I was writing the NetCDF file to begin with.

gerritholl:
Load an AreaDefinition from a CF NetCDF file when the grid mapping variable is encoding only the Coordinate Reference System (CRS) in Well Known Text (WKT), but lacks other metadata.

Work in progress.

 - [ ] Closes #460 
 - [x] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/test/test_utils.py
+++ b/pyresample/test/test_utils.py
@@ -683,6 +683,82 @@ class TestLoadCFArea_Public(unittest.TestCase):
         adef, cf_info = load_cf_area(cf_file)
         validate_llnocrs(adef, cf_info)
 
+    def test_load_from_crs_wkt(self):
+        """Test that we can load with CRS WKT only.
+
+        Test that we can reconstruct the area when the grid mapping variable
+        only contains the CRS WKT attribute.  This is the case for
+        equirectangular projections, for example, ``pyproj.CRS(4087).to_cf()``.
+
+        See also https://github.com/pytroll/pyresample/issues/460.
+        """
+        import xarray as xr
+
+        from pyresample import create_area_def
+        from pyresample.utils import load_cf_area
+
+        width = height = 100
+        crs_wkt = """
+PROJCRS["WGS 84 / World Equidistant Cylindrical",
+    BASEGEOGCRS["WGS 84",
+        ENSEMBLE["World Geodetic System 1984 ensemble",
+            MEMBER["World Geodetic System 1984 (Transit)"],
+            MEMBER["World Geodetic System 1984 (G730)"],
+            MEMBER["World Geodetic System 1984 (G873)"],
+            MEMBER["World Geodetic System 1984 (G1150)"],
+            MEMBER["World Geodetic System 1984 (G1674)"],
+            MEMBER["World Geodetic System 1984 (G1762)"],
+            MEMBER["World Geodetic System 1984 (G2139)"],
+            ELLIPSOID["WGS 84",6378137,298.257223563,
+                LENGTHUNIT["metre",1]],
+            ENSEMBLEACCURACY[2.0]],
+        PRIMEM["Greenwich",0,
+            ANGLEUNIT["degree",0.0174532925199433]],
+        ID["EPSG",4326]],
+    CONVERSION["World Equidistant Cylindrical",
+        METHOD["Equidistant Cylindrical",
+            ID["EPSG",1028]],
+        PARAMETER["Latitude of 1st standard parallel",0,
+            ANGLEUNIT["degree",0.0174532925199433],
+            ID["EPSG",8823]],
+        PARAMETER["Longitude of natural origin",0,
+            ANGLEUNIT["degree",0.0174532925199433],
+            ID["EPSG",8802]],
+        PARAMETER["False easting",0,
+            LENGTHUNIT["metre",1],
+            ID["EPSG",8806]],
+        PARAMETER["False northing",0,
+            LENGTHUNIT["metre",1],
+            ID["EPSG",8807]]],
+    CS[Cartesian,2],
+        AXIS["easting (X)",east,
+            ORDER[1],
+            LENGTHUNIT["metre",1]],
+        AXIS["northing (Y)",north,
+            ORDER[2],
+            LENGTHUNIT["metre",1]],
+    USAGE[
+        SCOPE["Graticule coordinates expressed in simple Cartesian form."],
+        AREA["World."],
+        BBOX[-90,-180,90,180]],
+    ID["EPSG",4087]]
+"""
+        ds = xr.Dataset(
+            {"data": (
+                ('y', 'x'),
+                np.zeros((width, height)),
+                {"grid_mapping": "area"}),
+             "area": (
+                (),
+                 0,
+                 {"crs_wkt": crs_wkt, "long_name": "test"})},
+            coords={'x': np.linspace(-7049500, -6950500, width),
+                    'y': np.linspace(-49500, 49500, height)})
+        (adef, cf_info) = load_cf_area(ds)
+        assert adef == create_area_def(
+            "test", 4087, resolution=1000, width=100,
+            height=100, center=(-7000000, 0))
+
 
 class TestLoadCFArea_Private(unittest.TestCase):
     """Test private routines involved in loading an AreaDefinition from netCDF/CF files."""


ghiggi:
@gerritholl pin me when this is merged or mostly done so that I can test it on some of my use cases. 
In meantime, I am drafting a PR refactoring the CFWriter internals to then create:
- a CF-Compliant xr.Dataset (cleaner than the current `scn.to_xarray_dataset()`)
- a CF-Compliant `ZarrWriter` 
- a xr.Dataset satpy accessor to generate satpy scene from xr.Dataset.

I will wait your PR before proceeding ....

ghiggi:
Side note:
If the Satpy Scene is on a SwathDefinition and we use the CFWriter, no `grid_mapping` attribute nor `crs` coords information is attached to netCDF. Maybe you can enforce the correct behavior in the context of this PR. 
Going further away, when remapping a Scene from AreaDef to a SwathDef, the `grid_mapping` attribute should also be updated  ...

gerritholl:
@ghiggi I am putting the work for this PR on hold.  For what I need to achieve, I can and will use https://github.com/pytroll/satpy/pull/2236 instead.  Therefore, fixing https://github.com/pytroll/satpy/issues/2226 for the equirectangular projection is no longer a priority from my side.

djhoese:
@TomLav I disagree with this.

>  If data producers omit the single-property grid mapping attributes in favour of the crs_wkt attribute

This is specific to data **producers**, not **consumers**. We are consumers in this case. If the WKT does not match the single-property CF attributes then that is a fault of the data producer or of the CF format as a whole.

I think the above point is pretty clear, but I'd also like to add that I strongly disagree with CF's choice of making these separate CF attributes and for making crs_wkt optional and supplemental. WKT is fully descriptive and the CF attributes are not. The CF attributes (and the CF standard) do not support every projection and the standard as a whole has to be updated to support new projections when they come along. Conversations in the xarray ecosystem and the OGC Zarr world have started leaning towards the idea of being CF-compatible-ish because of this and only using the CRS WKT attribute in their zarr standards.

TomLav:
Hei @djhoese, I am not unaware of the history and tensions between the different communities around the CF attributes and the WKT definitions. Maybe these things will converge with time.

I was not saying that our from_cf() should bark and not load the crs from WKT, but I do think that this should not happen silently either. I do not know if this should be controlled via a keyword or a warning, but I think there is value for a routine called from_cf() to have the option to strictly follow the CF convention. We can e.g. have a from_cf(strict=True|False).

This being said pyresample [load_cf_area()](https://github.com/pytroll/pyresample/blob/48f10d3c122b62f62640dbead57fba94e836a1a9/pyresample/utils/cf.py#L81) only checks if the :grid_mapping_name attribute is present and has a valid value, it then diverts to pyproj.from_cf() the decoding of the crs attributes and/or wkt. If pyproj adopts a loose approach, pyresample will inherit it silently.

You also state that you disagree with a quote that I took from the CF convention, which is fine, but there is little point discussing the CF convention here.

djhoese:
> You also state that you disagree with a quote that I took from the CF convention, which is fine, but there is little point discussing the CF convention here.

No, sorry. I disagree with your interpretation of the quote. The standard is saying that we can load from WKT without looking at the single-property attributes (as consumers of the data, not producers).

> but I do think that this should not happen silently either.

Why not? The CF convention gives us the `crs_wkt` attribute as an option. We can use it if it is there. There is no need to warn the user if we use it or don't use it. It is there to be used (and is more accurate). 

> to have the option to strictly follow the CF convention

We are following the CF convention by being a consumer of the data and using the `crs_wkt` attribute. There is nothing against the CF convention by doing this.

TomLav:
I see. We disagree on the interpretation of the CF convention.

You understand from the convention that files where the CRS is only specified with wkt are CF compliant.

I understand the convention that the CRS must first be defined in their single-attribute form, then we can duplicate and enrich with WKT.

For reference, my interpretation stems from the sentence: _The crs_wkt attribute is intended to act as a supplement to other single-property CF grid mapping attributes (as described in Appendix F); it is not intended to replace those attributes._ What follows with data producers does not change my interpretation of this first sentence.

I don't think we can settle this from Pytroll, this must have been discussed somewhere at CF, I will give this a look.

For the record: if it is so that CF allows a crs to only be defined by its WKT then of course from_cf() should silently do its job and load this. If CF does not allow this, I still think we should help the user and load the CRS, but we should also give the user the control to use non-strict checks. Do you agree?

djhoese:
> You understand from the convention that files where the CRS is only specified with wkt are CF compliant.
>
> I understand the convention that the CRS must first be defined in their single-attribute form, then we can duplicate and enrich with WKT.

Nope. I understand that **PRODUCERS** must always include the single-property attributes. The CRS WKT is supplemental.

BUT there is nothing saying you/we/pyresample, as **CONSUMERS** of the data, should not use the CRS WKT. If it is there, we can use it. The WKT will never be less accurate than the CF attributes. So if the CRS WKT and single-property attributes are both available, we should always use/prefer the WKT...as CONSUMERS.

TomLav:
For later reference, it seems that CF (as of CF-1.10) will not accept a crs variable defined only by its wkt description.'

If the text of the convention is possibly ambiguous (see discussion above), the [conformance document](https://github.com/cf-convention/cf-conventions/blob/fd908ce031d1c2afdc59004b3d01ab8c3ef5683d/conformance.adoc?plain=1#L310) is clear:

_The grid mapping variables must have the grid_mapping_name attribute. The legal values for the grid_mapping_name attribute are contained in Appendix F._

Which still leaves us all latitude as to what pyresample should do in such cases. As I stated above I believe it should try to find its way around and produce the area_def. But I wanted to clarify (at least for my own sake) what was CF-compliant or not(*).

I see that pyproj.CRS.to_cf() happily loads the CRS whether its definition uses only the WKT description (case B below), or both the CF-attribute and WKT descriptions (case A), or only the CF-attribute description (case C). The current behaviour of pyproj seems to be to load the CRS from crs_wkt if present (A and B) and decode the CF attributes only if there are no crs_wkt (case C). I think this is the behaviour you want for pyresample, @djhoese, so all good.
```
from copy import copy
from pyproj import CRS

a = area_def.crs.to_cf()
b = {'crs_wkt': a['crs_wkt']}
c = copy(a)
c.pop('crs_wkt')

crsA = CRS.from_cf(a)
crsB = CRS.from_cf(b)
crsC = CRS.from_cf(c)

print(crsA)
print(crsB)
print(crsC)
```

Thus, it should be easy for pyresample load_cf_area() to work smoothly in all 3 cases, we just need to decide how to warn/inform the user that we found a non-compliancy but marched on (in case B).

Turning [this](https://github.com/pytroll/pyresample/blob/48f10d3c122b62f62640dbead57fba94e836a1a9/pyresample/utils/cf.py#L95) `raise` into a warning might be all we need to do.

(*) of course the conformance document could be wrong and not be aligned with CF convention. But what counts to me is that the conformance checkers will implement the conformance document.

djhoese:
As mentioned on slack from our discussion yesterday and today, after looking at the pyresample CF code it is clear pyresample is doing a lot more validation of CF standards than I thought it was. In those cases it makes sense to have a `strict` keyword argument (I'd prefer a default of `False` I think) that essentially "guesses" at whatever information it needs (which variable is the x coordinate, which one the y, etc). I still didn't agree with the idea that the WKT should not be used as the primary source of CRS information if it exists. Meaning the single-property attributes would only be depended on if the `crs_wkt` didn't exist. I don't think this is wrong in a CF-sense and I think it makes more sense as WKT will never be less accurate than the CF attributes. It is just an unfortunate situation that the CF standard has put us in. So in summary a `strict` keyword argument that when True checks if there is only a `crs_wkt` and raises an error would be OK with me. If `False` then pyresample just tries its best to get the CRS information. In general we will depend on pyproj to parse the CF grid mapping variable and produce whatever the best CRS is, but we can have additional checks for CF validation when `strict=True`.

I would be ok with changing that `raise` to a warning.

PR #463, opened 2022-10-24
Fix EWA default for 'weight_delta_max' to match docstring

djhoese:
As described in https://github.com/pytroll/python-geotiepoints/pull/38 I discovered that the default `weight_delta_max` in the dask EWA code did not match the docstring. The docstring is correct for MODIS data and matches what has been in the EWA algorithm forever (it was originally written for MODIS code).

I really don't want to add a test for this.

 - [ ] Tests added 
 - [ ] Tests passed 


djhoese:
Wow, turns out it was wrong in the legacy resampler too. Fixed in both now.

mraspaud (approved):
LGTM

PR #465, opened 2022-11-08
[Future Spherical Class] Add SPoint and SMultiPoint

ghiggi:
This PR add the SPoint and SMultiPoint class of the future spherical geometry interface.

 - [x] Tests added 
 - [x] Tests passed 

### DIFF
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2013 - 2021 Pyresample developers
+# Copyright (c) 2022 Pyresample developers
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -136,7 +136,7 @@ class SMultiPoint(SCoordinate):
 
     def __repr__(self):
         """Get simplified representation of lon/lat arrays in degrees."""
-        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
+        vertices = np.rad2deg(self.vertices)
         return str(vertices)
 
     def plot(self, ax=None, **plot_kwargs):
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -22,7 +22,7 @@ from pyresample.spherical import SCoordinate
 
 
 class SPoint(SCoordinate):
-    """Define a point on the sphere.
+    """Object representing a single point on a sphere.
 
     The ``lon`` and ``lat`` coordinates must be provided in radians.
     """
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -68,7 +68,7 @@ class SMultiPoint(SCoordinate):
         return str(self.vertices)
 
     def __repr__(self):
-        """Get simplified representation of lon/lat arrays in degrees."""
+        """Get simplified representation of lon/lat arrays in radians."""
         return str(self.vertices)
 
     def to_shapely(self):
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -64,7 +64,7 @@ class SMultiPoint(SCoordinate):
         return np.allclose(self.lon, other.lon) and np.allclose(self.lat, other.lat)
 
     def __str__(self):
-        """Get simplified representation of lon/lat arrays in degrees."""
+        """Get simplified representation of lon/lat arrays in radians."""
         return str(self.vertices)
 
     def __repr__(self):
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -22,7 +22,7 @@ from pyresample.spherical import SCoordinate
 
 
 class SPoint(SCoordinate):
-    """Spherical Point.
+    """Define a point on the sphere.
 
     The ``lon`` and ``lat`` coordinates must be provided in radians.
     """
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -42,7 +42,7 @@ class SPoint(SCoordinate):
 
 
 class SMultiPoint(SCoordinate):
-    """Create a SMultiPoint object."""
+    """Object representing multiple points on a sphere."""
 
     def __init__(self, lon, lat):
         lon = np.asarray(lon)
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -131,7 +131,7 @@ class SMultiPoint(SCoordinate):
 
     def __str__(self):
         """Get simplified representation of lon/lat arrays in degrees."""
-        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
+        vertices = np.rad2deg(self.vertices)
         return str(vertices)
 
     def __repr__(self):
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2013 - 2021 Pyresample developers
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Define SPoint and SMultiPoint class."""
+import numpy as np
+
+from pyresample.spherical import SCoordinate
+
+
+def _plot(vertices, ax=None, **plot_kwargs):
+    """Plot the SPoint/SMultiPoint using Cartopy.
+
+    Assume vertices to in radians.
+    """
+    import matplotlib.pyplot as plt
+    try:
+        import cartopy.crs as ccrs
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError("Install cartopy to plot spherical geometries.")
+
+    # Create figure if ax not provided
+    ax_not_provided = False
+    if ax is None:
+        ax_not_provided = True
+        proj_crs = ccrs.PlateCarree()
+        fig, ax = plt.subplots(subplot_kw=dict(projection=proj_crs))
+
+    # Plot Points
+    ax.scatter(x=np.rad2deg(vertices[:, 0]),
+               y=np.rad2deg(vertices[:, 1]),
+               **plot_kwargs)
+
+    # Beautify plot by default
+    if ax_not_provided:
+        ax.stock_img()
+        ax.coastlines()
+        gl = ax.gridlines(draw_labels=True, linestyle='--')
+        gl.xlabels_top = False
+        gl.ylabels_right = False
+
+    return ax
+
+
+def check_lon_validity(lon):
+    """Check longitude validity."""
+    if np.any(np.isinf(lon)):
+        raise ValueError("Longitude values can not contain inf values.")
+
+
+def check_lat_validity(lat):
+    """Check latitude validity."""
+    if np.any(np.isinf(lat)):
+        raise ValueError("Latitude values can not contain inf values.")
+    if np.any(np.logical_or(lat > np.pi / 2, lat < -np.pi / 2)):
+        raise ValueError("Latitude values must range between [-pi/2, pi/2].")
+
+
+class SPoint(SCoordinate):
+    """Spherical Point.
+
+    The ``lon`` and ``lat`` coordinates must be provided in radians.
+    """
+
+    def __init__(self, lon, lat):
+        lon = np.asarray(lon)
+        lat = np.asarray(lat)
+        if lon.size > 1 or lat.size > 1:
+            raise ValueError("Use SMultiPoint to define multiple points.")
+        check_lon_validity(lon)
+        check_lat_validity(lat)
+
+        super().__init__(lon, lat)
+
+    @property
+    def vertices(self):
+        """Return SPoint vertices in a ndarray of shape [1,2]."""
+        return np.array([self.lon, self.lat])[None, :]
+
+    def plot(self, ax=None, **plot_kwargs):
+        """Plot the SPoint using Cartopy."""
+        ax = _plot(self.vertices, ax=ax, **plot_kwargs)
+        return ax
+
+    def to_shapely(self):
+        """Convert the SPoint to a shapely Point (in lon/lat degrees)."""
+        from shapely.geometry import Point
+        point = Point(*np.rad2deg(self.vertices[0]))
+        return point
+
+
+class SMultiPoint(SCoordinate):
+    """Create SPoint or SMultiPoint object."""
+
+    def __new__(cls, lon, lat):
+        """Create SPoint or SMultiPoint object."""
+        lon = np.asarray(lon)
+        lat = np.asarray(lat)
+
+        # If a single point, define SPoint
+        if lon.ndim == 0:
+            return SPoint(lon, lat)  # TODO: TO BE DEFINED
+        # Otherwise a SMultiPoint
+        else:
+            return super().__new__(cls)
+
+    def __init__(self, lon, lat):
+        super().__init__(lon, lat)
+
+    @property
+    def vertices(self):
+        """Return SMultiPoint vertices in a ndarray of shape [n,2]."""
+        return np.vstack((self.lon, self.lat)).T
+
+    def __eq__(self, other):
+        """Check equality."""
+        return np.allclose(self.lon, other.lon) and np.allclose(self.lat, other.lat)
+
+    def __str__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
+        return str(vertices)
+
+    def __repr__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
+        return str(vertices)
+
+    def plot(self, ax=None, **plot_kwargs):
+        """Plot the SMultiPoint using Cartopy."""
+        ax = _plot(self.vertices, ax=ax, **plot_kwargs)
+        return ax
+
+    def to_shapely(self):
+        """Convert the SMultiPoint to a shapely MultiPoint (in lon/lat degrees)."""
+        from shapely.geometry import MultiPoint
+        point = MultiPoint(np.rad2deg(self.vertices))
+        return point
--- a/pyresample/future/spherical/point.py
+++ b/pyresample/future/spherical/point.py
@@ -21,54 +21,6 @@ import numpy as np
 from pyresample.spherical import SCoordinate
 
 
-def _plot(vertices, ax=None, **plot_kwargs):
-    """Plot the SPoint/SMultiPoint using Cartopy.
-
-    Assume vertices to in radians.
-    """
-    import matplotlib.pyplot as plt
-    try:
-        import cartopy.crs as ccrs
-    except ModuleNotFoundError:
-        raise ModuleNotFoundError("Install cartopy to plot spherical geometries.")
-
-    # Create figure if ax not provided
-    ax_not_provided = False
-    if ax is None:
-        ax_not_provided = True
-        proj_crs = ccrs.PlateCarree()
-        fig, ax = plt.subplots(subplot_kw=dict(projection=proj_crs))
-
-    # Plot Points
-    ax.scatter(x=np.rad2deg(vertices[:, 0]),
-               y=np.rad2deg(vertices[:, 1]),
-               **plot_kwargs)
-
-    # Beautify plot by default
-    if ax_not_provided:
-        ax.stock_img()
-        ax.coastlines()
-        gl = ax.gridlines(draw_labels=True, linestyle='--')
-        gl.xlabels_top = False
-        gl.ylabels_right = False
-
-    return ax
-
-
-def check_lon_validity(lon):
-    """Check longitude validity."""
-    if np.any(np.isinf(lon)):
-        raise ValueError("Longitude values can not contain inf values.")
-
-
-def check_lat_validity(lat):
-    """Check latitude validity."""
-    if np.any(np.isinf(lat)):
-        raise ValueError("Latitude values can not contain inf values.")
-    if np.any(np.logical_or(lat > np.pi / 2, lat < -np.pi / 2)):
-        raise ValueError("Latitude values must range between [-pi/2, pi/2].")
-
-
 class SPoint(SCoordinate):
     """Spherical Point.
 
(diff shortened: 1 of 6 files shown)


ghiggi:
@djhoese @mraspaud this is ready for review.
Tomorrow I have faculty meetings, but I can finalize the last details on Friday if you manage to review it tomorrow.

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,151 @@
...
+        """Create SPoint or SMultiPoint object."""
+        lon = np.asarray(lon)
+        lat = np.asarray(lat)
+
+        # If a single point, define SPoint
+        if lon.ndim == 0:
+            return SPoint(lon, lat)  # TODO: TO BE DEFINED
+        # Otherwise a SMultiPoint
+        else:
+            return super().__new__(cls)
```
mraspaud, line 118: Please use a factory function instead of this. I will be confusing for a user to instantiate a `SMulitPoint` and get an `SPoint`. See an example here: https://github.com/pytroll/pyresample/blob/main/pyresample/slicer.py#L41-L51

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __init__(self, lon, lat):
+        super().__init__(lon, lat)
```
mraspaud, line 121: This shouldn't be needed

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __repr__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
+        return str(vertices)
+
+    def plot(self, ax=None, **plot_kwargs):
+        """Plot the SMultiPoint using Cartopy."""
+        ax = _plot(self.vertices, ax=ax, **plot_kwargs)
+        return ax
```
mraspaud, line 145: This is the same as in SPoint and could thus be moved in the base class

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,151 @@
...
+        return np.allclose(self.lon, other.lon) and np.allclose(self.lat, other.lat)
+
+    def __str__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
+        return str(vertices)
+
+    def __repr__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
```
mraspaud, line 139:
```suggestion
        vertices = np.rad2deg(self.vertices)
```

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,151 @@
...
+        """Return SMultiPoint vertices in a ndarray of shape [n,2]."""
+        return np.vstack((self.lon, self.lat)).T
+
+    def __eq__(self, other):
+        """Check equality."""
+        return np.allclose(self.lon, other.lon) and np.allclose(self.lat, other.lat)
+
+    def __str__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(np.vstack((self.lon, self.lat)).T)
```
mraspaud, line 134:
```suggestion
        vertices = np.rad2deg(self.vertices)
```

pyresample/spherical.py
```diff
@@ -34,6 +34,62 @@ def _unwrap_radians(val, mod=np.pi):
     return (val + mod) % (2 * mod) - mod
 
 
+def _xyz_to_vertices(x, y, z):
```
mraspaud, line 37: Could you add a docstring? I don't understand what this is used for.

pyresample/spherical.py
```diff
@@ -42,14 +98,16 @@ class SCoordinate(object):
     """
 
     def __init__(self, lon, lat):
-        if np.isfinite(lon):
-            self.lon = float(_unwrap_radians(lon))
-        else:
-            self.lon = float(lon)
+        lon = np.asarray(lon, dtype=np.float64)
```
mraspaud, line 101: It looks like the check for finite values is gone, why is this not needed anymore?

pyresample/spherical.py
```diff
@@ -125,12 +196,18 @@ def norm(self):
...
     def normalize(self):
         """Normalize the vector."""
-        self.cart /= np.sqrt(np.einsum('...i, ...i', self.cart, self.cart))
-
-        return self
+        # Note: if self.cart == [0,0,0], norm=0 and cart becomes [nan, nan, nan]
+        norm = self.norm()
+        cart = self.cart / norm[..., np.newaxis]  # enable vectorization
+        # mask_norm_0 = norm == 0
+        # cart[mask_norm_0, ...] = 0 # This remove nan propagation
```
mraspaud, line 203: This is not needed and can be removed.

pyresample/spherical.py
```diff
@@ -125,12 +196,18 @@ def norm(self):
...
+        return CCoordinate(cart)
```
mraspaud, line 204: The original method normalizes in place, this is changing the legacy behaviour.

pyresample/spherical.py
```diff
@@ -125,12 +196,18 @@ def norm(self):
 
     def normalize(self):
         """Normalize the vector."""
-        self.cart /= np.sqrt(np.einsum('...i, ...i', self.cart, self.cart))
-
-        return self
+        # Note: if self.cart == [0,0,0], norm=0 and cart becomes [nan, nan, nan]
+        norm = self.norm()
```
mraspaud, line 200: Thanks for refactoring :)

pyresample/spherical.py
```diff
@@ -176,9 +253,16 @@ def __rmul__(self, other):
...
-                           np.arcsin(self.cart[2]))
+        """Convert to SCoordinate object."""
+        if self.cart.ndim == 1:
+            lon = np.arctan2(self.cart[1], self.cart[0])
+            lat = np.arcsin(self.cart[2])
+        else:
+            lon = np.arctan2(self.cart[:, 1], self.cart[:, 0])
+            lat = np.arcsin(self.cart[:, 2])
+        # TODO: in future this should point to SPoint or SMultiPoint
+        # - This is not used within the file.
```
mraspaud, line 264: Can it be done now?

pyresample/test/test_spherical.py
```diff
@@ -146,12 +146,6 @@ def test_equality_at_antimeridian(self):
         point_end = SCoordinate(np.pi, 0)
         assert point_start == point_end
 
-    def test_equality_of_infinites(self):
-        """Test that infinite coordinates are equal."""
-        coord1 = SCoordinate(np.inf, np.inf)
-        coord2 = SCoordinate(np.inf, np.inf)
-        assert coord1 == coord2
-
```
mraspaud, line 154: Why is this not needed anymore?

mraspaud (changes requested):
Thanks for taking on the job of renovating this part of Pyresample, it is highly needed! Some suggestions, questions and comments inline.

pyresample/spherical.py
ghiggi: [...,1]

ghiggi: Would introduce backward incompatibility.

pyresample/spherical.py
ghiggi: I added an inplace argument

pyresample/future/spherical/point.py
ghiggi: This has been moved together with vertices properties to SCoordinate.

pyresample/spherical.py
```diff
@@ -194,14 +268,23 @@ def norm(self):
...
-        # mask_norm_0 = norm == 0
-        # cart[mask_norm_0, ...] = 0 # This remove nan propagation
-        return CCoordinate(cart)
+        norm = norm[..., np.newaxis]  # enable vectorization
+        if inplace:
+            self.cart /= norm
+            return None
+        else:
+            cart = self.cart / norm
+            return CCoordinate(cart)
```
mraspaud, line 287:
```suggestion
        cart = self.cart / norm
        return CCoordinate(cart)
```

mraspaud (approved):
Thanks, starts to look good.

pyresample/spherical.py
djhoese: @mraspaud do we need backwards compatibility with this? I'm fine with having a flag for now, but I think in general inplace operations feel like bad practice and should be avoided, especially by default.

pyresample/spherical.py
djhoese: Can we put this stuff under a feature flag so there is some pyresample global option (environment variable if we have to) to opt-in to these new features?

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,72 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Define SPoint and SMultiPoint class."""
```
djhoese, line 18: Is there a more descriptive definition of this category of objects?

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2013 - 2021 Pyresample developers
```
djhoese, line 4:
```suggestion
# Copyright (c) 2022 Pyresample developers
```

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,72 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Define SPoint and SMultiPoint class."""
+import numpy as np
+
+from pyresample.spherical import SCoordinate
+
+
+class SPoint(SCoordinate):
+    """Spherical Point.
```
djhoese, line 25: Point on a sphere?

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def to_shapely(self):
+        """Convert the SPoint to a shapely Point (in lon/lat degrees)."""
+        from shapely.geometry import Point
+        point = Point(*np.rad2deg(self.vertices[0]))
+        return point
+
+
+class SMultiPoint(SCoordinate):
+    """Create a SMultiPoint object."""
```
djhoese, line 45: "Multiple spherical points" might be better? Or "Object representing multiple points on a sphere"?

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __str__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(self.vertices)
+        return str(vertices)
+
+    def __repr__(self):
+        """Get simplified representation of lon/lat arrays in degrees."""
+        vertices = np.rad2deg(self.vertices)
+        return str(vertices)
```
djhoese, line 66: How much do we want to preserve backwards compatibility? I personally had a lot of trouble with the conversion being done in the str/repr methods when debugging things. I could never be sure if I was looking at the values I wanted to be looking at or if the values were being converted and printed.

If there is utility in having access to the converted vertices maybe there should be a `vertices_degrees` property or something.

djhoese: Yeah there are 3 spots in this class that do `np.rad2deg(self.vertices)`. If it isn't a public property then it should at least be a private one (`_` prefix).

pyresample/spherical.py
```diff
@@ -34,22 +34,121 @@ def _unwrap_radians(val, mod=np.pi):
     return (val + mod) % (2 * mod) - mod
 
 
+def _xyz_to_vertices(x, y, z):
+    """Create vertices array from x,y,z values or vectors.
+
+    If x, y, z are np.values, it create a 1x3 np.array.
```
djhoese, line 40: Is the other common term for this a "scalar array"?

pyresample/spherical.py
```diff
@@ -34,22 +34,121 @@ def _unwrap_radians(val, mod=np.pi):
...
+    else:
+        vertices = np.vstack([x, y, z]).T
+    return vertices
+
+
+def _ensure_is_array(arr):
+    """Ensure that a possible np.value input is converted to np.array."""
+    if arr.ndim == 0:
+        arr = np.array([arr])
+    return arr
```
djhoese, line 54: What's the difference between this and `np.asarray` (or even `np.asanyarray`)?

pyresample/spherical.py
```diff
@@ -34,22 +34,121 @@ def _unwrap_radians(val, mod=np.pi):
...
+
+
+def _vincenty_matrix(lon, lat, lon_ref, lat_ref):
+    """Compute a distance matrix using Vincenty formula.
+
+    The result must be multiplied by Earth radius to obtain distance in m or km.
+    The returned distance matrix has shape (n x n_ref).
+    """
```
djhoese, line 62: Can the units of these inputs and outputs be documented.

pyresample/spherical.py
```diff
@@ -34,22 +34,121 @@ def _unwrap_radians(val, mod=np.pi):
...
+    dist = np.arctan2(num ** .5, den)
+    return dist
+
+
+def _haversine_matrix(lon, lat, lon_ref, lat_ref):
+    """Compute a distance matrix using haversine formula.
+
+    The result must be multiplied by Earth radius to obtain distance in m or km.
+    The returned distance matrix has shape (n x n_ref).
+    """
```
djhoese, line 84: Units for inputs and outputs here too would be nice.

pyresample/spherical.py
```diff
@@ -34,22 +34,121 @@ def _unwrap_radians(val, mod=np.pi):
...
+    lon = lon[:, np.newaxis]
+    lat = lat[:, np.newaxis]
+    diff_lon = lon - lon_ref  # n x n_ref matrix
+    diff_lat = lat - lat_ref  # n x n_ref matrix
+    a = np.sin(diff_lat / 2.0) ** 2.0 + np.cos(lat) * np.cos(lat_ref) * np.sin(diff_lon / 2.0) ** 2.0
+    dist = 2.0 * np.arcsin(a ** .5)  # equivalent of; 2.0 * np.arctan2(np.sqrt(a), np.sqrt(1-a))
+    return dist
+
+
+def check_lon_validity(lon):
```
djhoese, line 98: Can these 3 check functions be private (`_` prefix)?

pyresample/spherical.py
```diff
@@ -34,22 +34,121 @@ def _unwrap_radians(val, mod=np.pi):
...
+        lon, lat = check_lon_lat(lon, lat)
+        self.lon = _unwrap_radians(lon)
         self.lat = lat
 
+    @property
+    def vertices(self):
+        """Return point(s) vertices in a ndarray of shape [n,2]."""
+        # Single values
+        if self.lon.ndim == 0:
+            vertices = np.array([self.lon, self.lat])[None, :]
```
djhoese, line 138:
```suggestion
            vertices = np.array([self.lon, self.lat])[np.newaxis, :]
```

pyresample/spherical.py
```diff
@@ -62,35 +161,48 @@ def cross2cart(self, point):
...
         h = np.cos(lat2)
         i = np.sin(lon2 - lon1)
-        res = CCoordinate(np.array([-ad * c + be * f,
-                                    ad * f + be * c,
-                                    g * h * i]))
-
+        x = -ad * c + be * f
+        y = ad * f + be * c
+        z = g * h * i
+        vertices = _xyz_to_vertices(x, y, z)
```
djhoese, line 167: Is `_xyz_to_vertices` doing the same number of copies and producing the same result as the `np.array` call was in the old version of the code?

pyresample/spherical.py
```diff
@@ -112,6 +224,39 @@ def __iter__(self):
...
+    def plot(self, ax=None, **plot_kwargs):
+        """Plot the point(s) using Cartopy.
+
+        Assume vertices to be in radians.
+        """
+        import matplotlib.pyplot as plt
+        try:
+            import cartopy.crs as ccrs
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError("Install cartopy to plot spherical geometries.")
```
djhoese, line 236: We've had users complain about not being told explicitly how to install something.

```suggestion
            raise ModuleNotFoundError("Install cartopy to plot spherical geometries. For example, 'pip install cartopy'.")
```

pyresample/spherical.py
```diff
@@ -112,6 +224,39 @@ def __iter__(self):
...
+
+        # Create figure if ax not provided
+        ax_not_provided = False
+        if ax is None:
+            ax_not_provided = True
+            proj_crs = ccrs.PlateCarree()
```
djhoese, line 242: Not saying this needs to be done now, but should this allow users to provide the CRS? Devil's advocate: What if I don't want cartopy to plot coastlines. I just want to see the spherical geometries plotted together. Why can't I use a matplotlib 3D plot? :wink:

pyresample/test/test_sgeom/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# later version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Test for spherical geometries."""
```
djhoese, line 18: Should this module maybe be called `test_future_spherical.py`? Or maybe rename the old one to `test_legacy_spherical.py` and name this `test_spherical.py`?

djhoese (changes requested):
This is looking really good. I haven't kept up on this development so I had a lot of questions and a couple requested changes, but otherwise I like how you cleaned up the code.

One other suggestion that may help tests in the future, should some of these SCoordinate objects you use in the tests be pytest fixtures or other helper methods so the code to generate them doesn't get repeated all over the tests? I see some of them are unique, but also some use the same angles as inputs.

Again, nice job, just a few more things.

pyresample/future/spherical/point.py
ghiggi:
```suggestion
    """Define a point on the sphere.
```

pyresample/future/spherical/point.py
ghiggi:
```suggestion
    """Object representing multiple points on a sphere."""
```

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,72 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Define single and multiple points on the sphere through SPoint and SMultiPoint classes."""
+import numpy as np
+
+from pyresample.spherical import SCoordinate
+
+
+class SPoint(SCoordinate):
+    """Define a point on the sphere.
```
ghiggi, line 25:
```suggestion
    """Object representing a single point on a sphere.
```

pyresample/spherical.py
ghiggi:
```suggestion
    If x, y, z are scalar arrays, it creates a 1x3 np.array.
```

pyresample/test/test_sgeom/__init__.py
ghiggi: I agree that the directory should be defined as `test_spherical` .
I have 5 ongoing PR that use now `test_sgeom` directory.
If for you is ok, I will rename it after the merging of these PRs ...

pyresample/future/spherical/point.py
ghiggi: I added the property vertices_in_degrees

ghiggi: I also had the same problems with the current str and repr methods.
I changed to radians ;)

pyresample/spherical.py
ghiggi: True. `np.asarray` is a better option.

pyresample/spherical.py
ghiggi: This has been addressed

pyresample/spherical.py
ghiggi: In the old version of the code, all the elements were only scalar values.
In `_xyz_to_vertices` we do `vertices = np.array([x, y, z])` so nothing has changed. 
   
With this PR, now we allow to work also with lon/lat arrays, so `_xyz_to_vertices` does `np.vstack([x, y, z]).T` and this creates a copy I guess, but I don't see how to avoid that.

pyresample/spherical.py
ghiggi: I agree with that and we also discussed it with Martin about that. 
If ok for you, we delegate the testing and final design of the plot functions after the new spherical object classes are implemented. 
For the moment, I added the function arguments: `proj_crs=None, add_coastline=True, add_gridlines = True`

pyresample/spherical.py
ghiggi: Done

pyresample/spherical.py
ghiggi: Done

pyresample/spherical.py
ghiggi: Done

pyresample/future/spherical/point.py
```diff
@@ -57,16 +65,14 @@ def __eq__(self, other):
 
     def __str__(self):
         """Get simplified representation of lon/lat arrays in degrees."""
-        vertices = np.rad2deg(self.vertices)
-        return str(vertices)
+        return str(self.vertices)
 
     def __repr__(self):
         """Get simplified representation of lon/lat arrays in degrees."""
```
ghiggi, line 71:
```suggestion
        """Get simplified representation of lon/lat arrays in radians."""
```

pyresample/future/spherical/point.py
```diff
@@ -57,16 +65,14 @@ def __eq__(self, other):
 
     def __str__(self):
         """Get simplified representation of lon/lat arrays in degrees."""
```
ghiggi, line 67:
```suggestion
        """Get simplified representation of lon/lat arrays in radians."""
```

ghiggi:
@mraspaud @djhoese ready for merging I guess ;)

pyresample/test/test_sgeom/__init__.py
djhoese: Sounds good to me. Thanks.

pyresample/future/spherical/point.py
djhoese: @mraspaud any issue with the str/reprs being in radians (the same units as the internal data structures) versus degrees like it was in the past?

pyresample/spherical.py
djhoese: Sounds good to me.

pyresample/spherical.py
djhoese: So I think `np.array` will do a copy no matter what you do because it is converting from a list to a numpy array. I think the same is true for `np.vstack`, but I suppose if `.T` isn't a copy (it shouldn't be...I think) then maybe this is all fine.

pyresample/future/spherical/point.py
```diff
@@ -0,0 +1,88 @@
...
+        """Check equality."""
+        return np.allclose(self.lon, other.lon) and np.allclose(self.lat, other.lat)
+
+    def __str__(self):
+        """Get simplified representation of lon/lat arrays in radians."""
+        return str(self.vertices)
+
+    def __repr__(self):
+        """Get simplified representation of lon/lat arrays in radians."""
+        return str(self.vertices)
```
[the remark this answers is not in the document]
ghiggi: @djhoese I will change that in the next SArc PR. You mean to have a repr like: `'SMultiPoint with xxx vertices'`.
Or what would be your preferred repr?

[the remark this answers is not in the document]
djhoese: :thinking: I'm honestly not sure. Maybe something as simple as `f"{self.__class__.__name__}: {repr(self.vertices)}"`?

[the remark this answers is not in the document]
ghiggi: Addressed in https://github.com/pytroll/pyresample/pull/478

PR #466, opened 2022-11-08
[Future Spherical Class] Add SExtent  

ghiggi:
This PR adds the SExtent class of the future spherical geometry interface.

 - [x] Tests added 
 - [x] Tests passed 

### DIFF
--- a/pyresample/future/spherical/extent.py
+++ b/pyresample/future/spherical/extent.py
@@ -1,0 +1,242 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2013-2022 Pyresample Developers
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Definition of SExtent class."""
+import numbers
+
+import numpy as np
+from shapely.geometry import MultiPolygon, Polygon
+
+
+def bounds_from_extent(extent):
+    """Get shapely bounds from a matplotlib/cartopy extent.
+
+    Shapely bounds: [x_min, y_min, x_max, y_max]
+    Matplotlib/Cartopy extent: [x_min, x_max, y_min, y_max]
+    """
+    bounds = [extent[0], extent[2], extent[1], extent[3]]
+    return bounds
+
+
+def _get_list_extents_from_args(args):
+    """Get list_extent from SExtent input arguments."""
+    args = list(args)
+    # Check something is passed to SExtent
+    if len(args) == 0:
+        raise ValueError("No extent passed to SExtent.")
+    # Check that the first args element is a list, tuple of np.array
+    if np.any([not isinstance(arg, (list, tuple, np.ndarray)) for arg in args]):
+        raise TypeError("Specify the extent(s) using list, tuple or np.array.")
+    # Get list extent
+    # - If SExtent(extent) or SExtent(list_extent)
+    if len(args) == 1:
+        list_extents = list(*args)
+        if len(list_extents) == 0:  # input case: [], or ()
+            raise ValueError("No extent passed to SExtent.")
+    else:
+        list_extents = args
+    # If a single list, check correct length and contain a number
+    if len(list_extents) == 4 and isinstance(list_extents[0], numbers.Number):
+        list_extents = [list_extents]
+    # Ensure that here after is a list of extents
+    if not isinstance(list_extents[0], (tuple, list)):
+        raise ValueError("SExtent expects a single extent or a list of extents.")
+    return list_extents
+
+
+def _check_extent_order(extent):
+    """Check extent order validity."""
+    if extent[0] > extent[1]:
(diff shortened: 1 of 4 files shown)


ghiggi:
@djhoese @mraspaud this is ready for review.
Tomorrow I have faculty meetings, but I can finalize the last details on Friday if you manage to review it tomorrow.

ghiggi:
Prioritize the review of https://github.com/pytroll/pyresample/pull/465, https://github.com/pytroll/pyresample/pull/473 and https://github.com/pytroll/pyresample/pull/475 first ;) 
This will only be needed to speed up the SPolygon computations

pyresample/future/spherical/extent.py
```diff
@@ -0,0 +1,300 @@
...
+    if not isinstance(unioned_polygon, Polygon):
+        return False
+    # Define global polygon
+    global_polygon = Polygon.from_bounds(-180, -90, 180, 90)
+    # Assess if global
+    is_global = global_polygon.equals(unioned_polygon)
+    return is_global
+
+
+class SExtent(object):
```
djhoese, line 143: I'm not sure if the other maintainers disagree, but I'm not a fan of the object subclass anymore.

```suggestion
class SExtent:
```

pyresample/future/spherical/extent.py
```diff
@@ -0,0 +1,300 @@
...
+    More specifically, extents composing an SExtent:
+    - must represent a polygon (not a point or a line),
+    - can touch each other,
+    - but can not intersect/contain/overlap each other.
+
+    Examples of valid SExtent definitions:
+      extent = [x_min, x_max, y_min, ymax]
+      sext = SExtent(extent)
+      sext = SExtent([extent, extent])
+      sext = SExtent(extent, extent, ...)
```
djhoese, line 170: Is there a need or benefit to supporting both of these types of multi-extent inputs?

pyresample/future/spherical/extent.py
```diff
@@ -0,0 +1,300 @@
...
+
+    """
+
+    def __init__(self, *args):
+        # Get list of extents
+        list_extents = _get_list_extents_from_args(args)
+        # Check extents format and value validity
+        self.list = [_check_valid_extent(ext) for ext in list_extents]
```
djhoese, line 178: Please use a name that isn't a keyword in the Python language/standard library. And does it need to be public (can it be prefixed with `_`)?

pyresample/future/spherical/extent.py
```diff
@@ -0,0 +1,300 @@
...
+        # Pre-compute shapely polygon
+        list_polygons = [Polygon.from_bounds(*bounds_from_extent(ext)) for ext in self.list]
+        self.polygons = MultiPolygon(list_polygons)
```
djhoese, line 181: Can this be private too? `self._polygons`?

pyresample/test/test_sgeom/test_extent.py
```diff
@@ -0,0 +1,576 @@
...
+"""Test cases for SExtent class."""
+import unittest
+
+import numpy as np
+import pytest
+
+from pyresample.future.spherical import SExtent
+
+
+class TestSExtent(unittest.TestCase):
```
djhoese, line 27:
```suggestion
class TestSExtent:
```

pyresample/test/test_sgeom/test_shapely.py
```diff
@@ -0,0 +1,81 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Test cases for shapely."""
+import unittest
+
+import numpy as np
+from shapely.geometry import MultiPolygon, Polygon
+from shapely.ops import unary_union
+
+
+class TestShapely(unittest.TestCase):
```
djhoese, line 26:
```suggestion
class TestShapely:
```

pyresample/test/test_sgeom/test_shapely.py
```diff
@@ -0,0 +1,81 @@
...
+        vertices = np.array(unified_polygon.exterior.coords)
+        expected_vertices = np.array([[-180., 90.],
+                                      [0., 90.],
+                                      [180., 90.],
+                                      [180., 0.],
+                                      [0., 0.],
+                                      [0., -90.],
+                                      [-180., -90.],
+                                      [-180., 90.]])
+        assert np.allclose(expected_vertices, vertices)
```
djhoese, line 47: This should give better error messages when things fail:

```suggestion
        np.testing.assert_allclose(expected_vertices, vertices)
```

pyresample/future/spherical/extent.py
```diff
@@ -0,0 +1,300 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Definition of SExtent class."""
+import numbers
+
+import numpy as np
+from shapely.geometry import MultiPolygon, Polygon
+from shapely.ops import unary_union
+
+
+def bounds_from_extent(extent):
```
djhoese, line 26: Public or private interface?

djhoese (changes requested):
So basically this is to handle all the edge cases of extents that wouldn't make sense to include in a generic Polygon class, right?

pyresample/future/spherical/extent.py
ghiggi: No need. Which one would you suggest to drop? 
I would keep the first.

djhoese: Sorry, it's been a long time since I've looked at this and I can't quite remember what this class does. So it merges extents (bounding box?) into a single MultiPolygon? The idea being you either have one contiguous extent or you have two extents that "cross" the antimeridian, right? Will there ever be a need for more than two extents?

ghiggi: Yes. One could discretize a SPolygon in lot of non-intersecting extents ... 
But the main use case is to deal with the antimeridian ...

djhoese: I'm not sure what I prefer. Initially I was going to say only use the second one (passing a list), but since the usual case is a single 4-element extent then maybe the first and the third should be the only ways to do it.

PR #473, opened 2022-11-14
Add boundary method to AreaDefinition and SwathDefinition

ghiggi:
This PR adds the method `boundary` to AreaDefinition and SwathDefinition.
It enables the correct retrieval of an `AreaBoundary` object, which in future will be used to interface with the new spherical interface. 

 - [x] Tests added 
 - [x] Tests passed 


pyresample/test/test_geometry.py
```diff
@@ -3023,6 +3023,238 @@ def _is_clockwise(lons, lats):
...
+        boundary = areadef.boundary(frequency=n_vertices, force_clockwise=False)
+        boundary.vertices.shape
+
+        # Check boundary vertices is in correct order
+        expected_vertices = np.array([[-7.92337283e+01, 6.94302533e-15],
+                                      [-7.32756984e+01, 4.40228215e+01],
+                                      [1.81006860e-14, 7.91481122e+01],
+                                      [7.32756984e+01, 4.40228215e+01],
+                                      [7.92337283e+01, -0.00000000e+00],
+                                      [7.92337283e+01, -6.94302533e-15],  # duplicate ?
```
ghiggi, line 3129: This test must be updated when https://github.com/pytroll/pyresample/pull/475 is merged.

pyresample/geometry.py
```diff
@@ -391,6 +391,28 @@ def get_edge_bbox_in_projection_coordinates(self, frequency: Optional[int] = Non
         x, y = self._get_bbox_elements(self.get_proj_coords, frequency)
         return np.hstack(x), np.hstack(y)
 
+    def boundary(self, frequency=None, force_clockwise=False):
```
djhoese, line 394: Thoughts on `to_boundary`?

pyresample/geometry.py
```diff
@@ -391,6 +391,28 @@ def get_edge_bbox_in_projection_coordinates(self, frequency: Optional[int] = Non
...
+        """Retrieve the AreaBoundary object.
+
+        Parameters
+        ----------
+        frequency:
+            The number of points to provide for each side. By default (None)
+            the full width and height will be provided.
```
djhoese, line 401: Technically, at least in the old method, this was an integer equivalent to a "stride" in slicing. At least I think it was. So you would say "I want a boundary made up of every N pixels along the edge".

pyresample/geometry.py
ghiggi: Yes. Here is how I planned to design the access to the new spherical interface: 
```
swath_def.boundary().polygon()    # SPolygon
swath_def.boundary().line()           # SLine
swath_def.boundary().polygon().to_shapely()  # shapely Polygon
swath_def.boundary().line().to_shapely()         # shapely Line
```
Would you prefer the following pattern?
```
swath_def.to_boundary().to_polygon().to_shapely()
swath_def.to_boundary().to_line().to_shapely()
```

Personally I prefer the first option .. it's more "readable"

pyresample/geometry.py
ghiggi: Yes this was the old behavior. But in April-May, I noted to @mraspaud that such striding behavior was causing to often drop out large portions of swath/area between the last point selected and the actual corner of the swath/area. 
In the context of the gradient search PR, the behavior has been changed. See `_get_bbox_slices` for more details: https://github.com/pytroll/pyresample/blob/1268d904c9bcd01342173dcee73809da59cc79c1/pyresample/geometry.py#L332

What I discussed with Martin in these days, in one of the upcoming PRs, is to allow specifying `frequency` also as a tuple (n_x, n_y) to customize the number of points to select on the x and y sides.
This would help for example to speed up spherical operations for swath definitions where for the y-axis (cross-track scan direction), much fewer points (i.e only 2) are needed to get the accurate swath boundary.

I agree that the term "frequency" is misleading, but it is used in a large portion of the code and changing it I think would cause lot of backward incompatibilities.

pyresample/geometry.py
djhoese: Sure. There seems like a law of Demeter violation, but maybe I don't fully understand that concept either.

Leave it as is.

ghiggi:
@djhoese is there something that I need to do prior merging?

djhoese:
Waiting for merge of #475, then remerge of main into this branch, then tests to get updated and pass.

djhoese:
Ok #475 is merged, can you merge main here and then update the test that needs it?

ghiggi:
Ready to be merged

PR #475, opened 2022-11-15
Fix for duplicate coordinates in bbox_lonlats for geostationary area 

### LINKED ISSUE #474: get_geostationary_bounding_box* contains duplicated vertices at the equator
The `get_geostationary_bounding_box` returns duplicated vertices at the equator. 
The code snippet below shows the "bad" behaviour. 

```python
from pyresample.geometry import (
    get_geostationary_bounding_box_in_proj_coords,
    get_geostationary_bounding_box_in_lonlats
)
# Define geostationary projection
proj_dict_geostationary = {
        'proj_id': "dummy_geo_projection",
        "area_id" : 'dummy_geo_projection',
        "description" : 'geostationary projection',
        "projection" : {'a': 6378169.00,  'b': 6356583.80, 'h': 35785831.00, 'lon_0': 0, 'proj': 'geos'},
        "area_extent" : (-5500000., -5500000., 5500000., 5500000.),
        "width" : 100,
        "height" : 100,
}
areadef = AreaDefinition(**proj_dict_geostationary)
geos_area = areadef

#### If nb_points is 4, it requires only 2 points at the equator
nb_points = 4
x, y = get_geostationary_bounding_box_in_proj_coords(geos_area=geos_area,
                                                     nb_points=nb_points) 
print(x)
print(y) # all 0 

lons, lats = get_geostationary_bounding_box_in_lonlats(geos_area=geos_area,
                                                       nb_points=nb_points)
print(lons)
print(lats) # all 0 

#### get_geostationary_bounding_box_in_proj_coords  returns duplicated values at the equator 
lons, lats = get_geostationary_bounding_box_in_lonlats(geos_area=geos_area,
                                                       nb_points=6)
print(lons)
print(lats)  
lons[2:4] # [79.23372832, 79.23372832]
lats[2:4]  # [-0.00000000e+00, 0.00000000e+00]
```
#### Expected Output

The `get_geostationary_bounding_box` functions in case `nbpoints=4` should return the top, left, bottom and right "extent" of the geo disk, instead of the current left and right duplicates.

ghiggi:
This PR removes the presence of duplicate consecutive coordinates in the geostationary are bbox/polygon.
This currently leads to the creation of SArcs that are actually points, and also leads to cross-product operations that lead to cartesian [0,0,0] and thus vector norm = 0 and thus division by 0 when normalizing. 

Furthermore, it closes #474: when we retrieve 4 vertices we now get the 4 correct geos_area "extent" coordinates.

 - [x] Closes #474  
 - [x] Tests added 
 - [x] Tests passed 

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -2712,8 +2712,7 @@ def get_geostationary_bounding_box_in_proj_coords(geos_area, nb_points=50):
 
     x *= h
     y *= h
-    # Here below wer remove one element with [:-1] to avoid duplicated values
-    #  at the equator
+    # We remove one element with [:-1] to avoid duplicate values at the equator
     x = np.clip(np.concatenate([x[:-1], x[::-1]]), min(ll_x, ur_x), max(ll_x, ur_x))
     y = np.clip(np.concatenate([y[:-1], -y]), min(ll_y, ur_y), max(ll_y, ur_y))
 
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -2691,24 +2691,32 @@ def get_geostationary_angle_extent(geos_area):
 def get_geostationary_bounding_box_in_proj_coords(geos_area, nb_points=50):
     """Get the bbox in geos projection coordinates of the valid pixels inside `geos_area`.
 
-    Args:
-      nb_points: Number of points on the polygon
+    Notes:
+    - The first and last element of the output vectors are equals.
+      This is required to enable direct creation of area sides.
+    - If nb_points is even, it will returns x and y vectors of length nb_points + 1.
+
+    Parameters
+    ----------
+    nb_points : Number of points on the polygon
+
     """
     xmax, ymax = get_geostationary_angle_extent(geos_area)
     h = get_geostationary_height(geos_area.crs)
 
     # generate points around the north hemisphere in satellite projection
     # make it a bit smaller so that we stay inside the valid area
-    x = np.cos(np.linspace(-np.pi, 0, int(nb_points / 2.0))) * (xmax - 0.0001)
-    y = -np.sin(np.linspace(-np.pi, 0, int(nb_points / 2.0))) * (ymax - 0.0001)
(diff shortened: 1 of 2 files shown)


pyresample/geometry.py
```diff
@@ -2691,24 +2691,31 @@ def get_geostationary_angle_extent(geos_area):
...
 
     ll_x, ll_y, ur_x, ur_y = geos_area.area_extent
 
     x *= h
     y *= h
-
-    x = np.clip(np.concatenate([x, x[::-1]]), min(ll_x, ur_x), max(ll_x, ur_x))
-    y = np.clip(np.concatenate([y, -y]), min(ll_y, ur_y), max(ll_y, ur_y))
+    # Here below wer remove one element with [:-1] to avoid duplicated values
+    #  at the equator
```
djhoese, line 2716:
```suggestion
    # We remove one element with [:-1] to avoid duplicate values at the equator
```

pyresample/geometry.py
```diff
@@ -2691,24 +2691,30 @@ def get_geostationary_angle_extent(geos_area):
 def get_geostationary_bounding_box_in_proj_coords(geos_area, nb_points=50):
     """Get the bbox in geos projection coordinates of the valid pixels inside `geos_area`.
 
-    Args:
-      nb_points: Number of points on the polygon
+    Notes:
+    - The first and last element of the output vectors are equals.
+    - If nb_points is even, it will returns x and y vectors of length nb_points + 1.
```
djhoese, line 2696:
```suggestion
    - The first and last element of the output vectors are equal.
    - If nb_points is even, it will return x and y vectors of length nb_points + 1.
```

djhoese (approved):
I think this makes

djhoese:
Some of the uncovered lines that coveralls is complaining about are a little concerning, but I assume you'll be working with this in the upcoming PRs so I'm not too worried. Most of the missed lines are the matplotlib plotting code which is annoying to test anyway. The one concerning one is the vertices to xyz helper function. The branch with the `.T` is reported as uncovered. Let's merge this for now and have you work on covering that line in one of the other PRs.

PR #478, opened 2022-11-17
Add legacy Arc code and tests

ghiggi:
This PR implements the new SArc class 

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/pyresample/future/spherical/arc.py
+++ b/pyresample/future/spherical/arc.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2022 Pyresample developers
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Define a single great-circle arc on the sphere through the SArc class."""
+import numpy as np
+
+from pyresample.future.spherical import EPSILON
+from pyresample.spherical import SCoordinate, _unwrap_radians
+
+
+class Arc(object):
+    """An arc of the great circle between two points."""
+
+    def __init__(self, start, end):
+        self.start, self.end = start, end
+
+    def __eq__(self, other):
+        """Check equality."""
+        if self.start == other.start and self.end == other.end:
+            return 1
+        return 0
+
+    def __ne__(self, other):
+        """Check not equal comparison."""
+        return not self.__eq__(other)
+
+    def __str__(self):
+        """Get simplified representation."""
+        return str(self.start) + " -> " + str(self.end)
+
+    def __repr__(self):
+        """Get simplified representation."""
+        return str(self.start) + " -> " + str(self.end)
+
+    def angle(self, other_arc):
+        """Oriented angle between two arcs.
+
+        Returns:
+            Angle in radians. A straight line will be 0. A clockwise path
+            will be a negative angle and counter-clockwise will be positive.
+
+        """
+        if self.start == other_arc.start:
+            a__ = self.start
+            b__ = self.end
+            c__ = other_arc.end
+        elif self.start == other_arc.end:
+            a__ = self.start
+            b__ = self.end
+            c__ = other_arc.start
+        elif self.end == other_arc.end:
+            a__ = self.end
+            b__ = self.start
+            c__ = other_arc.start
+        elif self.end == other_arc.start:
+            a__ = self.end
+            b__ = self.start
+            c__ = other_arc.end
+        else:
+            raise ValueError("No common point in angle computation.")
+
+        ua_ = a__.cross2cart(b__)
+        ub_ = a__.cross2cart(c__)
+
+        val = ua_.dot(ub_) / (ua_.norm() * ub_.norm())
+
+        if abs(val - 1) < EPSILON:
+            angle = 0
+        elif abs(val + 1) < EPSILON:
+            angle = np.pi
+        else:
+            angle = np.arccos(val)
+
+        n__ = ua_.normalize()
+        if n__.dot(c__.to_cart()) > 0:
+            return -angle
+        else:
+            return angle
+
+    def intersections(self, other_arc):
+        """Give the two intersections of the greats circles defined by the current arc and *other_arc*.
+
+        From http://williams.best.vwh.net/intersect.htm
+        """
+        end_lon = self.end.lon
+        other_end_lon = other_arc.end.lon
+
+        if self.end.lon - self.start.lon > np.pi:
+            end_lon -= 2 * np.pi
+        if other_arc.end.lon - other_arc.start.lon > np.pi:
+            other_end_lon -= 2 * np.pi
+        if self.end.lon - self.start.lon < -np.pi:
+            end_lon += 2 * np.pi
+        if other_arc.end.lon - other_arc.start.lon < -np.pi:
+            other_end_lon += 2 * np.pi
+
+        end_point = SCoordinate(end_lon, self.end.lat)
+        other_end_point = SCoordinate(other_end_lon, other_arc.end.lat)
+
+        ea_ = self.start.cross2cart(end_point).normalize()
+        eb_ = other_arc.start.cross2cart(other_end_point).normalize()
+
+        cross = ea_.cross(eb_)
+        lat = np.arctan2(cross.cart[2],
+                         np.sqrt(cross.cart[0] ** 2 + cross.cart[1] ** 2))
+        lon = np.arctan2(cross.cart[1], cross.cart[0])
+
+        return (SCoordinate(lon, lat),
+                SCoordinate(_unwrap_radians(lon + np.pi), -lat))
+
+    def intersects(self, other_arc):
+        """Check if the current arc and the *other_arc* intersect.
+
+        An arc is defined as the shortest tracks between two points.
+        """
+        return bool(self.intersection(other_arc))
+
+    def intersection(self, other_arc):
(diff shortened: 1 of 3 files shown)


pyresample/future/spherical/arc.py
```diff
@@ -0,0 +1,209 @@
...
+            ab_dist = a.hdistance(b)
+            cd_dist = c.hdistance(d)
+            ap_dist = a.hdistance(spoint)
+            bp_dist = b.hdistance(spoint)
+            cp_dist = c.hdistance(spoint)
+            dp_dist = d.hdistance(spoint)
+
+            if (((spoint in (a, b)) or (abs(ap_dist + bp_dist - ab_dist) < EPSILON)) and
+                    ((spoint in (c, d)) or (abs(cp_dist + dp_dist - cd_dist) < EPSILON))):
+                return spoint
```
djhoese, line 155: Could be clean up this code? Perhaps remove the `a/b/c/d` naming with longer more readable names? More importantly, can the `if` conditions be broken up into variables to describe what is being checked? For example:

```python
point_is_this_arc = spoint in (self.start, self.end)
point_is_other_arc = spoint in (other_arc.start, other_arc.end)
point_is_on_this_arc_segment = abs(...) < EPSILON  # (or better names for whatever this calculation is)
point_is_on_other_arc_segment = ...
point_is_on_this_arc = point_is_this_arc_end or point_is_on_this_arc_segment
point_is_on_other_arc_segment = point_is_other_arc_end or point_is_on_other_arc_segment
if point_is_on_this_arc and point_is_on_other_arc:
    return spoint
````

File: pyresample/future/spherical/point.py [hunk 2]
```diff
@@ -39,13 +39,32 @@ def from_degrees(cls, lon, lat):
         """Create SPoint from lon/lat coordinates in degrees."""
         return cls(np.deg2rad(lon), np.deg2rad(lat))
 
+    def is_pole(self):
+        """Test if the point is on a pole."""
+        if self.lat in [-np.pi / 2, np.pi / 2]:
+            return True
+        return False
```
djhoese, line 46:
```suggestion
    def is_pole(self):
        """Test if the point is on a pole."""
        return self.lat in (-np.pi / 2, np.pi / 2)
```

pyresample/future/spherical/point.py
```diff
@@ -39,13 +39,32 @@ def from_degrees(cls, lon, lat):
...
+
+    def is_on_equator(self):
+        """Test if the point is on the equator."""
+        if self.lat == 0:
+            return True
+        return False
```
djhoese, line 52:
```suggestion
    def is_on_equator(self):
        """Test if the point is on the equator."""
        return self.lat == 0
```

pyresample/future/spherical/point.py
```diff
@@ -39,13 +39,32 @@ def from_degrees(cls, lon, lat):
...
         """Get simplified representation of lon/lat arrays in radians."""
-        return str((float(self.lon), float(self.lat)))
+        coords = str((float(self.lon), float(self.lat)))
+        return f"{self.__class__.__name__}: {coords}"
+
+    def __eq__(self, other):
+        """Check equality."""
+        if self.lat == other.lat and self.is_pole():
+            return True
+        return np.allclose((self.lon, self.lat), (other.lon, other.lat))
```
djhoese, line 67: I'm tempted to want this to be like the below, but maybe how you have it is more clear:

```suggestion
        return (self.lat == other.lat and self.is_pole()) or np.allclose((self.lon, self.lat), (other.lon, other.lat))
```

pyresample/spherical.py
```diff
@@ -366,19 +366,19 @@ def __init__(self, start, end):
     def __eq__(self, other):
         """Check equality."""
         if self.start == other.start and self.end == other.end:
```
djhoese, line 368: Github is being annoying and not letting me select this full method

```suggestion
        return self.start == other.start and self.end == other.end
```

pyresample/spherical.py
```diff
@@ -366,19 +366,19 @@ def __init__(self, start, end):
...
-            return 1
-        return 0
+            return True
+        return False
```
djhoese, line 370:
```suggestion
```

pyresample/spherical.py
```diff
@@ -366,19 +366,19 @@ def __init__(self, start, end):
...
 
     def __ne__(self, other):
         """Check not equal comparison."""
         return not self.__eq__(other)
 
     def __str__(self):
-        """Get simplified representation."""
+        """Get simplified str representation."""
```
djhoese, line 377: I'm not sure these changes were needed, but if we're going to say string lets just say string (not str).

```suggestion
        """Get simplified string representation."""
```

pyresample/spherical.py
```diff
@@ -427,9 +427,9 @@ def angle(self, other_arc):
             return angle
 
     def intersections(self, other_arc):
-        """Give the two intersections of the greats circles defined by the current arc and *other_arc*.
+        """Compute the intersections points of the greats circles over which the arcs lies.
```
djhoese, line 430:
```suggestion
        """Compute the intersections points of the great circles over which the arcs lies.
```

pyresample/test/test_sgeom/test_arc.py
```diff
@@ -0,0 +1,312 @@
...
+import unittest
+
+import numpy as np
+import pytest
+from shapely.geometry import LineString
+
+from pyresample.future.spherical import SArc, SPoint
+
+
+class TestSArc(unittest.TestCase):
```
djhoese, line 30: Looks like there is no need for this class (a new one) to be based on unittest.

```suggestion
class TestSArc:
```

pyresample/test/test_sgeom/test_arc.py
```diff
@@ -0,0 +1,312 @@
...
+    """Test SArc class."""
+
+    # TODO: Fixtures defined here?
```
djhoese, line 33: You could define these as fixtures in the class, but considering this module is specific to testing the SArc class maybe it would be better to define them outside the class. But also...if we assume these classes/instances are immutable then you could also leave them like this.

pyresample/test/test_sgeom/test_arc.py
```diff
@@ -0,0 +1,312 @@
...
+        assert arc1.intersects(arc2)
+
+        # Test non-intersecting SArc(s)
+        arc1 = SArc(SPoint.from_degrees(-2.4982818108326734, 48.596644847869655),
+                    SPoint.from_degrees(-2.9571441235622835, 49.165688435261394))
+        arc2 = SArc(SPoint.from_degrees(-3.4976667413531688, 48.562704872921373),
+                    SPoint.from_degrees(-5.893976312685715, 48.445795283217116))
+        assert arc1.intersection_point(arc2) is None
+        assert not arc1.intersects(arc2)
+        assert not bool(None)  # this occurs within the intersects method
```
djhoese, line 175: These section comments (and the code) suggest these should all be separate tests. If they require more than an descriptive function name then a docstring can be added.

pyresample/test/test_sgeom/test_arc.py
```diff
@@ -0,0 +1,312 @@
...
+        p = equator_arc2.intersection_point(equator_arc1.reverse_direction())
+        assert np.allclose(p.vertices_in_degrees, (0, 0))
+
+        # Overlapping arcs
+        p = equator_arc2.intersection_point(equator_arc3)
+        assert isinstance(p, type(None))
+
+        # Disjoint arcs
+        p = equator_arc3.intersection_point(equator_arc1)
+        assert isinstance(p, type(None))
```
djhoese, line 252: Same here. These seem like separate tests. Perhaps a separate class or fixtures if they need to reuse the same input data?

djhoese (changes requested):
I have a couple code cleanup suggestions, but otherwise this looks good to me. I assume all the math is correct.

pyresample/test/test_sgeom/test_arc.py
ghiggi: I will clean out the unit test when I will finalize it after the SPolygon behaviour. 
I am not sure yet of all the behaviour ... and the current code helps me to work line by line

pyresample/test/test_sgeom/test_point.py
```diff
@@ -78,15 +90,42 @@ def test_raise_error_if_multi_point(self):
...
+        """Check the equality of a point on the antimeridian."""
+        p = SPoint.from_degrees(-180, 0)
+        p1 = SPoint.from_degrees(180, 0)
+        assert p == p1
+
     def test_str(self):
         """Check the string representation."""
         d = SPoint(1.0, 0.5)
-        self.assertEqual(str(d), "(1.0, 0.5)")
+        assert str(d) == "(1.0, 0.5)"
```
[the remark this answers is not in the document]
ghiggi: I dropped the unittest dependency

djhoese:
Looks like some of the github suggestions from my comments or your other recent changes have messed a few things up. Pre-commit and stickler and the website building are all mad.

ghiggi:
@djhoese I think now the code is ready for another round of review ;)

pyresample/future/spherical/arc.py
```diff
@@ -58,15 +60,11 @@ def __hash__(self):
 
     def is_on_equator(self):
```
djhoese, line 85: Should this be an `on_equator` property? Or `is_on_equator` I guess if we want to keep the verb.

pyresample/future/spherical/arc.py
```diff
@@ -34,9 +34,11 @@ def _check_valid_arc(start_point, end_point):
         raise ValueError("An SArc can not be represented by the same start and end SPoint.")
     if start_point.is_pole() and end_point.is_pole():
         raise ValueError("An SArc can not be uniquely defined between the two poles.")
-    if start_point.is_on_equator and end_point.is_on_equator and abs(start_point.lon - end_point.lon) == np.pi:
+    if start_point.is_on_equator() and end_point.is_on_equator() and abs(start_point.lon - end_point.lon) == np.pi:
         raise ValueError(
             "An SArc can not be uniquely defined on the equator if start and end points are 180 degrees apart.")
+    if start_point.get_antipode() == end_point:
```
djhoese, line 40: Similar to the `is_on_equator` should this just be a `antipode` property? Possibly cached if it is used a lot.

djhoese:
Looks like one of the tests is failing with the unstable dependencies. Not sure what's up with that.

PR #479, opened 2022-11-19
Fix bbox creation for SwathDefinitions with NaNs

djhoese:
I ran into this with some VIIRS data and some AVHRR data. In one case I had a couple rows/scans of data that had NaNs in them so some of the bbox coordinates were NaN. This resulted in bbox coordinates being returned, but being invalid (had NaNs). In another case the top couple scans of the swath were all NaNs so you'd get some hard to decipher error about NaN values. Sometimes these only made themselves obvious when you made a Boundary/Polygon and tried to do some calculation with them.

This PR fixes these cases or improves the error messages by removing any bbox points that have NaN for one of the coordinates.

CC @ghiggi 

 - [x] Tests added 
 - [x] Tests passed 


pnuu (approved):
LGTM

PR #487, opened 2022-12-08
FIX: Update pyproj min version to 3.0.0

### LINKED ISSUE #486: BUG: Import fails due to misconfigured setup.py

MRE:
From a fresh environment
- `conda install satpy`
- open a python terminal
- `import satpy`
- Get a error (see traceback below)

#### Problem description

[this should also explain **why** the current behaviour is a problem and why the 
expected output is a better solution.]

When installing satpy from a fresh environment, the version of pyproj that get installed is `'2.6.1.post1'`.
However, pyresample requires the `pyproj.aoi` module that was introduced in [pyproj 3.0.0](https://github.com/pyproj4/pyproj/blob/main/docs/history.rst#300).

This can _probably_ be fixed by updating _setup.py_ of **pyresample** project to `'pyproj>=3.0.0'` instead of `'pyproj>=2.2'`

Note: I'm new to pytroll's ecosystem.

#### Expected Output
No error.

```
import pyresample
  File "<stdin>", line 1, in <module>
  File "C:\Users\aouna\Anaconda3\lib\site-packages\pyresample\__init__.py", line 25, in <module>
    from pyresample import geometry  # noqa
  File "C:\Users\aouna\Anaconda3\lib\site-packages\pyresample\geometry.py", line 33, in <module>
    from pyproj.aoi import AreaOfUse
ModuleNotFoundError: No module named 'pyproj.aoi'
```

python 3.9.13
satpy: 0.38

bzah:
- [x] Closes #486  
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented 

This PR update the min version of pyproj to 3.0 as pyresample relies on it already.

The only test that I did was to manually install pyproj 3.0 and to make sure `import pyresample` does throw an error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ from setuptools import Extension, find_packages, setup
 
 import versioneer
 
-requirements = ['setuptools>=3.2', 'pyproj>=2.2', 'configobj',
+requirements = ['setuptools>=3.2', 'pyproj>=3.0', 'configobj',
                 'pykdtree>=1.3.1', 'pyyaml', 'numpy>=1.10.0',
                 ]
 


djhoese:
Thanks for the fix!

PR #499, opened 2023-02-10
Drop Python 3.8 support

djhoese:
It's what all the cool kids are doing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,7 +23,7 @@ jobs:
           miniforge-variant: Mambaforge
           miniforge-version: latest
           use-mamba: true
-          python-version: "3.9"
+          python-version: "3.11"
           environment-file: continuous_integration/environment.yaml
           activate-environment: test-environment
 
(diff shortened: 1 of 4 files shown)


djhoese:
Note I also removed some old multiprocessing import code in setup.py that said it could be removed when Python 2 support was dropped.

pnuu (approved):
LGTM

pnuu:
Gnaah. Don't know what I did there. But LGTM.

One thought I had was that maybe the `python_requires` shouldn't be so strict, most of the code will work fine with earlier versions. Though even in that case we should state that only the tested Python versions are actually supported.

djhoese:
I'm noticing codecov is complaining of lost coverage in the multiprocessing parts (`_spatial_mp.py`) for modules that haven't been changed, even in #498 which doesn't have these python version changes. Makes me wonder if something has changed in some package (numpy? pyproj?) that is no longer accessing the `__iter__` or other parts of these modules.

djhoese:
> Though even in that case we should state that only the tested Python versions are actually supported.

The problem is that `python_requires` can control what versions of the library PyPI returns. So if we make a release that supports Python 3.7 and then make a bug fix release that introduces syntax that is 3.8+ only then we'd have to remember to drop support in a bug fix version (or release a new minor release just for a bug fix). My point is I think CI needs to stay in sync with what we advertise as supported and it might be better to do it sooner rather than later. :man_shrugging:

pnuu:
Ah, right. That'd get complicated and confusing quickly. So yeah, lets stay in sync with the CI.

PR #521, opened 2023-06-01
Fix: data_reduce with Gaussian reduced grids

AlexandreMary:
Resampling from Gaussian reduced grids (seen as a 2D masked array with missing values at the end of each row) was failing, trying to reduce data.
This correction (using np.round instead of round) escapes the reduction in this case, and the resampling then works fine (albeit slower than optimal maybe).
Open to any better option.

 - [x] Tests passed (not worse than main branch on my distrib)

### DIFF
--- a/pyresample/data_reduce.py
+++ b/pyresample/data_reduce.py
@@ -279,13 +279,13 @@ def _get_valid_index(lons_side1, lons_side2, lons_side3, lons_side4,
     # 360: area covers south pole
     #   0: area covers no poles
     # else: area covers both poles
-    if round(angle_sum) == -360:
+    if np.round(angle_sum) == -360:
         # Covers NP
         valid_index = (lats >= lat_min_buffered)
-    elif round(angle_sum) == 360:
+    elif np.round(angle_sum) == 360:
         # Covers SP
         valid_index = (lats <= lat_max_buffered)
-    elif round(angle_sum) == 0:
+    elif np.round(angle_sum) == 0:
         # Covers no poles
         valid_lats = (lats >= lat_min_buffered) * (lats <= lat_max_buffered)
 
--- a/pyresample/data_reduce.py
+++ b/pyresample/data_reduce.py
@@ -225,6 +225,9 @@ def _get_valid_index(lons_side1, lons_side2, lons_side3, lons_side4,
                      lats_side1, lats_side2, lats_side3, lats_side4,
                      lons, lats, radius_of_influence):
     """Find relevant indices from grid boundaries using the winding number theorem."""
+    masked_grid = any([isinstance(side, np.ma.MaskedArray) and side.mask.any()
+                       for side in (lons_side1, lons_side2, lons_side3, lons_side4)])
+
     # Coarse reduction of data based on extrema analysis of the boundary
     # lon lat values of the target grid
     illegal_lons = (((lons_side1 < -180) | (lons_side1 > 180)).any() or
@@ -237,7 +240,7 @@ def _get_valid_index(lons_side1, lons_side2, lons_side3, lons_side4,
                     ((lats_side3 < -90) | (lats_side3 > 90)).any() or
                     ((lats_side4 < -90) | (lats_side4 > 90)).any())
 
-    if illegal_lons or illegal_lats:
+    if masked_grid or illegal_lons or illegal_lats:
         # Grid boundaries are not safe to operate on
         return np.ones(lons.size, dtype=bool)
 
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks for putting a fix together. Just to make sure I understand, it isn't that this broke in some way, it is that this never worked at all, right?

For your data, the problem is the edge longitudes are masked and/or NaNs, right? So the final `angle_sum` is probably masked or NaN? What is the end result you're hoping for? Should it do any data reduction? If so, how does that work with these masked values? I suppose the `invalid` lon/lat check near the top of the function needs to be updated to handle masked arrays in which case your data would get returned with no reduction, right?

And you say it was failing, were you getting an error? Could you paste it here so others will find this PR when googling the error message?

AlexandreMary:
Yes, I think it never worked.
Indeed the edge lons/lats are Masked, ending up in `angle_sum` being `MaskedConstant` and the resulting error:
`TypeError: type MaskedConstant doesn't define __round__ method`

From what I understand of this data reduction, indeed, it would require further development to implement an actual reduction in this case, which is probably not a priority. So indeed, I'm happy with no data reduction (performance-wise).

An update of the detection of `illegal` lons/lats could also be a way of working around the issue, indeed.

Thanks

AlexandreMary:
Unless you can suggest me another way of dealing with Reduced Gauss grids (such as IFS or Arpege NWP models')...

djhoese:
> Unless you can suggest me another way of dealing with Reduced Gauss grids (such as IFS or Arpege NWP models')...

I understand about half of what you said so probably not.

I think I'd prefer instead of this `np.round` change to change it to be more explicit and check for masked values or NaNs. You could/should maybe do a `.filled(np.nan)` call on the lon/lat side checks before `any()` is called. At least the possibility of a masked array needs to be handled.

AlexandreMary:
OK, I removed the `np.round` towards a detection of the `lons_side*` being actually masked arrays. More clean, indeed.

djhoese (commented):
This looks good I think. I just realized, I think if the invalid logic was inverted it would automatically handle the masked values. The conditions are checking for what is invalid, but for NaN and masked handling they should check for valid pixels. I'm ok leaving this as-is since no one has complained yet, but if you're up for it you could probably get the code cleaner that way (> instead of <=). Otherwise, do you think you could add a unit test for this?

AlexandreMary:
OK, I tried to add a test (in `test_data_reduce`) but can't figure out how it works exactly, `valid_index` is 1D and bigger than the first dimension (nlats):

```
    def test_area_con_reducedgauss_reduce(self):
        nlats = 2000
        max_nlons = 2 * nlats
        lats_resolution = 180 / (nlats + 1)
        lons_resolution = 360 / max_nlons
        lons = np.ma.masked_all((nlats, max_nlons))
        lats = np.ma.masked_all((nlats, max_nlons))
        for i in range(nlats):
            nlons = max(8, 2 * i)
            lats[i, :] = -90 + (i + 1) * lats_resolution
            for j in range(nlons):
                lons[i, j] = j * lons_resolution
        reduced_gauss_grid_def = geometry.GridDefinition(lons, lats)
        data = np.fromfunction(lambda y, x: (y + x), (nlats, max_nlons))
        lons = np.fromfunction(
            lambda y, x: -180 + (360.0 / 100) * x, (100, 100))
        lats = np.fromfunction(
            lambda y, x: -90 + (180.0 / 100) * y, (100, 100))
        grid_lons, grid_lats = reduced_gauss_grid_def.get_boundary_lonlats()
        valid_index = get_valid_index_from_lonlat_boundaries(grid_lons, grid_lats,
                                                             lons, lats, 7000)
        data = data[valid_index]
        cross_sum = data.sum()
        expected = 20685125.0
        self.assertAlmostEqual(cross_sum, expected)
```

But after all does it make sense to make a test on a case we know is not implemented for data reduction ?
More radically, we could also raise an error in case of masked grid, and so use `reduce_data=False` in the calling `kd_tree` functions (`get_neighbour_info`).

djhoese:
Sorry for the slow reply. I was on a vacation. What do you mean a case that is not implemented?

For the 1D issue and how it works...I'm honestly not sure as I haven't used this function before or at least not directly. It may be that it is a flattened version of the array.

AlexandreMary:
I meant that the data reduction in the case of a masked grid is not implemented (as we end up with masked constants or NaN in the `angle_sum`).
A cleaner fix would be to implement this data reduction for masked grids, but that is probably not a priority, for me nor for you.

So in the end, I can live without the fix, calling `get_neighbour_info(..., data_reduction=False)` and we can close this PR.

djhoese:
I'm tempted to say this is such a simple fix that we should include it. You somehow stumbled on it so it might come up again later. Having a test for it ensures (or at least it should) that the functionality is retained between refactors.

djhoese:
If you have no desire to continue the fix for this and will just workaround it then I'm OK closing it. I have no use for it myself so I can't pretend I'll volunteer to work on or maintain it. We can reopen this if you'd like.

AlexandreMary:
OK for me. Thanks

PR #532, opened 2023-07-17
Migrate usages of pyproj Proj to pyproj Transformer

### LINKED ISSUE #527: area definition for a rotated pole coordinate system
Hey all,
thanks a lot! I am exploring pyresample a bit for the use in resampling land cover data as input for our regional climate model. I found the [BucketResampler](https://pyresample.readthedocs.io/en/latest/api/pyresample.bucket.html#pyresample.bucket.BucketResampler) to be exactly what i need, especially the ability to compute fractions on a corser grids! That's a great tool! However, the regional modely mainly work on rotated grid transformations and i am struggling a little bit to get the transformation right.

So this would be my area definition, it is for Europe on a 0.44 degree resolution and rotated pole and an area extent defined in the projection coordinates:

```python
from pyresample.geometry import AreaDefinition

area_id = 'EUR-44'
description = 'Europe'
proj_id = 'rotated_pole'
projection = '+proj=ob_tran +o_proj=longlat +o_lon_p=0 +o_lat_p=39.25 +lon_0=18 +datum=WGS84 +no_defs +type=crs'
width = 106
height = 103
area_extent = (-28.21, -23.21, 17.99, 21.67)

area_def = AreaDefinition(area_id, description, proj_id, projection,
                          width, height, area_extent = area_extent)

```
#### Problem description

I think, i don't get the projection and area extent to work together. E.g. when i plot the latitude and longitude coordinates from this area definition, it doesn't look like it's getting the area extent right (e.g. ranges from -180, 180):

```python

plt.imshow(area_def.get_lonlats()[0]) # plot longitudes
plt.colorbar()
```
grafik

#### Expected Output

I usually would do the transformation myself, with something like this (and i guess pyresample does something simliar under the hood?)

```python
from pyproj import CRS, Transformer
import numpy as np

# resolution
dlon = (area_extent[2] - area_extent[0]) / (width-1) # 0.44
dlat = (area_extent[3] - area_extent[1]) / (height-1) # 0.44
# longitude, latitude in regular coordinates (rotated)
rlon = np.arange(0, width) * dlon + area_extent[0]
rlat = np.arange(0, height) * dlat + area_extent[1]

# transform into lon lat coordinates
transformer = Transformer.from_crs(CRS(projection), CRS("EPSG:4326"),  always_xy=True)
lon, lat = transformer.transform(np.meshgrid(rlon, rlat)[0], np.meshgrid(rlon, rlat)[1])
```
and to compare, the longitude plot looks like expected (e.g, i covers the expected range of longitudes for my European domain)
```python
plt.imshow(lon)
plt.colorbar()
```
grafik

I want to mention that all those thinks work fine when i define my domain in `ESPG:4326`, e.g., like this:

```python
area_id = 'EUR-44i'
description = 'Europe'
proj_id = 'regular'
projection = 'EPSG:4326'
width = 221
height = 103
area_extent = (-44.75, 21.75, 65.25, 72.75)

area_def = AreaDefinition(area_id, description, proj_id, projection,
                          width, height, area_extent = area_extent)
```

I am not sure, if what i expect is actually reasonable? Thanks a lot your help and apologies for the long issue!

```
pyproj=3.3.1
pyresample=1.27.0
```

djhoese:
This PR switches from using pyproj's old `Proj` interface to the newer `Transformer` class. Transformers should have better overall handling of more complex projection transformation.

One note is that I did have to add special handling of CRSes with prime meridian shifts (`+pm=180`) to make `get_lonlats` return what it used to. That is, lon/lats on a prime meridian 0 CRS.

CC @larsbuntemeyer

 - [x] Closes #527  
 - [ ] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/_spatial_mp.py
+++ b/pyresample/_spatial_mp.py
@@ -24,6 +24,7 @@ import multiprocessing as mp
 
 import numpy as np
 import pyproj
+from pyproj import CRS
 
 try:
     import numexpr as ne
@@ -31,6 +32,7 @@ except ImportError:
     ne = None
 
 from ._multi_proc import Scheduler, shmem_as_ndarray
+from .utils.proj4 import get_geodetic_crs_with_no_datum_shift
 
 # Earth radius
 R = 6370997.0
@@ -238,12 +240,15 @@ def _parallel_proj(scheduler, data1, data2, res1, res2, proj_args, proj_kwargs,
         _res2 = shmem_as_ndarray(res2)
 
         # Initialise pyproj
-        proj = pyproj.Proj(*proj_args, **proj_kwargs)
+        proj_def = proj_args[0] if proj_args else proj_kwargs
+        crs = CRS.from_user_input(proj_def)
+        gcrs = get_geodetic_crs_with_no_datum_shift(crs)
+        transformer = pyproj.Transformer.from_crs(gcrs, crs, always_xy=True)
+        trans_kwargs = {"radians": radians, "errcheck": errcheck, "direction": "INVERSE" if inverse else "FORWARD"}
 
         # Reproject data segment
         for s in scheduler:
-            _res1[s], _res2[s] = proj(_data1[s], _data2[s], inverse=inverse,
-                                      radians=radians, errcheck=errcheck)
+            _res1[s], _res2[s] = transformer.transform(_data1[s], _data2[s], **trans_kwargs)
 
     # An error occured, increment the return value ierr.
     # Access to ierr is serialized by multiprocessing.
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -18,7 +18,6 @@
 """Classes for geometry operations."""
 from __future__ import annotations
 
-import contextlib
 import hashlib
 import math
 import warnings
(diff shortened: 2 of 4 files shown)


larsbuntemeyer:
Thanks for fixing this @djhoese !

mraspaud (approved):
LGTM

pyresample/_spatial_mp.py
```diff
@@ -238,12 +240,15 @@ def _parallel_proj(scheduler, data1, data2, res1, res2, proj_args, proj_kwargs,
         _res2 = shmem_as_ndarray(res2)
 
         # Initialise pyproj
-        proj = pyproj.Proj(*proj_args, **proj_kwargs)
+        proj_def = proj_args[0] if proj_args else proj_kwargs
+        crs = CRS.from_user_input(proj_def)
+        gcrs = get_geodetic_crs_with_no_datum_shift(crs)
+        transformer = pyproj.Transformer.from_crs(gcrs, crs, always_xy=True)
+        trans_kwargs = {"radians": radians, "errcheck": errcheck, "direction": "INVERSE" if inverse else "FORWARD"}
```
snowman2, line 247: A little late, but I wanted to note that using the enumeration for transform direction provides a slight performance improvement.

djhoese: Thanks. Are we talking a line of code difference (to look up the value in the Enum) are like a completely different code path?

snowman2: If you provide the enumeration you skip the overhead of building the enum and the type checks in cython appear to be faster. Related https://github.com/pyproj4/pyproj/pull/1208

djhoese: Could that enum please be added to `pyproj/__init__.py` for easier access? All the other objects are convenience-imported there.

djhoese: Understood. Sounds good.

djhoese: Ah that PEP for lazy importing was rejected. Just noticed.

PR #545, opened 2023-10-10
Convert AreaDefinitions to odc geoboxes

BENR0:
Convert `AreaDefinitions` to odc geoboxes (https://odc-geo.readthedocs.io/en/latest/).

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented


pyresample/geometry.py
```diff
@@ -1972,6 +1972,19 @@ def _cartopy_proj_params(self):
...
+
+        See: https://odc-geo.readthedocs.io/en/latest/
+        """
+        try:
+            from odc.geo.geobox import GeoBox
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError("Please install 'odc-geo' to use this method.")
+
+        return GeoBox.from_bbox(bbox=self.area_extent, crs=self.crs,
+                                resolution=np.mean([self.pixel_size_x, self.pixel_size_y]), tight=True)
```
djhoese, line 1986: I think resolution can be a 2 element pair using the `Resolution` class:

https://odc-geo.readthedocs.io/en/latest/_api/odc.geo.Resolution.html#odc.geo.Resolution

setup.py
```diff
@@ -40,8 +40,16 @@
...
                   'xarray_bilinear': ['xarray', 'dask', 'zarr'],
+                  'odc-geo': ['odc-geo'],
                   'tests': test_requires}
 
+all_extras = []
+for extra_deps in extras_require.values():
+    all_extras.extend(extra_deps)
+extras_require['all'] = list(set(all_extras))
+
+setup_requires = ['numpy>=1.10.0', 'cython']
```
djhoese, line 51: This isn't actually used anywhere, plus this kind of thing (setup_requires) should be defined in the pyproject.toml build section.

pyresample/test/test_geometry/test_area.py
```diff
@@ -403,6 +403,35 @@ def test_cartopy_crs_latlon_bounds(self, create_test_area):
         latlong_crs = area_def.to_cartopy_crs()
         np.testing.assert_allclose(latlong_crs.bounds, [-180, 180, -90, 90])
 
+    def test_to_odc_geobox(self, stere_area, create_test_area):
```
djhoese, line 406: Can we xfail or skip if odc is missing? I think pytest has a decorator specifically for skipping if import fails.

djhoese (changes requested):
Love it! I had some requests, but otherwise I think this is an obvious addition. I'd be curious what @ghiggi's opinion is on this as I'm wondering if we could essentially drop our existing boundary stuff (eventually). That assumes I understand what odc-geo's boxes can do which I may not have a great understanding.

setup.py
BENR0: Good point. I have to admit I just copied it to get the different extra install options working.

pyresample/test/test_geometry/test_area.py
BENR0: That's a good Idea. I will also add a test for missing `odc-geo`since that is not tested yet.

BENR0:
Glad that you like it was not sure if it is a valuable addition. I just used it because I was to lazy to create Affine transformations from scratch and was thinking that maybe I can test some stuff for rerprojections and maybe for the html representation.

pyresample/geometry.py
BENR0: I know I can't really remember why I didn't use it but will check and change it eventually.

ghiggi:
Cool feature @BENR0. I never used the odc-geo tool. But from a quick look @djhoese: 
- does not support swath definition data
- I am not sure it currently support areas crossing the antimeridian (@BENR0 could you test the result with GOES-17 AreaDef?)
- The implemented geometry operations are implemented wrapping shapely ... so classical problems ...
 --> We could maybe use the odc-geo for the `get_bbox_lonlats` of AreaDefinition. But it's worth?

pyresample/test/test_geometry/test_area.py
djhoese: And thanks for using the `create_test_area` fixture. It makes my work on Pyresample 2.0 much easier.

djhoese:
> does not support swath definition data

Sure, but I think that's the point. They're projected bounding boxes.

> I am not sure it currently support areas crossing the antimeridian (@BENR0 could you test the result with GOES-17 AreaDef? area_def = satpy.resample.get_area_def("goes_west_abi_f_1km")
The implemented geometry operations are implemented wrapping shapely ... so classical problems ...
--> We could maybe use the odc-geo for the get_bbox_lonlats of AreaDefinition. But it's worth?

I would suspect they don't currently support the anti-meridian as that is technically not a valid contiguous space on any CRS. If you have data that crosses the anti-meridian of a CRS then your data can't be represented correctly on that CRS. This is strict and difficult to reconcile in real world applications with data crossing the poles, but it is technically true. I know you have workarounds in your other PRs since we do end up having to deal with it, but I just wanted to point this out.

As for using odc-geo in other methods of the geometry objects: no thank you. Not yet. Not unless it is a clear win for doing calculations. No need for the extra dependency.

pyresample/test/test_geometry/test_area.py
BENR0: I added a test for the case where `odc-geo` is not installed and changed the setup to include it as a requirement for testing. I thought about it and I think the test should not be skipped if `odc-geo` is not installed instead `odc-geo` should be required for testing as it is now.

mraspaud (approved):
LGTM

djhoese:
Not that I expected this to happen in this PR, but we really need to come up with a new pattern for conversion of areas to other things. I'm not sure dumping it all in geometry.py is a good idea.

PR #551, opened 2023-10-18
Fix shortcut for prime meridian checks

djhoese:
This shortcut exists in the code to avoid having to convert a CRS to a PROJ.4 dict if it isn't needed. I'm not sure how or why, but `.prime_meridian` is not an integer, it is an object, and comparing it to `== 0` will always give False from what I'm seeing. I should have always been using `.prime_meridian.longitude`.

I'm not exactly sure how to test this as it is just a performance thing (avoid unnecessary conversions). The end result should have always been the same.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
This was a pretty simple PR so just merging to get it off of my TODO list.

Issue #554, opened 2023-11-17
The finding of the optimal radius of influence makes assumption on the ordering of the dimensions in the longitude array


adybbroe:
#### Code Sample, a minimal, complete, and verifiable piece of code

See also comments in a recent Satpy PR: https://github.com/pytroll/satpy/pull/2565

```python
# Your code here

```
#### Problem description

[this should also explain **why** the current behaviour is a problem and why the 
expected output is a better solution.]

djhoese:
As mentioned in your pull request we can't assume xarray is available in Pyresample. That is standard in Satpy, but not pyresample. If I'm understanding your PR correctly, your swath DataArrays have xarray dimensions `("x", "y")`? Is that correct? In that case, I have two possible solutions that I see:

1. Update the geocentric resolution to take pixels in the middle row first two columns like it is now *and* the middle column first two rows and take the largest resolution size found.
2. Update your reader in Satpy to be `("y", "x")`. While technically it shouldn't matter, I'm a little scared that this is only one piece of the Satpy/Pyresample puzzle that you're finding an issue with, but that we make the assumption of `("y", "x")` in many places.

adybbroe:
Ok, thanks for the comments and suggestions. And, yes, the dimensions are `("x", "y")` which is non-standard, I know (see below). 

Concerning option 2: Yes, but actually in this case it is the data format which is "wrong", or non-standard. So, for now I rather keep the reader as it is, and wait for the agency (ESA) to fix the format. They have acknowledged the issue at least.

But, as the reader returns an xarray DataArray I thought it would be appropriate to actually try use that information, rather than make assumption on the data layout.

Having said that, I also like your option 1. So, would you prefer:

  1. Keep the suggestion above when the data is an xarray DataArray, and then improve the code for Numpy cases as you propose in your option 1? Or
  2. Skip using the xarray capability and improve as in your option 1, which will will work in all cases?

adybbroe:
Ok, thanks for the comments and suggestions. And, yes, the dimensions are `("x", "y")` which is non-standard, I know (see below). 

Concerning option 2: Yes, but actually in this case it is the data format which is "wrong", or non-standard. So, for now I rather keep the reader as it is, and wait for the agency (ESA) to fix the format. They have acknowledged the issue at least.

But, as the reader returns an xarray DataArray I thought it would be appropriate to actually try use that information, rather than make assumption on the data layout.

Having said that, I also like your option 1. So, would you prefer:

[...]

PR #555, opened 2023-11-17
Fix derivation of correct radius of influence when data layout is not standard

adybbroe:
This is supposed to solve #554 

Make use of the fact that longitudes are an xarray data array and use 'y' dimension for rows/scanlines

This makes the code resilient towards any other (awkward) data layout than the standard (where first dimension is usually the rows).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2010-2020 Pyresample developers
+# Copyright (C) 2010-2023 Pyresample developers
 #
 # This program is free software: you can redistribute it and/or modify it under
 # the terms of the GNU Lesser General Public License as published by the Free
@@ -693,7 +693,8 @@ class CoordinateDefinition(BaseDefinition):
         if self.ndim == 1:
             raise RuntimeError("Can't confidently determine geocentric "
                                "resolution for 1D swath.")
-        rows = self.shape[0]
+
+        rows = self.lons['y'].shape[0]
         start_row = rows // 2  # middle row
         src = CRS('+proj=latlong +datum=WGS84')
         if radius:
@@ -701,8 +702,8 @@ class CoordinateDefinition(BaseDefinition):
         else:
             dst = CRS("+proj=cart +ellps={}".format(ellps))
         # simply take the first two columns of the middle of the swath
-        lons = self.lons[start_row: start_row + 1, :2]
-        lats = self.lats[start_row: start_row + 1, :2]
+        lons = self.lons.sel(y=start_row)[:2]
+        lats = self.lats.sel(y=start_row)[:2]
         if hasattr(lons.data, 'compute'):
             # dask arrays, compute them together
             import dask.array as da


adybbroe:
It looks like I can assume that `self.lons` is an Xarray DataArray, and never (just) a numpy array, correct?

djhoese:
> It looks like I can assume that self.lons is an Xarray DataArray, and never (just) a numpy array, correct?

For Satpy, yes, but we can't (and shouldn't) make that assumption for pyresample.

adybbroe:
> > It looks like I can assume that self.lons is an Xarray DataArray, and never (just) a numpy array, correct?
> 
> For Satpy, yes, but we can't (and shouldn't) make that assumption for pyresample.

Ok, got me there then! I was looking for tests on that, but found only with xarray, but possible that I just overlooked such tests...

pyresample/geometry.py
```diff
@@ -28,6 +28,7 @@
 from typing import Optional, Sequence, Union
 
 import numpy as np
+import xarray as xr
```
djhoese, line 31: This makes xarray a hard requirement on pyresample which it is not. Plus `DataArray` is already imported in a try/except below.

pyresample/geometry.py
```diff
@@ -693,16 +694,29 @@ def geocentric_resolution(self, ellps='WGS84', radius=None, nadir_factor=2):
...
                                "resolution for 1D swath.")
-        rows = self.shape[0]
+
+        if isinstance(self.lons, xr.DataArray):
+            rows = self.lons['y'].shape[0]
+        else:
+            # Data have no information on dimensions, so we assume first dimension (the rows) is the y-axis:
+            logger.warning('As Numpy data arrays carry no information on the data layout we here ' +
+                           'assume the first dimension (the rows) is the y-axis (the satellite scans)')
+            rows = self.shape[0]
```
djhoese, line 704: So users get a warning for *every* execution of this method? No, we can't do this.

djhoese (changes requested):
Please review my comment on the issue you made:

https://github.com/pytroll/pyresample/issues/554#issuecomment-1816676723

pyresample/geometry.py
gerritholl: `warnings.warn` would in theory only issue a warning on the first execution.

djhoese: Lol right, sure, but the overall point still stands. I would still prefer that users receive no warning for a usage that has existed for years and years and is perfectly fine in 99.9999% of cases that we run into.

My preference is still that the Satpy reader reorient data, but I also have no experience with that instrument.

djhoese: Ok I take back the "years and years" thing, I thought this was a different method being modified...but still.

gerritholl: I agree that in this particular case, there should be no warning at all.

adybbroe: Ha, very good, I did trigger a reaction here! I was pretty sure you wouldn't like this when I wrote it, but wanted to discuss it. Done now. So sorry, I take it back again.

adybbroe: Ha, very good, I did trigger a reaction here! I was pretty sure you wouldn't like this when I wrote it, but wanted to discuss it. Done now. So sorry, I take it back again.

adybbroe:
@djhoese I made a comment in the issue: https://github.com/pytroll/pyresample/issues/554#issuecomment-1820556077

Which solution do you propose I pursue? And does @gerritholl and @mraspaud have opinions as well?

djhoese:
@adybbroe I'm coming back to this and it looks like in #554 we came to a conclusion of updating the method as I described in my comment. Or at least that's how I understood it. Let me know what you think here or in that thread.

PR #566, opened 2023-12-08
Refactor area boundary sides retrieval with `_geographic_sides` and `_projection_sides` methods

ghiggi:
- [ ] Closes #xxxx 
 - [ ] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented 

This PR introduces the methods:
- `_get_geographic_sides`
- `_get_projection_sides`

These two methods allow to retrieve the boundary sides for whatever `pyresample` area, whether is geostationary or not. 
In a future PR, these two methods will also include the logic to retrieve the boundary of `AreaDefinition` with out-of-Earth coordinates.

This PR:
- does not impact `get_area_slices` method
- does not impact the `AreaSlicer` and `SwathSlicer` class yet 
- does not fix GEO boundary definition issues
- does not fix boundary retrieval for out-of-earth disk areas

BUT :
- `_get_geographic_sides` is called within `get_bbox_lonlats`
- So from now on, `get_bbox_lonlats` returns non-`Inf` sides vertices for geostationary area ! 
- Since `get_edge_lonlats` and `boundary()` depends on `get_bbox_lonlats`, they now also returns non-`Inf` GEO vertices 
- Note that the order of boundary vertices of `get_bbox_lonlats` has not yet be fixed to be correct.
 
`_get_geographic_sides` 
- it's required in the coming PR to create the `SphericalBoundary` class
- `get_bbox_lonlats` is gonna be deprecated in upcoming PRs

`_get_projection_sides`
- is not yet used and tested in this PR 
- it's required for the coming PR to create the `SphericalBoundary` and `PlanarBoundary` class
- it's required for the coming PR to deprecate usage of `get_edge_bbox_in_projection_coordinates` in `SwathSlicer`/`AreaSlicer`
- it's required for the coming PR to deprecate usage of `get_geostationary_bounding_box_in_proj_coords` in `AreaSlicer`


pyresample/geometry.py
```diff
@@ -416,16 +422,6 @@ def get_edge_lonlats(self, vertices_per_side=None, frequency=None):
...
         return blons, blats
 
-    def get_edge_bbox_in_projection_coordinates(self, vertices_per_side: Optional[int] = None,
-                                                frequency: Optional[int] = None):
-        """Return the bounding box in projection coordinates."""
-        if frequency is not None:
-            warnings.warn("The `frequency` argument is pending deprecation, use `vertices_per_side` instead",
-                          PendingDeprecationWarning, stacklevel=2)
-        vertices_per_side = vertices_per_side or frequency
-        x, y = self._get_bbox_elements(self.get_proj_coords, vertices_per_side)
```
ghiggi, line 426: @djhoese there is something going wrong with the CI. 
It does not run the tests on the code I pushed. 
It comply of a in-existing `_get_bbox_elements` method, but it's not called anymore at all in the repo !

pyresample/geometry.py
```diff
@@ -1563,8 +1598,16 @@ def is_geostationary(self):
             return False
         return 'geostationary' in coord_operation.method_name.lower()
 
-    def _get_geo_boundary_sides(self, vertices_per_side=None):
-        """Retrieve the boundary sides list for geostationary projections."""
+    def _get_geostationary_boundary_sides(self, vertices_per_side=None, coordinates="geographic"):
```
djhoese, line 1601: I'm not a huge fan of this keyword argument, but this is also a private method at the moment so maybe it's OK for now. I can see how the beginning and end of the method are the same so splitting it into two separate methods based on coordinate type would be annoying. @mraspaud other ideas?

ghiggi: I would suggest to merge meanwhile, and we can split this in two functions later on ... 
With this PR merged, I can open other 3 parallel PRs.

djhoese:
Ugh, that's what I get for trying to fix something in the github GUI. Working on it...

djhoese:
I'll merge this when CI passes. Another idea I had and maybe mentioned during out last meeting:

How useful are some of these helper methods outside of boundary creation? What if the new Boundary classes had a `.from_area` or `.from_geometry` classmethod or how we talked about a factory function (something like `boundary_for_geometry(...)`)? All these little helper methods could be moved there and out of the geometry.py module.

PR #577, opened 2024-01-03
Fix area definition YAML not warning on typos

### LINKED ISSUE #570: errors in area definition should not be silently ignored

```python
from pyresample.area_config import load_area_from_string

ardef = """dyn_2km:
  description: dyn_2km
  projection:
    EPSG: 4087
  revolution: 2000
"""

load_area_from_string(ardef_dyn)
```
#### Problem description

I would like this to fail, noting that `revolution` is not a valid property/keyword argument/attribute.

#### Expected Output

I expect a failure.

The code succeeds.

pyresample main

djhoese:
If a user has a typo or some other unused parameter in an area YAML definition then pyresample silently ignores it and the user is left confused as to why things aren't working as expected. This PR fixes this by producing a UserWarning in these cases.

I could make this an error but I feel that limits the future flexibility and maintainability of the YAML reading. I could also make a configuration (`pyresample.config`) parameter that controls whether this is a warnings or error...or maybe just another keyword argument, but I'm not sure we need any more kwargs for these functions.

TODO: I'd like to rework some of the metadata handling in the future area definitions. I might not do that in this PR as it is decently separate. I noticed while working on this that any unused kwarg is put into the future AreaDefinition's `.attrs` dictionary property. This is error prone as it means that any unrecognized property will go in as "metadata". I'd like to add an explicit `attrs:` section to the YAML for defining these metadata values.

 - [x] Closes #570  
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/area_config.py
+++ b/pyresample/area_config.py
@@ -216,7 +216,25 @@ def _create_area_def_from_dict(area_name, params):
                                                                           'upper_right_xy', 'units'])
     params['resolution'] = _capture_subarguments(params, 'resolution', ['resolution', 'dx', 'dy', 'units'])
     params['radius'] = _capture_subarguments(params, 'radius', ['radius', 'dx', 'dy', 'units'])
-    area_def = create_area_def(**params)
+    area_id = params.pop("area_id", area_name)
+    kwargs = {}
+    if "description" in params:
+        kwargs["description"] = params.pop("description")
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM, thanks!

PR #582, opened 2024-02-12
Fix Cython 3 and Numpy 2 compatibility

### LINKED ISSUE #537: Upgrade to Cython 3.0 and check annotations
Cython 3 has come out and there are a lot of big changes. I've run into a couple gotchas so far and we'll need to be careful. I think we should make sure to force `cython>=3` in the pyproject.toml. Issues I've run into so far:

1. DEF/constants are being deprecated. See https://github.com/cython/cython/issues/4310. However, I haven't found good replacements for some of the use cases so I'm not sure this is the first thing we tackle.
2. `cdef` functions now default to exceptions being possible (as if you had done `except -1` or similar). For many cases of the code I write this isn't what I want. On `nogil` functions this is extra bad because it adds a GIL acquire and release to check the error state. The fix is to do `noexcept nogil:` at the end of the function declaration. Checking the HTML output of `cython -a some_module.pyx` makes it pretty clear since our `nogil` functions shouldn't have any yellow (non-C code).
3. We can and should probably drop the old numpy API by defining in setup.py:

```
        define_macros = [("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")]
```

djhoese:
* Switch to Cython 3 for building.
* Add `numpy.import_array()` for all Cython extensions to fix compatibility with Numpy 2.0

TODO: I should update wheel building to use numpy 2 when creating wheels so future wheels will already be compatible.

 - [x] Closes #537  
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


pyproject.toml
```diff
@@ -18,3 +18,19 @@ convention = "google"
 
 [tool.ruff.mccabe]
 max-complexity = 10
+
+[tool.mypy]
+python_version = "3.9"
```
mraspaud, line 23: Shouldn't we go for a newer version of python here?

mraspaud (approved):
Looks good so far!

pyproject.toml
djhoese: Pyresample hasn't dropped Python 3.9 yet so I think it needs to (or should) match what the package's compatibility is specified as (3.9+).

djhoese:
Let's see if I can get all my profiling results for gradient search organized. These are all doing ABI C02 (500m resolution) full disk to a large `eqc` target area and the final compute is going to nowhere so there is no to-disk writing involved in this. Only 4 dask workers involved. Chunk size set via PYTROLL_CHUNK_SIZE to 1356 but I think that still rounds up.

## Summary

My indexing changes are negligible in this use case. They were maybe a bit faster (~2 seconds of 47) but to me that's nothing. Both newer Cython 3 versions (the first 2 results below) are about the same as `main` on Cython 2. I'll try more workers in a second to see how it goes.

In reverse chronological/development order:

### This PR

image

### This PR but without the new column indexing

image

### Main

Note: I had to force Cython `<3` to be used here otherwise it took 6+ minutes to run because it was hitting the GIL a ton. We need to get this PR merged and released as soon as possible.

image

djhoese:
Yeah no differences with 8 workers over 4 workers.

setup.py
```diff
@@ -56,20 +57,30 @@
...
-              extra_compile_args=extra_compile_args),
+              extra_compile_args=extra_compile_args,
+              cython_directives={"language_level": 3},
+              define_macros=[("NPY_NO_DEPRECATED_API", "NPY_1_7_API_VERSION")],
+              ),
 ]
 
-cmdclass = versioneer.get_cmdclass()
+cmdclass = versioneer.get_cmdclass(cmdclass={"build_ext": build_ext})
+print(cmdclass)
```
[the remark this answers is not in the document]
mraspaud:
```suggestion
```

djhoese:
As mentioned on slack the unstable CI environment is failing due to upstream libraries like shapely and rasterio not being compatible with newest numpy 2 changes. I'll merge this PR as-is and then make another PR to test against my own custom fork of these libraries with fixes.

djhoese:
Just realized that I never updated CI to build wheels against numpy 2. I'll do that in another PR.

PR #596, opened 2024-04-26
Fix AreaDefinition array index methods mishandling outer edge values

djhoese:
An attempt to fix https://github.com/ssec/polar2grid/issues/691

Basically, current bounding operations via `get_area_slices` have trouble when the resulting lon/lat extents match *almost* exactly with the source area definition. This PR should fix a couple important bugs in the array index retrieval methods of the AreaDefinition:

1. Floating point precision issues where a X coordinate that is just outside the area (a float precision different) can still be considered part of the area (not masked).
2. Not including -0.5 to 0.0 pixel indexes as "0" (it is the left most part of the left most pixels.
3. Including an extra 0.5 pixels on the right of the area.

Locally I see one failure where my particular approach has changed the underlying value for elements that are masked and this test specifically checks that that doesn't happen (they stay unchanged).

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,6 @@ jobs:
           git+https://github.com/storpipfugl/pykdtree \
           git+https://github.com/dask/dask \
           git+https://github.com/dask/distributed \
-          git+https://github.com/zarr-developers/zarr \
           git+https://github.com/Unidata/cftime \
           git+https://github.com/mapbox/rasterio \
           git+https://github.com/pyproj4/pyproj \
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1451,11 +1451,14 @@ def masked_ints(func):
         is_scalar = np.isscalar(xm) and np.isscalar(ym)
 
         x__, y__ = func(self, xm, ym)
+        epsilon = 0.02  # arbitrary buffer for floating point precision
+        x_mask = ((x__ < -0.5 - epsilon) | (x__ > self.width - 0.5 + epsilon))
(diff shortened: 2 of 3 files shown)


pnuu (approved):
LGTM.

Just need to restart the CI jobs that were cancelled.

pnuu:
I merged with `main` to trigger the tests, there were no "re-run failed jobs" available anymore.

mraspaud (changes requested):
I'm surprised there is no test accompanying the change? That would make me understand the problem better I suppose :)

djhoese:
> I'm surprised there is no test accompanying the change? That would make me understand the problem better I suppose :)

Are you telling me you expect me to finish my PRs before you review them? And why don't you remember every conversation you've ever had?

Yes...I didn't realize how incomplete this PR was. I just remember we didn't decide on how it should work. I'll see if I can finish it today.

djhoese:
@mraspaud Tests refactored and added. The bottom line is that you (the user) are not guaranteed values outside of the area that are masked are valid underneath the mask. This is the main question.

For other similar methods like `get_lonlat_from_array_indices`, no masking is done no matter what. You can request coordinates outside of the AreaDefinition. The special thing about the method in question here is that it is returning array indices, not coordinates. There's `get_array_coordinates_from_lonlat` for that case. So...what are we allowed to do with masked values?

mraspaud (approved):
All right, I think this looks reasonable. I assuming the existing test cover other user cases, and since tests pass, feel free to merge!

djhoese:
Right. The only test that didn't pass was the one changed in this PR where it specifically checked that values under the mask (where `mask` is `True`) remained valid-ish.

Issue #354, opened 2021-05-08
`get_sum` not matched with `bucket_sum`


zxdawn:
#### Code

```python
import proplot as plot
import pyresample
import xarray as xr
from satpy import Scene
from pyresample import create_area_def
from pyresample.bucket import BucketResampler

abi_dir = '../data/GOES-16/ABI_L2/'
channel = 'HT'
reader = 'abi_l2_nc'

lon_min = -125
lon_max = -65
lat_min = 25
lat_max = 50
resample_res = 0.2 # resolution of resample grid (units: degree)

area = create_area_def('track_area',
                       {'proj': 'longlat', 'datum': 'WGS84'},
                       area_extent=[lon_min-resample_res/2,
                                    lat_min-resample_res/2,
                                    lon_max+resample_res/2,
                                    lat_max+resample_res/2],
                       resolution=resample_res,
                       units='degrees',
                       description=f'{resample_res}x{resample_res} degree lat-lon grid')

scn = Scene(['../data/GOES-16/ABI_L2/OR_ABI-L2-ACHAC-M6_G16_s20201530001131_e20201530003504_c20201530006018.nc'],
            reader=reader)
scn.load([channel])
new_scn = scn.resample(area, cache_dir=cache_dir, resample='bucket_sum')

lons, lats = scn[channel].area.get_lonlats()
resampler = BucketResampler(area,
                            xr.DataArray(lons).chunk(chunks={'dim_0':100}).data,
                            xr.DataArray(lats).chunk(chunks={'dim_0':100}).data)
sums = resampler.get_sum(scn[channel])

new_lons, new_lats = new_scn[channel].area.get_lonlats()
fig, axs = plot.subplots(axwidth=6, ncols=3, share=0)

scn[channel].plot(ax=axs[0], cmap='viridis')

m1 = axs[1].pcolormesh(new_lons, new_lats, new_scn[channel], cmap='viridis')
axs[1].colorbar(m1, loc='r')

m2 = axs[2].pcolormesh(new_lons, new_lats, sums, cmap='viridis')
axs[2].colorbar(m2, loc='r')

axs.format(collabels=['Original CTH', 'bucket_sum', 'Manual summation'])

fig.savefig('comparison.jpg')
```
#### Problem description

1. Abnormal strips of the `get_sum` plot
2. Missing data in the lower-left corner of the `bucket_sum` plot
3. `get_sum` not matched with `bucket_sum`

#### Actual Result, Traceback if applicable

[...]

PR #602, opened 2024-06-14
Add support for `fill_value` and `set_empty_bucket_to` in BucketResampler `get_sum`

ameraner:
This PR adds support for `fill_value` and `set_empty_bucket_to` in BucketResampler `get_sum`. This is a nice feature, but most importantly it fixes the satpy `bucket_sum` interface.

Closes https://github.com/pytroll/satpy/issues/2805, closes https://github.com/pytroll/satpy/issues/1746, closes https://github.com/pytroll/pyresample/issues/354

 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -238,7 +238,7 @@ class BucketResampler(object):
         data = data.ravel()
 
         # Remove fill_values values from the data when used as weights
-        invalid_mask = get_invalid_mask(data, fill_value)
+        invalid_mask = _get_invalid_mask(data, fill_value)
         weights = da.where(invalid_mask, 0, data)
 
         # Rechunk indices to match the data chunking
@@ -260,7 +260,7 @@ class BucketResampler(object):
 
     def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic, fill_value):
         if not skipna:
-            missing_val = get_invalid_mask(data, fill_value)
+            missing_val = _get_invalid_mask(data, fill_value)
             missing_val_bins, _ = da.histogram(self.idxs[missing_val], bins=out_size,
                                                range=(0, out_size))
             statistic = da.where(missing_val_bins > 0, fill_value, statistic)
@@ -469,7 +469,7 @@ class BucketResampler(object):
         return results
 
 
-def get_invalid_mask(data, fill_value):
+def _get_invalid_mask(data, fill_value):
     """Get a boolean array where values equal to fill_value in data are True."""
     if np.isnan(fill_value):
         return np.isnan(data)
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -224,7 +224,7 @@ class BucketResampler(object):
             Note that a bucket could become 0 as the result of a sum
             of positive and negative values. If the user needs to identify these zero-buckets reliably,
             `get_count()` can be used for this purpose.
-            Default: np.nan
+            Default: 0
 
         Returns
         -------
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -202,8 +202,7 @@ class BucketResampler(object):
         target_shape = self.target_area.shape
         self.idxs = self.y_idxs * target_shape[1] + self.x_idxs
 
-    def get_sum(self, data, fill_value=np.nan, skipna=True,
-                set_empty_bucket_to=0):
+    def get_sum(self, data, fill_value=np.nan, skipna=True, empty_bucket_value=0):
         """Calculate sums for each bin with drop-in-a-bucket resampling.
 
         Parameters
@@ -215,12 +214,12 @@ class BucketResampler(object):
             Default: np.nan
         skipna : boolean (optional)
             If True, skips missing values (as marked by NaN or `fill_value`) for the sum calculation
-            (similarly to Numpy's `nansum`). Buckets containing only missing values are set to `set_empty_bucket_to`.
+            (similarly to Numpy's `nansum`). Buckets containing only missing values are set to `empty_bucket_value`.
             If False, sets the bucket to fill_value if one or more missing values are present in the bucket
             (similarly to Numpy's `sum`).
-            In both cases, empty buckets are set to `set_empty_bucket_to`.
+            In both cases, empty buckets are set to `empty_bucket_value`.
             Default: True
-        set_empty_bucket_to : float
+        empty_bucket_value : float
             Set empty buckets to the given value. Empty buckets are considered as the buckets with value 0.
             Note that a bucket could become 0 as the result of a sum
             of positive and negative values. If the user needs to identify these zero-buckets reliably,
@@ -255,7 +254,9 @@ class BucketResampler(object):
 
         # TODO remove following line in favour of weights = data when dask histogram bug (issue #6935) is fixed
         sums = self._mask_bins_with_nan_if_not_skipna(skipna, data, out_size, sums, fill_value)
-        sums = da.where(sums == 0, set_empty_bucket_to, sums)
+
+        if empty_bucket_value != 0:
+            sums = da.where(sums == 0, empty_bucket_value, sums)
 
         return sums.reshape(self.target_area.shape)
 
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -210,7 +210,7 @@ class BucketResampler(object):
         data : Numpy or Dask array
             Data to be binned and summed.
         fill_value : float
-            Fill value to mark missing/invalid values in the input data.
+            Fill value of the input data marking missing/invalid values.
             Default: np.nan
         skipna : boolean (optional)
             If True, skips missing values (as marked by NaN or `fill_value`) for the sum calculation
--- a/pyresample/bucket/__init__.py
+++ b/pyresample/bucket/__init__.py
@@ -238,10 +238,8 @@ class BucketResampler(object):
         data = data.ravel()
 
         # Remove fill_values values from the data when used as weights
-        if np.isnan(fill_value):
-            weights = da.where(np.isnan(data), 0, data)
-        else:
-            weights = da.where(data == fill_value, 0, data)
+        invalid_mask = get_invalid_mask(data, fill_value)
+        weights = da.where(invalid_mask, 0, data)
 
         # Rechunk indices to match the data chunking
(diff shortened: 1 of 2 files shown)


pyresample/bucket/__init__.py
```diff
@@ -202,20 +202,30 @@ def _get_indices(self):
...
-                (similarly to Numpy's `sum`).
-                In both cases, empty buckets are set to 0.
-                Default: True
+            If True, skips missing values (as marked by NaN or `fill_value`) for the sum calculation
+            (similarly to Numpy's `nansum`). Buckets containing only missing values are set to `set_empty_bucket_to`.
+            If False, sets the bucket to fill_value if one or more missing values are present in the bucket
+            (similarly to Numpy's `sum`).
+            In both cases, empty buckets are set to `set_empty_bucket_to`.
+            Default: True
+        set_empty_bucket_to : float
```
djhoese, line 223: Can we work on a better name for this? The "set" part is a verb and seems more like the name for a function than a variable. What about `empty_fill` or something else? Also in what cases would this parameter not be 0?

ameraner: good point, thanks. `empty_fill` doesn't feel very self-explanatory to me though.. maybe `emtpy_bucket_value`?

In my use case, for example, I'm setting it to `np.nan` to maintain a sparse array after resampling, which is easier/lighter to handle and maintains transparency properties.

djhoese: enpty_bucket_value is better but "value" seems redundant. Like calling a variable "var". But I'm not sure it is that big of a deal that it *needs* changing.

ameraner: I was proposing `emtpy_bucket_value` since it goes along `fill_value`. But we can agree on `empty_bucket` :)

> But I'm not sure it is that big of a deal that it needs changing.

Do you mean the variable name or the entire concept of the new kwarg?

djhoese: I meant that `empty_bucket_value` is probably fine and good point with `fill_value` (although that might be historic with CF `_FillValue` and similar). Anyway...`empty_bucket_value` sounds good.

ameraner: Done, I also added a `if empty_bucket_value != 0:` check to avoid an unnecessary `where` call for the default case.

pyresample/bucket/__init__.py
```diff
@@ -228,8 +237,11 @@ def get_sum(self, data, skipna=True):
             data = data.data
         data = data.ravel()
 
-        # Remove NaN values from the data when used as weights
-        weights = da.where(np.isnan(data), 0, data)
+        # Remove fill_values values from the data when used as weights
+        if np.isnan(fill_value):
+            weights = da.where(np.isnan(data), 0, data)
+        else:
+            weights = da.where(data == fill_value, 0, data)
```
djhoese, line 244: Instead of the `if` statement, what about a helper for getting the invalid mask that does the nan check on the fill value and returns the appropriate mask? That way this stays `weights = da.where(invalid_mask, 0, data)`.

pyresample/bucket/__init__.py
```diff
@@ -202,20 +202,29 @@ def _get_indices(self):
...
+            If False, sets the bucket to fill_value if one or more missing values are present in the bucket
+            (similarly to Numpy's `sum`).
+            In both cases, empty buckets are set to `empty_bucket_value`.
+            Default: True
+        empty_bucket_value : float
+            Set empty buckets to the given value. Empty buckets are considered as the buckets with value 0.
+            Note that a bucket could become 0 as the result of a sum
+            of positive and negative values. If the user needs to identify these zero-buckets reliably,
+            `get_count()` can be used for this purpose.
+            Default: np.nan
```
djhoese, line 227: Why did you change the default to NaN instead of 0?

pyresample/bucket/__init__.py
```diff
@@ -241,16 +253,22 @@ def get_sum(self, data, skipna=True):
...
+    def _mask_bins_with_nan_if_not_skipna(self, skipna, data, out_size, statistic, fill_value):
         if not skipna:
-            nans = np.isnan(data)
-            nan_bins, _ = da.histogram(self.idxs[nans], bins=out_size,
-                                       range=(0, out_size))
-            statistic = da.where(nan_bins > 0, np.nan, statistic)
+            if np.isnan(fill_value):
+                missing_val = np.isnan(data)
+            else:
+                missing_val = data == fill_value
```
djhoese, line 268: This could use that same mask helper function I mentioned above.

pyresample/test/test_bucket.py
```diff
@@ -174,10 +174,61 @@ def test_get_sum_nan_data_skipna_true(self):
...
+        data = da.from_array(np.array([[2., 255], [5., 255]]),
+                             chunks=self.chunks)
+
+        result = self._get_sum_result(data, skipna=False, fill_value=255)
+        # 2 + fill_value  is fill_value, all-fill_value is fill_value
+        self.assertEqual(np.count_nonzero(result == 255), 2)
+        # 5 is untouched in a single bin
+        self.assertEqual(np.count_nonzero(result == 5.), 1)
+        # rest is 0
+        self.assertEqual(np.nanmin(result), 0)
```
djhoese, line 194: These tests all seem very similar. Could you separate these tests into a pytest-based class and use parametrization to get all the cases covered in a single test? Bonus points if you update this existing class to pytest (resampler fixture, remove setup method, remove base class, use `assert` instead of `self.assertX`).

djhoese (changes requested):

pyresample/bucket/__init__.py
ameraner: Done

pyresample/bucket/__init__.py
ameraner: Sorry, that was a mistake in the docstring, well spotted. Fixed now.

pyresample/bucket/__init__.py
ameraner: Done

pyresample/test/test_bucket.py
ameraner: Done, and I like bonus points so I moved everything to pytest :)

ameraner:
I addressed all comments, and most importantly I switched all bucket tests to pytest. I realised there are many ways to do this (like do we need a base test class containing all tests?), so please let me know if you like it like this. I also used fixtures with module scope to optimise it a bit.

I also parametrised the get_sum arguments tests into one test as proposed. It took a while to figure out the combination and simplification of the test flow, but indeed I think it improved the test readability, and made the testing of more corner cases much easier. So thanks for pushing me to do it :)

pyresample/bucket/__init__.py
```diff
@@ -456,6 +469,14 @@ def get_fractions(self, data, categories=None, fill_value=np.nan):
         return results
 
 
+def get_invalid_mask(data, fill_value):
```
djhoese, line 472: Small nitpick, maybe do a `_` prefix to this function to show that it isn't a public interface.

pyresample/bucket/__init__.py
```diff
@@ -202,20 +202,29 @@ def _get_indices(self):
...
-    def get_sum(self, data, skipna=True):
+    def get_sum(self, data, fill_value=np.nan, skipna=True, empty_bucket_value=0):
         """Calculate sums for each bin with drop-in-a-bucket resampling.
 
         Parameters
         ----------
         data : Numpy or Dask array
             Data to be binned and summed.
+        fill_value : float
+            Fill value to mark missing/invalid values in the input data.
```
djhoese, line 213: The previous wording suggest to me on first reading that the `fill_value` was used to write/assign to some array, but if I'm understanding it correctly it is already existing in the input data so I thought maybe this could be less confusing:

```suggestion
            Fill value of the input data marking missing/invalid values.
```

djhoese (changes requested):
Small suggestions, but otherwise looks good. I'd really like @pnuu to review this given he's the original author (right?).

pyresample/bucket/__init__.py
pnuu: Yep, agreed.

pnuu (commented):
LGTM, just the one nitpick-y change that Dave pointed out.

djhoese:
@ameraner I'd like to make a release soon. Any idea if you'll have time to finish this PR up soon? Looks like it might just be small things left to change. Also, does this PR also close #364?

djhoese:
CC @zxdawn about comparisons to #364

pyresample/bucket/__init__.py
ameraner: Ok, done.

pyresample/bucket/__init__.py
ameraner: indeed, fixed as proposed.

ameraner:
Hi @djhoese , thanks for the review and the ping, indeed it would be great to have this in the next release. I have pushed the modifications as requested. 

I believe it should close the PR, yes, as it implements the prosposed functionalities, even with a better granularity using the `fill_value` and `empty_bucket_value` arguments.

djhoese (approved):

pyresample/test/test_bucket.py
```diff
@@ -17,328 +17,377 @@
...
+@pytest.fixture(scope="module")
+def lats():
+    """Get latitudes for tests."""
+    return da.from_array(np.array([[60., 60.00001], [60.2, 60.3]]), chunks=CHUNKS)
+
+
+@pytest.fixture(scope="module")
+def resampler(adef, lons, lats):
+    """Get initialised resampler for tests."""
+    return bucket.BucketResampler(adef, lons, lats)
```
djhoese, line 69: Is it a problem to make this module scoped when it seems the resampler object caches stuff in the instance variables:

```python
        self.x_idxs = None
        self.y_idxs = None
        self.y_idxs = None
        self.idxs = None
        self.idxs = None
        self._get_indices()
        self._get_indices()
        self.counts = None
        self.counts = None
```

zxdawn:
I agree this can close my old PR because it considers both nan and fill_value. Thanks!

pyresample/test/test_bucket.py
ameraner: The cached variables are related to the indices computed from the input lat, lon and target areadef, which are also module-scoped fixtures. So in this setting I think this is fine. As a cross-check, I just switched to function-scoped and all tests still pass, so there is no cross-contamination. Also, the tests that need different inputs or patched/mocked components do not use the resampler fixture.

PR #610, opened 2024-07-25
Fix SwathDefinition html representation error when lons/lats 1D

### LINKED ISSUE #609: Error in SwathDefinition html representation if lon/lat arrays are dask arrays

This was discovered while using the `li_l2_nc` reader but affects all `SwathDefinitions` where the lat/lon `DataArrays` don't have a `resolution` set in the attributes.

#### Problem description

If lon/lat data in the `SwathDefinition` are dask arrays the html representation throws a type error due to `None Type` not having a round method because no resolution can be found in the attributes. 

Especially in the Li point data case this problem might not be easily solveable.

#### Expected Output
A html represetation of the `SwathDefinition`. 

```Python

    260     # use resolution from lat/lons dataarray attributes -> are these always set? -> Maybe try/except?
    261     area_name = f"{lon_attrs.get('sensor')} swath"
--> 262     resolution_str = "/".join([str(round(x.get("resolution"), 1)) for x in [lat_attrs, lon_attrs]])
    263     area_units = "m"
    265 height, width = area.lons.shape

TypeError: type NoneType doesn't define __round__ method
```

BENR0:
This is a quick fix for #609. It just sets things like resolution and width/height to Na in the case of 1D lat/lon arrays so the html representation does not break.
It also does some quick fixing of formatting issues I noticed.

 - [x] Closes #609 
 - [ ] Tests added 
 - [ ] Tests passed 

### DIFF
--- a/pyresample/_formatting_html.py
+++ b/pyresample/_formatting_html.py
@@ -246,25 +246,25 @@ def swath_area_attrs_section(area: 'geom.SwathDefinition') -> str: # noqa F821
     """
     if np.ndim(area.lons) == 1:
         area_name = "1D Swath"
-        resolution_str = "{resolution}/{resolution}".format(resolution="Na")
-        height, width = "Na", "Na"
+        resolution_str = "NA/NA"
+        height, width = "NA", "NA"
         area_units = "m"
     else:
-        if isinstance(area.lons, np.ndarray) & isinstance(area.lats, np.ndarray):
+        if isinstance(area.lons, np.ndarray) and isinstance(area.lats, np.ndarray):
             # Calculate and estimated resolution from lats/lons in meter
             area_name = "Arbitrary Swath"
             resolution_y = np.mean(area.lats[0:-1, :] - area.lats[1::, :])
             resolution_x = np.mean(area.lons[:, 1::] - area.lons[:, 0:-1])
             resolution = np.mean(np.array([resolution_x, resolution_y]))
             resolution = np.round(40075000 * resolution / 360, 1)
-            resolution_str = f"{resolution}/{resolution}"
+            resolution_str = f"{resolution}x{resolution}"
         else:
             lon_attrs = area.lons.attrs
             lat_attrs = area.lats.attrs
 
             # use resolution from lat/lons dataarray attributes -> are these always set? -> Maybe try/except?
             area_name = f"{lon_attrs.get('sensor')} Swath"
-            resolution_str = "/".join([str(round(x.get("resolution"), 1)) for x in [lat_attrs, lon_attrs]])
+            resolution_str = "x".join([str(round(x.get("resolution"), 1)) for x in [lat_attrs, lon_attrs]])
 
         area_units = "m"
         height, width = area.lons.shape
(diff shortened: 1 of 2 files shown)


pyresample/_formatting_html.py
```diff
@@ -244,25 +244,30 @@ def swath_area_attrs_section(area: 'geom.SwathDefinition') -> str: # noqa F821
...
-        resolution_str = f"{resolution}/{resolution}"
+    if np.ndim(area.lons) == 1:
+        area_name = "1D Swath"
+        resolution_str = "{resolution}/{resolution}".format(resolution="Na")
+        height, width = "Na", "Na"
         area_units = "m"
     else:
-        lon_attrs = area.lons.attrs
-        lat_attrs = area.lats.attrs
+        if isinstance(area.lons, np.ndarray) & isinstance(area.lats, np.ndarray):
```
djhoese, line 253:
```suggestion
        if isinstance(area.lons, np.ndarray) and isinstance(area.lats, np.ndarray):
```

djhoese (changes requested):

pyresample/_formatting_html.py
```diff
@@ -244,25 +244,30 @@ def swath_area_attrs_section(area: 'geom.SwathDefinition') -> str: # noqa F821
...
-        area_name = "Arbitrary Swath"
-        resolution_y = np.mean(area.lats[0:-1, :] - area.lats[1::, :])
-        resolution_x = np.mean(area.lons[:, 1::] - area.lons[:, 0:-1])
-        resolution = np.mean(np.array([resolution_x, resolution_y]))
-        resolution = np.round(40075000 * resolution / 360, 1)
-        resolution_str = f"{resolution}/{resolution}"
+    if np.ndim(area.lons) == 1:
+        area_name = "1D Swath"
+        resolution_str = "{resolution}/{resolution}".format(resolution="Na")
+        height, width = "Na", "Na"
```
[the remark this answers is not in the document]
BENR0: Yeah you are right, I will just hard code it. I just took the code from the other cases but does not make sense here. Good idea with the "x" instead of "/" definitely makes more sense.

Well as mentioned somewhere before the resolution calculation in case of the attributes not having resolution set is just a rough estimate and therefore the circumference of the earth (40075000) is just hard coded.

pyresample/_formatting_html.py
```diff
@@ -246,25 +246,25 @@ def swath_area_attrs_section(area: 'geom.SwathDefinition') -> str: # noqa F821
     """
     if np.ndim(area.lons) == 1:
         area_name = "1D Swath"
-        resolution_str = "{resolution}/{resolution}".format(resolution="Na")
-        height, width = "Na", "Na"
+        resolution_str = "NA/NA"
```
djhoese, line 249:
```suggestion
        resolution_str = "NAxNA"
```

Right?

djhoese (approved):

djhoese:
Looks like the unstable environment are failing for some related things, but not *only* related things. I wonder if this is something we could/should fix in this PR:

```
 FAILED pyresample/test/test_formatting.py::test_plot_area_def_w_swath_def[LegacySwathDefinition] - TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
FAILED pyresample/test/test_formatting.py::test_plot_area_def_w_swath_def[SwathDefinition] - TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
FAILED pyresample/test/test_formatting.py::test_area_def_cartopy_installed - TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
FAILED pyresample/test/test_formatting.py::test_area_repr_w_static_files - TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
FAILED pyresample/test/test_formatting.py::test_area_repr_wo_static_files - TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
FAILED pyresample/test/test_sgeom/test_point.py::TestSMultiPoint::test_to_shapely - TypeError: ufunc 'create_collection' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

PR #614, opened 2024-08-23
Fix geocentric_resolution compatibility with numpy 2.1.0

djhoese:
Satpy unstable environment started failing and regular environments will start failing too when conda-forge has numpy 2.1.0 available.

Starting with numpy 2.1.0 the numpy function `histogram_bin_edges` will raise a `ValueError` if the array you give it can't be split into the number of buckets you request (default 10). This happens when the elements of an array are so close together (near floating point precision) that numpy can't possibly split things into N buckets. There error you get is:

```
E               ValueError: Too many bins for data range. Cannot create 10 finite-sized bins.
```

This PR works around this by adding a try/except to the call and just returning the two elements of the input array. As far as I can tell this is basically equivalent.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented


djhoese:
Hm pre-commit on my local machine is behaving differently than pre-commit.ci.

mraspaud (approved):
Lgtm, thanks for updating the PR template along the way.

PR #618, opened 2024-09-02
Fix dtype for swath -> area resampling with gradient search

pnuu:
Resampling `float32` swath data resulted in `float64` output. This is a simple fix for this data type promotion.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/gradient/__init__.py
+++ b/pyresample/gradient/__init__.py
@@ -421,6 +421,7 @@ def parallel_gradient_search(data, src_x, src_y, dst_x, dst_y,
                 dst_x[i], dst_y[i],
                 method=method)
             res = da.from_delayed(res, (num_bands, ) + dst_x[i].shape,
+                                  meta=np.array((), dtype=dtype),
                                   dtype=arr.dtype)
         if dst_mosaic_locations[i] in chunks:
             if not is_pad:
--- a/pyresample/gradient/__init__.py
+++ b/pyresample/gradient/__init__.py
@@ -421,7 +421,7 @@ def parallel_gradient_search(data, src_x, src_y, dst_x, dst_y,
                 dst_x[i], dst_y[i],
                 method=method)
             res = da.from_delayed(res, (num_bands, ) + dst_x[i].shape,
-                                  meta=np.array((), dtype=dtype),
+                                  meta=np.array((), dtype=arr.dtype),
                                   dtype=arr.dtype)
         if dst_mosaic_locations[i] in chunks:
             if not is_pad:
--- a/pyresample/gradient/__init__.py
+++ b/pyresample/gradient/__init__.py
@@ -414,14 +414,14 @@ def parallel_gradient_search(data, src_x, src_y, dst_x, dst_y,
         else:
             is_pad = False
             res = dask.delayed(_gradient_resample_data)(
-                arr.astype(np.float64),
+                arr,
                 src_x[i], src_y[i],
                 src_gradient_xl[i], src_gradient_xp[i],
                 src_gradient_yl[i], src_gradient_yp[i],
                 dst_x[i], dst_y[i],
                 method=method)
             res = da.from_delayed(res, (num_bands, ) + dst_x[i].shape,
-                                  dtype=np.float64).astype(arr.dtype)
+                                  dtype=arr.dtype)
         if dst_mosaic_locations[i] in chunks:
             if not is_pad:
                 chunks[dst_mosaic_locations[i]].append(res)
--- a/pyresample/gradient/__init__.py
+++ b/pyresample/gradient/__init__.py
@@ -421,7 +421,7 @@ def parallel_gradient_search(data, src_x, src_y, dst_x, dst_y,
                 dst_x[i], dst_y[i],
                 method=method)
             res = da.from_delayed(res, (num_bands, ) + dst_x[i].shape,
-                                  dtype=np.float64)
+                                  dtype=np.float64).astype(arr.dtype)
         if dst_mosaic_locations[i] in chunks:
             if not is_pad:
                 chunks[dst_mosaic_locations[i]].append(res)
--- a/pyresample/gradient/_gradient_search.pyx
+++ b/pyresample/gradient/_gradient_search.pyx
@@ -27,22 +27,19 @@ cimport cython
 cimport numpy as np
 from libc.math cimport fabs, isinf
 
-ctypedef np.double_t DTYPE_t
-
 ctypedef fused data_type:
     double
     float
 
-ctypedef double float_index
-f_index_dtype = float
-ctypedef int i_index_type
-
+ctypedef fused float_index:
+    double
+    float
 
 np.import_array()
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cdef inline void nn(const data_type[:, :, :] data, i_index_type l0, int p0, double dl, double dp, int lmax, int pmax, data_type[:] res) noexcept nogil:
+cdef inline void nn(const data_type[:, :, :] data, int l0, int p0, float_index dl, float_index dp, int lmax, int pmax, data_type[:] res) noexcept nogil:
     cdef int nnl, nnp
     cdef size_t z_size = res.shape[0]
     cdef size_t i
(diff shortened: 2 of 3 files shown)


pnuu:
Performance comparisons with my 12 AVHRR composites:

Satpy `main`, Pyresample `main`:
Screenshot 2024-09-02 at 14-44-01 Bokeh Plot

Satpy https://github.com/pytroll/satpy/pull/2893 , Pyresample `main`: 
Screenshot 2024-09-02 at 14-45-19 Bokeh Plot

Satpy https://github.com/pytroll/satpy/pull/2893 , Pyresample: this PR
Screenshot 2024-09-02 at 14-46-07 Bokeh Plot

So from 5.6 GB to 3.5 GB of RAM and from 87.3 s to 68.6 s processing time.

pyresample/gradient/__init__.py
```diff
@@ -421,7 +421,7 @@ def parallel_gradient_search(data, src_x, src_y, dst_x, dst_y,
                 dst_x[i], dst_y[i],
                 method=method)
             res = da.from_delayed(res, (num_bands, ) + dst_x[i].shape,
-                                  dtype=np.float64)
+                                  dtype=np.float64).astype(arr.dtype)
```
mraspaud, line 424: Ok, I think we can do better than this. What I see here (and a bit above is that we convert the input data to float64, do the resampling, then convert back to float32.
Using cython fused types, we should be able to template the cython function to accept both float64, float32, or even ints. [This cython doc section](http://docs.cython.org/en/latest/src/userguide/fusedtypes.html#selecting-specializations) should help. I can give you a hand if you don't feel comfortable with this.

mraspaud:
@pnuu I pushed a couple of commits that introduce fused types in the cython so that both float64 and float32 input data is allowed and handled correctly in principle.
I also made a fused type for the coordinates, in case we want to experiment with float32 for these also. But that will require changes in the python code which I haven't looked at.

pnuu:
@mraspaud definitely a better solution!

I don't see a difference in the performance though:
Screenshot 2024-09-03 at 14-17-44 Bokeh Plot

mraspaud:
@pnuu did you recompile the cython code?

pnuu:
Yeah, I ran `pip install -e .` after which I saw a modified docstring chang in import. I now also did `python setup.py build_ext --inplace` and got this graph (no real change)
Screenshot 2024-09-03 at 15-00-55 Bokeh Plot

mraspaud:
After a discussion on slack, it was noted that float32 is not precise enough for cartesian coordinates at least, so I removed the fused type for coordinates, only allowing float64 (double in cython)

mraspaud:
@pnuu where are we on this one?

pnuu:
I think we're good and ready to merge. The RAM footpring nor processing time didn't change with the coordinate dtype de-fusing from the previous graph above. All the datasets stay as they were, for example with https://github.com/pytroll/satpy/pull/2893 the datasets/composites stay `float32` as without this PR the were converted to `float64`.

mraspaud:
@djhoese would you mind giving your review on this?

pyresample/test/test_gradient.py
```diff
@@ -249,25 +249,29 @@ def test_resample_area_to_area_3d_single_channel(self):
...
+    @pytest.mark.parametrize("input_dtype", (np.float32, np.float64))
+    def test_resample_swath_to_area_2d(self, input_dtype):
         """Resample swath to area, 2d."""
-        data = xr.DataArray(da.ones(self.src_swath.shape, dtype=np.float64),
+        data = xr.DataArray(da.ones(self.src_swath.shape, dtype=input_dtype),
                             dims=['y', 'x'])
         with np.errstate(invalid="ignore"):  # 'inf' space pixels cause runtime warnings
             res = self.swath_resampler.compute(
                 data, method='bil').compute(scheduler='single-threaded')
+        assert res.dtype == data.dtype
```
djhoese, line 260: Sorry for this small but annoying request, but could you get the dask-based result and compute the numpy result from it and compare their dtypes? It is the fastest way to catch dask bugs where dask thinks it is one dtype but the final computed array is another.

pyresample/test/test_gradient.py
```diff
@@ -249,25 +249,29 @@ def test_resample_area_to_area_3d_single_channel(self):
...
         """Resample area to area, 3d."""
         data = xr.DataArray(da.ones((3, ) + self.src_swath.shape,
-                                    dtype=np.float64) *
+                                    dtype=input_dtype) *
                             np.array([1, 2, 3])[:, np.newaxis, np.newaxis],
                             dims=['bands', 'y', 'x'])
         with np.errstate(invalid="ignore"):  # 'inf' space pixels cause runtime warnings
             res = self.swath_resampler.compute(
                 data, method='bil').compute(scheduler='single-threaded')
+        assert res.dtype == data.dtype
```
djhoese, line 274: dask and numpy dtype check here too

pyresample/gradient/__init__.py
```diff
@@ -414,14 +414,14 @@ def parallel_gradient_search(data, src_x, src_y, dst_x, dst_y,
...
-                arr.astype(np.float64),
+                arr,
                 src_x[i], src_y[i],
                 src_gradient_xl[i], src_gradient_xp[i],
                 src_gradient_yl[i], src_gradient_yp[i],
                 dst_x[i], dst_y[i],
                 method=method)
             res = da.from_delayed(res, (num_bands, ) + dst_x[i].shape,
-                                  dtype=np.float64)
+                                  dtype=arr.dtype)
```
djhoese, line 424: This avoids some extra dask checks.

```suggestion
                                  meta=np.array((), dtype=dtype),
                                  dtype=arr.dtype)
```

pyresample/gradient/_gradient_search.pyx
```diff
@@ -23,18 +23,22 @@
...
-ctypedef np.double_t DTYPE_t
 cimport cython
+cimport numpy as np
 from libc.math cimport fabs, isinf
 
+ctypedef fused data_type:
+    double
+    float
+
+ctypedef double float_index
```
djhoese, line 34:
```suggestion
ctypedef np.float64_t float_index
```

pyresample/gradient/_gradient_search.pyx
```diff
@@ -23,18 +23,22 @@
...
-
-DTYPE = np.double
-ctypedef np.double_t DTYPE_t
 cimport cython
+cimport numpy as np
 from libc.math cimport fabs, isinf
 
+ctypedef fused data_type:
+    double
+    float
```
djhoese, line 32: I think we could be more explicit that we are using numpy types. Thoughts?

```suggestion
    np.float64_t
    np.float32_t
```

djhoese (changes requested):
A couple small requests.

pyresample/test/test_gradient.py
mraspaud: Very good idea indeed! Let's see what I can come up with.

mraspaud: I used `values` in the end, tell me what you think.

djhoese: You *could* even assert that the returned array type is dask.

mraspaud: > Correct me if I'm wrong, but the above `res = ` line includes a `.compute(scheduler='single-threaded')` so the result of that is already dask-computed. I think you'd need to split that line into two steps: one for the dask array based result, then the computed result (numpy), then do the `res_np` and `res_dask` comparison (or whatever you end up naming the two variables).

You right of course, sorry about that. Should be fixed now. I'm also making sure the shape is right both with numpy and dask arrays.

mraspaud: > You _could_ even assert that the returned array type is dask.

In the spirit of duck typing, I will avoid that.

djhoese: Yeah, but this is the tests so I disagree. You are passing in a specific type, you expect a specific type back. If you got a numpy array when you provided a dask array as input then that would be a big bug. Not to mention you could/should check for how many dask computations are happening during the processing and make sure it is zero (or whatever you expect).

Duck typing makes sense in the work code, not the test code...in my opinion.

pnuu:
Thanks for the updates and tweaks @mraspaud!

pnuu:
Yikes, the resulting image has transposed!

day_microphysical_3a_with_night_microphysical_20240901_200143-0

pnuu:
Apparently this is the case also with `main`. The kd-tree based nearest neighbour resampling produces images that are oriented properly:

day_microphysical_3a_with_night_microphysical_20240901_200143-0

mraspaud:
> Yikes, the resulting image has transposed!
> 
> day_microphysical_3a_with_night_microphysical_20240901_200143-0

What???

pnuu:
@mraspaud this seems to be a problem only with polar data (AVHRR L1b in my test case), so `StackingGradientSearchResampler` class. The `ResampleBlocksGradientSearchResampler` class used for data having `AreaDefinition` has been working properly.

mraspaud:
@pnuu ok, if this a problem also in main, can we merge this and fix the other issue in another PR?

pnuu:
@mraspaud yes, I'm fine with that :+1:

PR #624, opened 2024-10-07
Move setuptools to setup_requires

avalentino:
Apparently `setuptools` is not required at runtime

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [X] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ from setuptools import find_packages, setup
 
 import versioneer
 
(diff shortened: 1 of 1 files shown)


djhoese (approved):
I was pretty sure we used setuptools somewhere inside pyresample, but it looks like that is all done through importlib now so I think this is fine.

djhoese:
...and I should have said: thanks.

PR #645, opened 2025-01-30
Investigate antimeridian improvements and pyproj updates

djhoese:
I had these changes lying around and needed to get them out of the way. Maybe I'll get back to this soon...for now it is here for comment if people are really really bored.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/future/resamplers/nearest.py
+++ b/pyresample/future/resamplers/nearest.py
@@ -41,8 +41,8 @@ except ImportError:
     dask = None
 
 
-def query_no_distance(target_lons, target_lats, valid_output_index,
-                      mask=None, valid_input_index=None,
+def query_no_distance(target_lons, target_lats,
+                      mask=None,
                       neighbours=None, epsilon=None, radius=None,
                       kdtree=None):
     """Query the kdtree. No distances are returned.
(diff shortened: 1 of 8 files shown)


djhoese:
Needs versioneer fixes from #648

djhoese:
@mraspaud @pnuu I'm curious what you guys think about this PR and if I should split it up or go further or I don't know what. The main important change that I wanted was the DynamicAreaDefinition's change to use `+lon_wrap=180` which allows extents to go beyond 180 longitude (0-360 essentially). However, that gets complicated because for that to be useful the nearest neighbor resampling also needs to support those coordinates which means removing the valid input/output index arrays. Then we get more complicated since there is base kdtree resampling (numpy-style), then an xarray-based/dask class used by satpy with caching, and an xarray/dask class with included caching in the "future" subpackage of pyresample.

Do I maybe switch this PR to be just the DynamicAreaDefinition changes? Then do another PR for the future class to drop valid input/output arrays? Then do another PR for the xarray-based class (and maybe the numpy functions) to drop valid input/output? I'm not sure.

This doesn't even begin to talk about my real desire which is to explore the XYZ transformation by pyproj changes after all of this (which I have working locally).

pnuu:
I'd say do the `DynamicAreaDefinition` first and the others later on.

Just a thought on the input/output array stuff perhaps be adapted based on the validity range of the coordinates? I don't remember exactly, but I think at the masking we would know whether the coordinates are from -180 to 180 or from 0 to 360. I'm not say not to remove them if it makes new things possible, just a thought.

djhoese:
For validity arrays, I think the bottom line, from what I can tell, is that in modern practices they are unnecessary. From the sound of it, most users are probably masking their geolocation already or Satpy is or the input files people are reading do. Or as a last resort the computations by pyproj will make them infinity or NaN.

PR #648, opened 2025-02-17
Update EWA pyproj Proj usage to Transformer

djhoese:
The pyproj `Proj` object is deprecated and is not as accurate/explicit as the `Transformer` interface. I did some of these changes in my #645 branch, but decided that these changes should be their own PR. Unfortunately this also meant that I changed the low-level interfaces of ll2cr to accept a source CRS and a target CRS instead of just the target projection information.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/ewa/_ll2cr.pyx
+++ b/pyresample/ewa/_ll2cr.pyx
@@ -23,10 +23,13 @@
 __docformat__ = "restructuredtext en"
 
 import numpy
-from pyproj import Proj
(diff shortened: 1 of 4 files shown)


djhoese:
Merging this to get the versioneer fixes and also I'm the primary EWA user anyway so...it should be fine and if not I'll fix it.

PR #663, opened 2025-05-29
Update API documentation

djhoese:
This PR adopts the same API strategy as Satpy. It also fixes various reference warnings/issues.

Note: I have not run sphinx in nitpicky mode to look at *every* reference issue. I'm not sure I'm ready to go down that rabbit hole yet.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
This also adds documentation for private (`_`) interfaces including the legacy EWA ones that Satpy still references in its documentation.

djhoese:
Side note: @pnuu, in the bilinear resampler, there was some warning about the way delayed was used. I didn't understand the exact error message but did notice the Delayed function was a method on the resampler class. This almost always a problem for dask in distributed processing as it means the entire resampler class has to be serialized and sent between nodes. I fixed that by using the `slicer` function directly. Additionally, in the long run, we need to replace the Delayed usage with a `blockwise` or `map_blocks` call with a single large chunk to get around dask not optimizing task graphs with mixed delayed and array usages.

Nilzkool:
Hi @djhoese , I noticed you’ve been keeping docs up-to-date, always good to see.

I built a small LLM-based tool for maintaining high-level docs (SDK guides, API references, READMEs, tutorials). It auto-updates them on PRs to prevent drift.

No pressure at all, but thought it might be useful: https://github.com/marketplace/deepdocsai

pnuu (approved):

pnuu:
@djhoese the bilinear interpolator would also need to be adjusted to work properly with `xr.DataArray` dimension similar to `nearest`.

PR #669, opened 2025-07-16
Remove unused dependency importlib_metadata

karyotakisg:
As part of our ongoing research on Python dependency management we noticed a potential improvement in your project’s dependency management.

Specifically, the dependency `importlib_metadata` is specified as a requirement in the `setup.py` file, when in reality it is not needed.

This PR removes it from `setup.py`, which helps keeping the dependency list clean.
Hope this is helpful!

Best regards

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,8 +29,6 @@ requirements = ['pyproj>=3.0', 'configobj',
                 "shapely", "donfig", "platformdirs",
                 ]
 
-if sys.version_info < (3, 10):
-    requirements.append('importlib_metadata')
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -29,8 +29,6 @@
                 "shapely", "donfig", "platformdirs",
                 ]
 
-if sys.version_info < (3, 10):
-    requirements.append('importlib_metadata')
```
djhoese, line 32:
```suggestion
```

djhoese:
Ugh CI is failing because of new zarr. I'll have to look at that a different time.

PR #675, opened 2025-08-12
Fix BucketResampler using area's proj_dict instead crs

djhoese:
This is causing pyproj warnings about loss of precision in Satpy tests.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


pnuu (approved):
LGTM, thanks for the fix!

djhoese:
My goodness, now coveralls is down.

PR #677, opened 2025-08-18
Fix boundaries losing precision for accurate polygon calculations

djhoese:
In a real world case identified by @spruceboy, if a 32-bit coordinate is given to `AreaBoundary` (or any boundary class) that is exactly at 90 degrees north, the conversion to radians and then validity checks in the spherical polygon class will raise an error.

```python
  File "/opt/cspp/polar2grid_v_3_1/libexec/python_runtime/lib/python3.11/site-packages/pyresample/spherical.py", line 571, in aedges
    yield Arc(SCoordinate(lon_start, lat_start),
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/cspp/polar2grid_v_3_1/libexec/python_runtime/lib/python3.11/site-packages/pyresample/spherical.py", line 131, in __init__
    lon, lat = _check_lon_lat(lon, lat)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/cspp/polar2grid_v_3_1/libexec/python_runtime/lib/python3.11/site-packages/pyresample/spherical.py", line 119, in _check_lon_lat
    _check_lat_validity(lat)
  File "/opt/cspp/polar2grid_v_3_1/libexec/python_runtime/lib/python3.11/site-packages/pyresample/spherical.py", line 111, in _check_lat_validity
    raise ValueError("Latitude values must range between [-pi/2, pi/2].")
```

What I discovered is that due to precision issues you end up with the input coordinate being `4.371139006309477e-08` greater than `np.pi / 2` and failing the validity check. See https://github.com/ssec/polar2grid/issues/766 for more info. 

The alternative solution could be allowing an epsilon greater/less than 90/-90 in the validity check but then the question becomes how to handle that coordinate? Should it be clipped? Should it be left alone and hope that the math is "good enough". Maybe it is just better to be more accurate with this PR's fix.

CC @ghiggi

@spruceboy, feel free to hack your Polar2Grid installation and add the `dtype=np.float64` shown in the changes here for `pyresample/boundary/legacy_boundary.py`.

 - [ ] Closes #xxxx 
 - [x] Tests added 
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/boundary/legacy_boundary.py
+++ b/pyresample/boundary/legacy_boundary.py
@@ -45,8 +45,9 @@ class Boundary(object):
     def contour_poly(self):
         """Get the Spherical polygon corresponding to the Boundary."""
         if self._contour_poly is None:
+            # force 64-bit for more accuracy
             self._contour_poly = SphPolygon(
-                np.deg2rad(np.vstack(self.contour()).T))
+                np.deg2rad(np.vstack(self.contour()).T, dtype=np.float64))
         return self._contour_poly
 
     def draw(self, mapper, options, **more_options):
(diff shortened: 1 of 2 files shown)


mraspaud (approved):
LGTM!

PR #699, opened 2026-02-21
Add units parameter to resolution warning log - at least in some cases

### LINKED ISSUE #458: Warning log message is incorrectly showing resolution as 'meters'
#### Problem description

The code here:
https://github.com/pytroll/pyresample/blob/main/pyresample/area_config.py#L724

Raises a warning log message if the output area size are not a whole number. This is fine, but the message uses the word `meters` even in cases where the input resolution is given in degrees.

### Example
```python
area_ext = (-180, -65, 180, 65)
res = (0.03, 0.03)
targ_srs = create_area_def("source_area", "EPSG:4326", area_extent=area_ext, resolution=res)
```
Gives this message:
```
WARNING:root:shape found from radius and resolution does not contain only integers: (4333.333333333334, 12000.0)
Rounding shape to (4334, 12000) and resolution from (0.03, 0.03) meters to (0.03, 0.029995385325334564) meters
```

But the resolution passed to `create_area_def` is in degrees...

LeviDPC:
Pull request to fix  #458 (and fixes the example of the problem given in that ticket.)

It does so by adding an optional parameter units to the _round_shape function and having _extrapolate_information give the appropriate units when calling _round_shape. 

This should fix all instances of this as the warning only has the potential to be thrown when resolution and radius are both set and the only other instance of calling _round_shape (in _format_list) does not set those parameters.

### DIFF
--- a/pyresample/area_config.py
+++ b/pyresample/area_config.py
@@ -739,7 +739,7 @@ def _convert_coordinate_for_metered_units(var, units: str, src_crs: CRS, dst_crs
     return transformer.transform(*var)
 
 
-def _round_shape(shape, radius=None, resolution=None):
+def _round_shape(shape, radius=None, resolution=None, units="meters"):
     """Make sure shape is an integer.
 
     Rounds down if shape is less than .01 above nearest whole number to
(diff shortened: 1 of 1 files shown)


djhoese:
Thanks Levi!

PR #708, opened 2026-03-02
Improve nearest-neighbor query perf

Zaczero:
query_no_distance:
runtime: `0.01383s -> 0.01279s` (-7.5%).
rss: `221096KB -> 204204KB` (-7.6%).

_verify_input_object_type numpy array:
runtime: `0.02705s -> 0.02002s` (-26%).
rss: same

_compute_radius_of_influence nan case:
runtime: `19.71s -> 2.10s`.
rss: same

 - [x] Tests added 
 - [x] Tests passed

### DIFF
--- a/pyresample/future/resamplers/nearest.py
+++ b/pyresample/future/resamplers/nearest.py
@@ -66,22 +66,32 @@ def query_no_distance(target_lons, target_lats, valid_output_index,
         distance_upper_bound=radius,
         mask=mask)
 
+    if neighbours == 1:
+        # Nearest-neighbor resampling only consumes one neighbor, so avoid
+        # building and masking the full (rows, cols, neighbours) array shape.
+        index_array = np.asarray(index_array, dtype=np.int64)
+        index_array[index_array >= kdtree.n] = -1
+        out = np.full(voi.size, -1, dtype=np.int64)
+        out[voir] = index_array
+        return out.reshape(voi.shape + (1,))
+
     if index_array.ndim == 1:
         index_array = index_array[:, None]
 
     # KDTree query returns out-of-bounds neighbors as `len(arr)`
     # which is an invalid index, we mask those out so -1 represents
     # invalid values
+    #
     # voi is 2D (trows, tcols)
     # index_array is 2D (valid output pixels, neighbors)
     # there are as many Trues in voi as rows in index_array
-    good_pixels = index_array < kdtree.n
-    res_ia = np.empty(shape, dtype=int)
-    mask = np.zeros(shape, dtype=bool)
-    mask[voi, :] = good_pixels
-    res_ia[mask] = index_array[good_pixels]
-    res_ia[~mask] = -1
-    return res_ia
+    #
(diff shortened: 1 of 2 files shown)


djhoese (commented):
Nice job. Thanks for finding this. I've always disliked the indexing done in the query so I'm glad to see it cleaned up. I had a couple questions, but otherwise the majority of this looks good to me.

Otherwise could you check pre-commit's failures? Thanks.

Zaczero:
> In your original description you talked about the performance of `_verify_input_object_type`. How does this PR affect that? Where is this function?

pyresample/future/resamplers/nearest.py:496 replaces data.copy() with data.copy(deep=False) before swapping .data to a dask array. That avoids a deep copy of the underlying NumPy buffer.

> Oh and what was your test case for your timing? I'm surprised by how fast the query stuff was, but even more by how slow the NaN case was for the radius of influence stuff.

I don't remember now, but I was measuring the specific code paths in isolation, not in a bigger system. I like efficient software.

> One last thing: I'm not ignoring your other pyresample and satpy PRs, but I'm still catching up on real work after being out for a couple days and I'm trying to fully understand the changes before commenting. Thanks again for all these changes/fixes/improvements.

Np, I was delaying too because I started working on the data I got processed :smile:.

djhoese:
> I don't remember now, but I was measuring the specific code paths in isolation, not in a bigger system. I like efficient software.

Were you running things multiple times? I'm unable to get my laptop with a real world case to perform in a consistently better way. I'll try a few more things to sanity check...

djhoese:
I don't remember what data you said you're processing, but I'm taking ABI L1b data (G16) and resampling it using nearest neighbor to the Satpy builtin "brazil" area. With current pyresample `main` I first got 11.5s so I switched to this branch and got 10.87s so I thought "awesome, faster". Then I ran it over and over again and got numbers from 11.63s to 11.86s. Then I went back to `main` and got 11.11s to 11.83s.

```python
from datetime import datetime
from satpy import Scene
from glob import glob

scn = Scene(reader="abi_l1b", filenames=glob("/data/satellite/abi/20180623/OR_ABI-L1b-RadF-M3C*G16*.nc"))
scn.load(["C02"])

first = datetime.now()
new_scn = scn.resample("brazil", resampler="nearest")
new_scn["C02"].compute()
#new_scn.save_datasets(filename="C02_brazil.tif")
second = datetime.now()

print((second - first).total_seconds())

```

Zaczero:
In my PR body, I performed 3 distinct measurements on the individual methods directly, not the full pipeline. I do confirm that the improvements are still reproducible. I've also noticed I implemented one more valuable specialization to query_no_distance that brings 6% performance improvement in the neighbors=1 common case.

djhoese:
Thanks. I've been off of work the last week or so but am catching up on things today. I'll try to review this when I get a chance. And yeah if you're testing the individual functions it makes sense to see these low-level improvements but then I not see them on the high-level when the majority of time is spent building the KDTree and the actual query time.

mraspaud (approved):
LGTM, thanks for taking the time to simplify the processing!

PR #712, opened 2026-03-09
use vertices_per_side argument rather than frequency which is deprecated

### LINKED ISSUE #711: `get_polygon_to_contain` uses the deprecated frequency argument
`get_polygon_to_contain` uses the deprecated frequency argument:

The `get_polygon_to_contain` function and call to `get_edge_bbox_in_projection_coordinates`:

https://github.com/pytroll/pyresample/blob/60e6132d934491c1aa12ccd7382c3246f9d1eb5e/pyresample/slicer.py#L173

And the "PendingDeprecationWarning" in `get_edge_bbox_in_projection_coordinates`:

https://github.com/pytroll/pyresample/blob/60e6132d934491c1aa12ccd7382c3246f9d1eb5e/pyresample/geometry.py#L1681

jensdebruijn:
`get_polygon_to_contain` used the deprecated frequency argument:

The `get_polygon_to_contain` function and call to `get_edge_bbox_in_projection_coordinates`:

https://github.com/pytroll/pyresample/blob/60e6132d934491c1aa12ccd7382c3246f9d1eb5e/pyresample/slicer.py#L173

And the "PendingDeprecationWarning" in `get_edge_bbox_in_projection_coordinates`:

https://github.com/pytroll/pyresample/blob/60e6132d934491c1aa12ccd7382c3246f9d1eb5e/pyresample/geometry.py#L1681

 - [x] Closes #711 
 - [x] Tests passed

### DIFF
--- a/pyresample/future/geometry/_subset.py
+++ b/pyresample/future/geometry/_subset.py
@@ -100,7 +100,7 @@ def _get_area_boundary(area_to_cover: AreaDefinition) -> Boundary:
         if area_to_cover.is_geostationary:
             return Boundary(*get_geostationary_bounding_box_in_lonlats(area_to_cover))
         boundary_shape = max(max(*area_to_cover.shape) // 100 + 1, 3)
-        return area_to_cover.boundary(frequency=boundary_shape, force_clockwise=True)
+        return area_to_cover.boundary(vertices_per_side=boundary_shape, force_clockwise=True)
     except ValueError as err:
         raise NotImplementedError("Can't determine boundary of area to cover") from err
 
--- a/pyresample/slicer.py
+++ b/pyresample/slicer.py
@@ -170,7 +170,7 @@ class AreaSlicer(Slicer):
         from shapely.geometry import Polygon
 
         try:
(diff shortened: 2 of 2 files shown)


djhoese (commented):
If I remember correctly @mraspaud made the original switch to the newer `vertices_per_side` kwarg. @mraspaud could you verify that the changes made here make sense? Do you have any memory why they weren't changed when `vertices_per_side` was added originally?

jensdebruijn:
I did check the functions and `vertices_per_sides` and `frequency` are interpreted identically in the called functions ([here](https://github.com/pytroll/pyresample/blob/60e6132d934491c1aa12ccd7382c3246f9d1eb5e/pyresample/geometry.py#L436) and [here](https://github.com/pytroll/pyresample/blob/60e6132d934491c1aa12ccd7382c3246f9d1eb5e/pyresample/geometry.py#L1677)). So as far as I understand, this change wouldn't change behaviour at all, only stop the warning.

My knowledge of pyresample is quite limited, but what I understand from your comment is that perhaps they should be interpreted differently?

mraspaud (approved):
LGTM! Thanks for reducing the warning count!

PR #721, opened 2026-05-18
Use pyproject.toml fully instead of setup.py

### LINKED ISSUE #722: Cannot import unless being in the pyresample source root - when installing with pip (not using conda-forge)
I was going to test and work with the latest main, installed with `pip install  -e .` in a mamba environment with Python 3.12.
Surprised to see it fails to even import pyresample unless standing in the source root!

```
(python3.12) ➜  ~ ipython
Python 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.4.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: Use `--theme`, or the `%colors` magic to change IPython's themes and colors.

In [1]: from pyresample import geometry
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[1], line 1
----> 1 from pyresample import geometry

File ~/usr/src/forks/pyresample/pyresample/geometry.py:35
     32 from pyproj import Geod, Proj
     33 from pyproj.aoi import AreaOfUse
---> 35 from pyresample import CHUNK_SIZE
     36 from pyresample._spatial_mp import Cartesian, Cartesian_MP, Proj_MP
     37 from pyresample.area_config import create_area_def

ImportError: cannot import name 'CHUNK_SIZE' from 'pyresample' (unknown location)

In [2]: import pyresample

In [3]: pyresample.__file__

In [4]: from pyresample import CHUNK_SIZE
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[4], line 1
----> 1 from pyresample import CHUNK_SIZE

ImportError: cannot import name 'CHUNK_SIZE' from 'pyresample' (unknown location)
```

All the above behaved normally when standing in the source root (where the `setup.py` is placed) - including writing the path to the pyresample package when invoking `pyresample.__file__` (see above).

mraspaud:
Use pyproject.toml to entirely remove setup.py

This PR was generated mainly using AI

 - [x] Tests passed 
 - [x]  Closes #722


pyproject.toml
```diff
@@ -1,10 +1,105 @@
...
+dependencies = ["hatch-cython"]
+
+[tool.hatch.build.targets.wheel.hooks.cython.options]
+include_numpy = true
+compile_py = false
+directives = { language_level = 3, freethreading_compatible = true }
+define_macros = [["NPY_NO_DEPRECATED_API", "NPY_1_25_API_VERSION"]]
+compile_args = [
+  { platforms = ["linux", "darwin", "freebsd"], arg = "-O3" },
+  { platforms = ["linux", "darwin", "freebsd"], arg = "-Wno-unused-function" },
```
djhoese, line 98: These 2 lines can't be combined?

djhoese (approved):
The unstable environment in CI might need to be updated to not install versioneer. And maybe the RTD config too.

I'm fine with this, but it does scare me for any further customizations that we may want. For example, I think pykdtree's setup.py has a lot of custom stuff depending on what it finds in the environment (ex. OpenMP) so I'm worried about what pyresample would have to do to handle a similar case.

adybbroe:
This saved my (PCW) week, thanks Martin!
I will make an issue shortly...

pyproject.toml
mraspaud: My understanding is that hatch-cython wants them separate

adybbroe:
Should close #722 (edited the pre-amble reflecting that - see above)

mraspaud:
> The unstable environment in CI might need to be updated to not install versioneer. And maybe the RTD config too.
> 
> I'm fine with this, but it does scare me for any further customizations that we may want. For example, I think pykdtree's setup.py has a lot of custom stuff depending on what it finds in the environment (ex. OpenMP) so I'm worried about what pyresample would have to do to handle a similar case.

Noted! I think we will be able to fix this though, so we'll go ahead and merge this.

adybbroe (approved):
LGTM

PR #724, opened 2026-05-20
Add support to take bounding box info from metadata passed from the level-1 reader

adybbroe:
Add support to take bounding box info from metadata in the area object instead of calculating it from the geolocation data

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2010-2025 Pyresample developers
+# Copyright (C) 2010-2026 Pyresample developers
 #
 # This program is free software: you can redistribute it and/or modify it under
 # the terms of the GNU Lesser General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


pyresample/geometry.py
```diff
@@ -916,9 +942,13 @@ def _compute_omerc_parameters(self, ellipsoid):
...
         if any(np.isnan((lon1, lon2, lat1, lat, lat2))):
             thelons = self.lons[:, int(cols / 2)]
-            thelons = thelons.where(thelons.notnull(), drop=True)
+            mask = thelons.notnull().compute()
+            thelons = thelons.where(mask, drop=True)
+
             thelats = self.lats[:, int(cols / 2)]
-            thelats = thelats.where(thelats.notnull(), drop=True)
+            mask = thelats.notnull().compute()
+            thelats = thelats.where(mask, drop=True)
```
djhoese: It would be good to do the compute at the same time like `thelons, thelats = da.compute(thelons.notnull(), thelats.notnull())`. That way any shared computation tasks are only run once.

adybbroe: > It would be good to do the compute at the same time like `thelons, thelats = da.compute(thelons.notnull(), thelats.notnull())`. That way any shared computation tasks are only run once.

Fixed, thanks

PR #725, opened 2026-05-20
Fix EWA ll2cr not handling non-C contiguous arrays

### LINKED ISSUE #723: EWA resampler fails (with METimage/VII data) with non-contiguous lonlat arrays
The EWA resampler currently fails when trying to resample a METimage/VII swath.
To reproduce:
```python
scn = Scene(filenames=filenames, reader='vii_l1b_nc')
scn.load(["vii_10690"])
scn_r = scn.resample("EPSG_4326_720x360", resampler='ewa', rows_per_scan=24)
scn_r.save_datasets(writer='simple_image')
```
results in
```bash
  File "/tcenas/home/andream/anaconda3/envs/devenv/lib/python3.11/site-packages/dask/base.py", line 662, in compute
    results = schedule(dsk, keys, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tcenas/home/andream/code/satpy_latest/pyresample/pyresample/ewa/dask_ewa.py", line 60, in _call_ll2cr
    swath_points_in_grid, cols, rows = ll2cr(new_src, target_geo_def)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tcenas/home/andream/code/satpy_latest/pyresample/pyresample/ewa/ewa.py", line 78, in ll2cr
    swath_points_in_grid = _ll2cr.ll2cr_static(lons, lats, fill,
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyresample/ewa/_ll2cr.pyx", line 242, in pyresample.ewa._ll2cr.ll2cr_static
  File "<stringsource>", line 664, in View.MemoryView.memoryview_cwrapper
  File "<stringsource>", line 352, in View.MemoryView.memoryview.__cinit__
ValueError: ndarray is not C-contiguous
```
A simple fix seems to be adding the following lines after this block (where the lonlats are already computed ndarrays) https://github.com/pytroll/pyresample/blob/5d91b3aa66743eedaa63558965ce502066af851e/pyresample/ewa/ewa.py#L59-L68
```python
    lons = np.ascontiguousarray(lons)
    lats = np.ascontiguousarray(lats)
```

Is this a good idea? Or is there a better way?

Fixing it at the source without breaking lazyness is probably not trivial, due to the lon-lats of METimage being computed with rather complex geotiepoints interpolations and coordinate transforms... 

cc @cesclc

djhoese:
CC @ameraner 

In #723 it was pointed out that non-C contiguous lon/lat inputs to the EWA ll2cr function are not handled. This PR fixes this by specifying the `order` kwarg to `astype` as well as using `numpy.require` to force write-ability on the inputs. I think this preserves the existing functionality while handling more use cases.

The main complication here is that old ll2cr has an optimization if some users want to use it where results are written in-place on the lon/lat inputs to avoid making extra arrays. This of course completely overwrites the lon/lat data in the arrays and is why the default is to copy the data, but I'd still like to support this behavior if it can be used in the future.

 - [x] Closes #723 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/ewa/ewa.py
+++ b/pyresample/ewa/ewa.py
@@ -53,18 +53,18 @@ def ll2cr(swath_def, area_def, fill=np.nan, copy=True):
 
         ll2cr uses the pyproj library which is limited to 64-bit float
         navigation arrays in order to not do additional copying or casting
-        of data types.
+        of data types. Additionally, this function requires input arrays to
+        be C-contiguous and writable. Therefore, to write results in-place
+        into the lons/lats of the provided SwathDefinition as an optimization,
+        the longitude and latitude array inputs must be C-contiguous 64-bit
+        floating point writable arrays.
 
     """
     lons, lats = swath_def.get_lonlats()
     # ll2cr requires 64-bit floats due to pyproj limitations
-    # also need a copy of lons, lats since they are written to in-place
-    try:
-        lons = lons.astype(np.float64, copy=copy)
-        lats = lats.astype(np.float64, copy=copy)
-    except TypeError:
-        lons = lons.astype(np.float64)
-        lats = lats.astype(np.float64)
+    # also need a contiguous writable copy of lons, lats since they are written to in-place
+    lons = np.require(lons.astype(np.float64, copy=copy, order="C"), requirements=["C", "W"])
+    lats = np.require(lats.astype(np.float64, copy=copy, order="C"), requirements=["C", "W"])
 
     # Break the input area up in to the expected parameters for ll2cr
     p = area_def.crs_wkt if hasattr(area_def, 'crs_wkt') else area_def.proj_str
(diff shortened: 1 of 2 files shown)


pyresample/ewa/ewa.py
```diff
@@ -53,18 +53,18 @@ def ll2cr(swath_def, area_def, fill=np.nan, copy=True):
...
-        lats = lats.astype(np.float64, copy=copy)
-    except TypeError:
-        lons = lons.astype(np.float64)
-        lats = lats.astype(np.float64)
+    # also need a contiguous writable copy of lons, lats since they are written to in-place
+    lons = np.require(lons.astype(np.float64, copy=copy, order="C"), requirements=["C", "W"])
+    lats = np.require(lats.astype(np.float64, copy=copy, order="C"), requirements=["C", "W"])
 
     # Break the input area up in to the expected parameters for ll2cr
     p = area_def.crs_wkt if hasattr(area_def, 'crs_wkt') else area_def.proj_str
```
ameraner: Just a side note, this variable seems to be unused.

ameraner (approved):
Thank you very much for the quick fix! It solves my issue.

pyresample/ewa/ewa.py
djhoese: Thanks. Fixed. A while ago I changed from the ll2cr code taking projection strings to taking the actual pyproj CRS object and this must have been left over.

PR #726, opened 2026-05-21
Add resample `force` kwarg

### LINKED ISSUE #660: Pyresample cannot be used for gapfilling due to `geometries_are_the_same` check
I have a use case where I would like to use the nearest neighbour approach to fill some gaps in my data. In other words, I would like to resample a Scene to its own same area again. If I do this, currently the resampling is skipped due to:
https://github.com/pytroll/pyresample/blob/3c072554474a7d0087e98b8c3c3396557d631ea9/pyresample/resampler.py#L121-L122
for optimisation purposes.

A solution could be to add a kwarg `skip_if_same_geometry=True` to `resample`, in order to be able to force the resampling when `False` is passed.

ameraner:
This PR adds a `force` kwarg to the `resampler.resample` call. When set to `True`, it bypasses the area equality checks that skip the resampling if the geometries are the same. This is useful e.g. when data has NaN gaps (for example after a parallax correction) and one wants to gapfill it, e.g. with nearest neighbour.

 - [x] Closes #660 
 - [x] Tests added 
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/pyresample/resampler.py
+++ b/pyresample/resampler.py
@@ -95,7 +95,7 @@ class BaseResampler:
         """
         raise NotImplementedError
 
-    def resample(self, data, cache_dir=None, mask_area=None, **kwargs):
+    def resample(self, data, cache_dir=None, mask_area=None, force=False,**kwargs):
         """Resample `data` by calling `precompute` and `compute` methods.
 
         Only certain resampling classes may use `cache_dir` and the `mask`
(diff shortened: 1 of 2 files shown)


djhoese (approved):
Looks good to me. I can't help but think this smells like we're doing something wrong in Satpy's design that this is needed.

djhoese (approved):
I was pretty sure when I last reviewed this there was another location I found where this equality/comparison needs to be forced, but now I can't find it. It may have just been that we'll need to keep this in mind when I do further work on the `pyresample.future` resampler stuff.

PR #729, opened 2026-06-29
Densify the geostationary bounding polygon before reprojecting in AreaSlicer

### LINKED ISSUE #728: gradient_search leaves part of the destination empty for a partial geostationary disk
**Describe the bug**

When the source is a partial geostationary disk (for example MSG SEVIRI rapid scan data, which covers only the northern part of the disk, or a region of interest), resampling to an area in a different projection with `gradient_search` leaves a large part of the destination empty even where the source clearly has data. The populated region is bounded by a smooth arc.

The cause is in `AreaSlicer.get_polygon_to_contain` (`pyresample/slicer.py`). For a geostationary source it builds the bounding polygon of the disk and reprojects it into the destination projection to compute the source slice. When the disk is clipped to a partial area extent, the clip introduces straight chord edges (for instance the southern boundary of a rapid scan strip), each represented by only its two end points. Those chords are not straight in the destination projection, but with only two vertices the reprojected polygon keeps them straight and cuts the corner. The resulting source slice is too small, so `resample_blocks` fills the uncovered destination pixels with the fill value. A full disk source is unaffected, because its whole boundary is the densely sampled disk arc with no straight chord.

**To Reproduce**

```python
from pyresample.slicer import create_slicer

# Partial geostationary disk: the northern 1392 lines of an MSG rapid scan image
source = AreaDefinition(
    "rss", "rss", None,
    {"a": 6378169.0, "b": 6356583.8, "h": 35785831.0, "lon_0": 9.5,
     "proj": "geos", "units": "m"},
    3712, 1392,
    (5568748.2758, 5568748.2758, -5568748.2758, 1392187.0689))

# Polar stereographic destination over Europe, fully inside the disk
euro4 = AreaDefinition(
    "euro4", "euro4", None,
    {"proj": "stere", "ellps": "bessel", "lat_0": 90.0, "lon_0": 14.0,
     "lat_ts": 60.0, "units": "m"},
    1024, 1024,
    (-2717181.7304994687, -5571048.14031214,
     1378818.2695005313, -1475048.1403121399))

x_slice, y_slice = create_slicer(source, euro4).get_slices()
print(y_slice)
```

**Expected behavior**

The whole `euro4` area falls inside the rapid scan strip, so the source slice should cover every source row the destination needs (down to about row 694), and a `gradient_search` resample should fill the whole destination.

**Actual results**

```
slice(1151, 1335, None)
```

Only the northern rows are kept (the slice should start near row 694). A `gradient_search` resample of this source to `euro4` then fills only about 44 percent of the destination, with the populated area bounded by an arc.

**Environment Info:**

 - pyresample: reproduced with 1.34.2 and 1.35.0 (the relevant code is unchanged on `main`)
 - shapely: 2.x
 - pyproj: 3.x

**Additional context**

Reported downstream in satpy as pytroll/satpy#3403. A fix that densifies the bounding polygon with `segmentize` before reprojecting it makes the example above fill the whole destination. Happy to open a pull request.

lorenzo-cingano:
`AreaSlicer.get_polygon_to_contain` builds the bounding polygon of a geostationary source disk and reprojects it into the destination projection to compute the source slice. When the source is a partial disk (rapid scan or region of interest data), clipping the disk to its area extent introduces straight chord edges with only two vertices. Those chords are not straight in the destination projection, so the reprojected polygon cuts the corner and the source slice comes out too small, leaving part of the destination empty. A full disk source is unaffected, because its whole boundary is the densely sampled disk arc.

This densifies the polygon with `segmentize` before reprojecting it, so the chord edges keep enough vertices to follow their true shape in the destination projection. The problem was found in satpy when resampling MSG rapid scan data to a polar stereographic area with `gradient_search`.

`segmentize` needs shapely 2.0 or newer. shapely is currently unpinned in the dependencies, so a `shapely>=2.0` pin can be added if you want it explicit.

 - [x] Closes #728
 - [x] Tests added
 - [x] Tests passed
 - [x] Fully documented

### DIFF
--- a/pyresample/slicer.py
+++ b/pyresample/slicer.py
@@ -175,7 +175,18 @@ class AreaSlicer(Slicer):
             x, y = self.area_to_contain.get_edge_lonlats(vertices_per_side=10)
         if self.area_to_crop.is_geostationary:
             x_geos, y_geos = get_geostationary_bounding_box_in_proj_coords(self.area_to_crop, 360)
-            x_geos, y_geos = self._source_transformer.transform(x_geos, y_geos, direction=TransformDirection.INVERSE)
+            geos_poly = Polygon(zip(x_geos, y_geos, strict=True))
+            # Densify the polygon before reprojecting it to the destination crs.
+            # Clipping a partial disk (for example rapid scan or region of interest
+            # data) to its area extent introduces straight chord edges. Those chords
+            # are not straight in the destination crs, so without intermediate
+            # vertices the reprojected polygon cuts the corner and the computed slice
+            # is too small, leaving part of the destination without data.
+            segment_length = np.max(np.abs(self.area_to_crop.area_extent)) / 100
+            geos_poly = geos_poly.segmentize(segment_length)
+            x_geos, y_geos = geos_poly.exterior.coords.xy
+            x_geos, y_geos = self._source_transformer.transform(
+                np.asarray(x_geos), np.asarray(y_geos), direction=TransformDirection.INVERSE)
             geos_poly = Polygon(zip(x_geos, y_geos, strict=True))
             poly = Polygon(zip(x, y, strict=True))
             poly = poly.intersection(geos_poly)
--- a/pyresample/test/test_slicer.py
+++ b/pyresample/test/test_slicer.py
@@ -64,6 +64,25 @@ class TestAreaSlicer(unittest.TestCase):
         assert x_slice.start > 0 and x_slice.stop < 100
         assert y_slice.start > 0 and y_slice.stop >= 100
 
+    def test_partial_geostationary_disk_covering_dest_area_is_not_truncated(self):
+        """Test that a partial geostationary disk covering the destination is not truncated.
+
+        The southern boundary of a partial disk is a straight chord in the source
+        projection. It has to be densified before being reprojected to the
+        destination crs, otherwise the computed source slice stops short and part
+        of the destination is left without data.
+        """
+        src_area = AreaDefinition('rss', 'rss area', None,
+                                  {'ellps': 'WGS84', 'h': '35785831', 'proj': 'geos', 'lon_0': 9.5},
+                                  3712, 1392,
+                                  (5550000.0, 5550000.0, -5550000.0, 2000000.0))
+        slicer = create_slicer(src_area, self.dst_area)
+        x_slice, y_slice = slicer.get_slices()
+        # The whole destination falls inside the strip; its southern edge maps to
+        # source row 584, so the slice must reach well below the strip top rows.
+        assert y_slice.start < 700
+        assert y_slice.stop >= 1328
+
     def test_source_area_does_not_cover_dest_area_at_all(self):
         """Test source area does not cover dest area at all."""
         src_area = AreaDefinition('dst', 'dst area', None,


djhoese:
Thanks for this PR! I just wanted to let you know that we're not ignoring you, but most of us are on vacation or (like me) don't have time to review this right now. I hope to review it next week though.

Manny7717 (commented):
## Review

Verified locally on head `99f9c93c`: **19/19 tests in pyresample/test/test_slicer.py pass** (shapely 2.1.2). The approach is sound — densifying in the source projection before the inverse transform is exactly where the straight chord edges need intermediate vertices, since chords in the geos projection become curves in the destination CRS.

Two things to look at (inline): a Shapely minimum-version concern for `segmentize`, and a degenerate-extent crash in the `segment_length` computation. Both are cheap to guard.

pyresample/slicer.py
```diff
@@ -175,7 +175,18 @@ def get_polygon_to_contain(self):
...
-            x_geos, y_geos = self._source_transformer.transform(x_geos, y_geos, direction=TransformDirection.INVERSE)
+            geos_poly = Polygon(zip(x_geos, y_geos, strict=True))
+            # Densify the polygon before reprojecting it to the destination crs.
+            # Clipping a partial disk (for example rapid scan or region of interest
+            # data) to its area extent introduces straight chord edges. Those chords
+            # are not straight in the destination crs, so without intermediate
+            # vertices the reprojected polygon cuts the corner and the computed slice
+            # is too small, leaving part of the destination without data.
+            segment_length = np.max(np.abs(self.area_to_crop.area_extent)) / 100
+            geos_poly = geos_poly.segmentize(segment_length)
```
Manny7717, line 186: `segmentize` requires Shapely >= 2.0 (GEOS >= 3.10), but `pyproject.toml` declares just `"shapely"` with no minimum. On a Shapely 1.8 install this raises AttributeError. Worth confirming the minimum supported Shapely version and either pinning it or guarding with a hasattr/fallback.

pyresample/slicer.py
```diff
@@ -175,7 +175,18 @@ def get_polygon_to_contain(self):
...
             x_geos, y_geos = get_geostationary_bounding_box_in_proj_coords(self.area_to_crop, 360)
-            x_geos, y_geos = self._source_transformer.transform(x_geos, y_geos, direction=TransformDirection.INVERSE)
+            geos_poly = Polygon(zip(x_geos, y_geos, strict=True))
+            # Densify the polygon before reprojecting it to the destination crs.
+            # Clipping a partial disk (for example rapid scan or region of interest
+            # data) to its area extent introduces straight chord edges. Those chords
+            # are not straight in the destination crs, so without intermediate
+            # vertices the reprojected polygon cuts the corner and the computed slice
+            # is too small, leaving part of the destination without data.
+            segment_length = np.max(np.abs(self.area_to_crop.area_extent)) / 100
```
Manny7717, line 185: If `area_extent` is degenerate (e.g. all zeros), `segment_length` becomes 0 and `segmentize(0)` raises `GEOSException: Tolerance must be positive` (verified against shapely 2.1.2). A `if segment_length > 0:` guard (or `max(segment_length, 1e-9)`) would make this robust. Also: with the 1/100 divisor, a very small partial-disk extent produces a fine mesh — worth a sanity bound if performance matters.

pyresample/slicer.py
djhoese: Shapely 2.0 came out in 2022. At this point someone would have to try pretty hard to get shapely 1.x. That said, I wouldn't mind the version limit being specified in `pyproject.toml` @lorenzo-cingano if you could make that change.

pyresample/slicer.py
```diff
@@ -175,7 +175,18 @@ def get_polygon_to_contain(self):
...
+            geos_poly = geos_poly.segmentize(segment_length)
+            x_geos, y_geos = geos_poly.exterior.coords.xy
+            x_geos, y_geos = self._source_transformer.transform(
+                np.asarray(x_geos), np.asarray(y_geos), direction=TransformDirection.INVERSE)
```
djhoese, line 189: I'm curious if pyproj can automatically handle this `np.asarray`. It should recognize that the x_geos and y_geos objects support the buffer interface and use them accordingly. Would you mind trying to remove the `np.asarray` here and see if it still behaves?

pyresample/slicer.py
djhoese: All `area_extent` values being 0 or being a very small distance is a much larger issue. An error from libgeos being raised seems reasonable to me. That said, I think this is the wrong calculation. You are trying to determine the maximum segment length and then break it into 100 parts, right? The area extent doesn't tell you. The difference of the x and y extents does and I don't think numpy has to be used for that calculation and more likely is that it is actually worse. So...

```suggestion
            atc_extent = self.area_to_crop.area_extent
            segment_length = max(abs(atc_extent[2] - atc_extent[0]), abs(atc_extent[3] - atc_extent[1])) / 100
```

Let me know if I'm missing something.

PR #735, opened 2026-07-18
Add explicit semi-minor axes for spherical projections

### LINKED ISSUE #200: Fix incomplete areas in sample area configurations
There are various examples of how to define an AreaDefinition throughout the documentation and in test files like https://github.com/pytroll/pyresample/blob/master/pyresample/test/test_files/areas.yaml. However, some of these have projection definitions with only `a` defined (equatorial radius). This is technically incomplete and although it works in PROJ, rasterio/GDAL are a little more strict and will fail. We've run in to this a couple times with Satpy when trying to save geotiffs. We don't get an error throughout resampling and other coordinate transformations, but saving to geotiffs fails.

All area definitions (even in tests) should no longer define only `a` and should be discouraged as a valid use. Any existing projections should define `b` so that it equals `a` which I think is the default in PROJ/pyproj (@snowman2 do you know?).

floze-the-genius:
## Summary

- complete projection dictionaries that specify only `a` by copying them and adding an explicit matching `b`
- preserve projections that already define an ellipsoid or sphere through `b`, `R`, `ellps`, `datum`, `rf`, `f`, `es`, or `e`
- update the shipped YAML and legacy area fixtures so their spherical definitions are explicit
- add regression coverage for normalization, input immutability, the `create_area_def` integration path, and existing ellipsoid parameters
- keep the Python 3.11 mypy analysis compatible with current dependency drift by running its isolated hook on Python 3.12 with `numpy<2.3`

This keeps the historical spherical interpretation while ensuring stricter CRS consumers receive a complete ellipsoid definition. The mypy compatibility constraint is limited to pre-commit's isolated environment and does not change pyresample's runtime dependencies or Python 3.11 analysis target.

## Verification

- `JOBS=8 .venv/bin/pytest -q pyresample/test/test_area_config.py` (`22 passed`)
- `JOBS=8 .venv/bin/pytest -q pyresample/test` (`1166 passed, 24 skipped, 2 subtests passed`)
- `.venv/bin/ruff check pyresample/area_config.py pyresample/test/test_area_config.py`
- isolated `pre-commit run --all-files` (Ruff, whitespace, EOF, YAML, Bandit, mypy, isort)
- `git diff --check`

- [x] Closes #200
- [x] Tests added
- [x] Tests passed
- [x] Fully documented

AI disclosure: OpenAI Codex assisted with investigation, implementation, and test execution; the resulting diff and test outputs were reviewed before submission.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,6 +21,7 @@ repos:
   rev: 'v2.1.0'  # Use the sha / tag you want to point at
   hooks:
     - id: mypy
+      language_version: python3.12
       additional_dependencies:
         - types-docutils
         - types-PyYAML
@@ -28,7 +29,7 @@ repos:
         - type_extensions
         - types-setuptools
         # Typed libraries
-        - numpy
+        - numpy<2.3
         - pytest
       args: [ --warn-unused-configs ]
 - repo: https://github.com/pycqa/isort
--- a/pyresample/area_config.py
+++ b/pyresample/area_config.py
@@ -501,6 +501,7 @@ def create_area_def(area_id, projection, width=None, height=None, area_extent=No
     # (hold on to EPSG code as much as possible)
     if isinstance(projection, dict) and 'EPSG' in projection:
         projection = "EPSG:{}".format(projection['EPSG'])
+    projection = _add_missing_spherical_axis(projection)
 
     try:
         crs = _get_proj_data(projection)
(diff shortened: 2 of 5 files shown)


djhoese:
Thanks for creating this PR. It brings up an odd issue and I'm not sure how I feel about fixing it. The description you made talks about "stricter CRS consumers", but do you have any examples of what/who those are? I hesitate to include this in pyresample as it feels like we are defining/forcing defaults that the PROJ C library may be filling in on its own. I do realize PROJ is probably always defining it as you have here.

floze-the-genius:
You are right to question this. The concrete consumer I had in mind came from the original 2019 report in #200: Satpy GeoTIFF export through rasterio/GDAL failed even though PROJ accepted the projection. I re-ran that path against the current environment here (pyproj 3.7.2 and rasterio 1.5.0): both the `a`-only and explicit `a`/`b` definitions normalize to the same `+R=6371228` CRS (EPSG:3409), and both write and reopen a GeoTIFF successfully.\n\nSo I do not have a current stricter consumer that justifies adding runtime normalization in `create_area_def`; that part of this PR is broader than the evidence supports and duplicates behavior PROJ already provides. I am closing this rather than forcing a default based on a historical failure. If a fixture/docs cleanup is still wanted independently, it would be better handled as a narrower change with current acceptance criteria.

djhoese:
Thanks for the clarification. In the future, while I can understand if the English language is a barrier, it would be nice to have a conversation with a person rather than an AI.

PR #740, opened 2026-08-28
fix(area): complete area definitions with explicit b parameter

Manny7717:
## What

Completes all incomplete area definitions in the repository. Several example projections (test fixtures and documentation) only define the equatorial radius `a`; while PROJ accepts this and treats the ellipsoid as a sphere, stricter tools such as **GDAL reject these definitions when writing geotiffs** (reported in #200, encountered by Satpy users).

## Change

Add `b = a` to every projection that only defined `a`:

- `pyresample/test/test_files/areas.yaml` — 13 area definitions
- `pyresample/test/test_geometry/test_area.py` — projection dicts and PROJ.4 strings used as test fixtures
- Docs: `docs/areas.cfg`, and howto examples in `geo_def.rst`, `geometry_utils.rst`, `plot.rst`, `plot_cartopy_basemap.rst`, `plot_projections.rst`, `spherical_geometry.rst` (dict, PROJ.4 string, YAML and `.cfg` forms)
- Added a note in `geo_def.rst` documenting that both `a` and `b` should always be defined

## Why it's safe

`a`-only and `a+b` (with `b = a`) definitions normalize to the **identical** CRS in pyproj (`+R=6371228` sphere form, `CRS.equals()` → `True`), so no coordinate behavior changes — this only removes the GDAL incompatibility.

## Tests

- `python -m pytest pyresample/test/test_area_config.py pyresample/test/test_geometry/test_area.py` → **221 passed** (same as before the change)
- All `.. doctest::` blocks in the edited howtos still pass unchanged (verified expected outputs are byte-identical)
- All YAML code blocks in edited docs parse with `yaml.safe_load`
- `ruff check` clean on the modified test file

### DIFF
--- a/docs/areas.cfg
+++ b/docs/areas.cfg
@@ -237,7 +237,7 @@ REGION: euro_north {
 REGION: ease_sh {
        NAME:           Antarctic EASE grid
        PCS_ID:         ease_sh
-       PCS_DEF:        proj=laea, lat_0=-90, lon_0=0, a=6371228.0, units=m
+       PCS_DEF:        proj=laea, lat_0=-90, lon_0=0, a=6371228.0, b=6371228.0, units=m
        XSIZE:          425
        YSIZE:          425
        AREA_EXTENT:    (-5326849.0625,-5326849.0625,5326849.0625,5326849.0625)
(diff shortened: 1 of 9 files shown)


djhoese:
This is a duplicate of #735. As commented there the original case I described in #200 doesn't seem to error out any more so these changes may not be necessary. Similarly pyresample has been using pyproj CRS objects so a lot of this a/b and other projection parameter resolution has been pushed to pyproj to resolve and normalize. Additionally PROJ.4 strings are relatively deprecated and should be relied on as little as possible once they are converted to a CRS object (stick with WKT definitions instead).

I'm going to close this and the #200 issue for now. If one of the other maintainers or power users of pyresample wants to present a strong opinion on this matter then go for it and I will happily reconsider. As the original issue creator for #200 though I'm comfortable closing this for now. Thanks @Manny7717 for making the PR but I'm choosing to not merge this at least for now.

PR #741, opened 2026-08-28
Fix user fill value not being used in EWA

Manny7717:
## What

`fornav()` resolves a fill value from the input data (NaN for floats, -999 for integers, or the caller's `fill` argument) but then passed hardcoded `np.nan` as both the input and output fill to the C resampler (`_fornav.fornav_wrapper`). Two consequences (reported in #689):

- Integer swath data cannot be resampled — NaN cannot be written into an integer output array.
- A user-specified `fill` (e.g. `fill=0`) was ignored for float data: uncovered pixels still got NaN.

## Change

Pass the resolved `fill` value as both input and output fill to `fornav_wrapper`. Default behavior is unchanged (float data with `fill=None` still gets NaN, matching the documented default); only `fill`-specified and integer paths change. The dask EWA resampler already forwarded its fill value and is unaffected.

## Tests

- New `test_fornav_integer_swath_with_fill`: int8 swath + `fill=0` resamples without error; output contains only data and fill values.
- New `test_fornav_fill_value_used_in_output`: float32 swath + `fill=0` — uncovered pixels are 0, no NaN anywhere.
- Both tests fail on the previous implementation (verified) and pass with the fix.
- `pytest test_ewa_fornav.py test_dask_ewa.py` → 89 passed, 20 skipped (pre-existing skips). ruff clean.

### DIFF
--- a/pyresample/ewa/ewa.py
+++ b/pyresample/ewa/ewa.py
@@ -167,7 +167,7 @@ def fornav(cols, rows, area_def, data_in,
     rows_per_scan = rows_per_scan or data_in[0].shape[0]
 
     results = _fornav.fornav_wrapper(cols, rows, data_in, out,
-                                     np.nan, np.nan, rows_per_scan,
+                                     fill, fill, rows_per_scan,
                                      weight_count=weight_count,
                                      weight_min=weight_min,
                                      weight_distance_max=weight_distance_max,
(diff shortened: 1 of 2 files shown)


djhoese (changes requested):
Thanks for fixing this. The fix looks good. Could you please combine the two new tests outside of the unittest TestCase class and use pytest's parametrize? This will also require changing the assert method calls to normal `assert X` conditions.

djhoese:
Thanks for this. I'm sorry it took so long to merge such an obvious change. Other EWA resampling PRs that were merged made some conflicts. Claude also pointed out to me that the integer fill value was out of the range for the integer dtype (int8, default was -999.0). I've tacked on a fix for that to this PR. I rebased your branch and force pushed it. This should be good to merge once CI finishes. Thanks again.

djhoese (approved):

PR #742, opened 2026-08-28
docs: replace deprecated ImageContainer example in swath howto

### LINKED ISSUE #632: Documentation - ImageContainer is deprecated and should not appear as the default in RTD pages

```python
# Your code here

```
#### Problem description

[this should also explain **why** the current behaviour is a problem and why the
expected output is a better solution.]

`.../pyresample/image.py:252: FutureWarning: Usage of ImageContainer is deprecated, please use NumpyResamplerBilinear class instead`

Manny7717:
## What

Fixes #632: the RTD swath howto's resampling example used `image.ImageContainerNearest`, which emits `FutureWarning: Usage of ImageContainer is deprecated` whenever the example runs. The `pyresample.image` module is deprecated in favor of `pyresample.kd_tree` / `pyresample.bilinear` (and the newer xarray-based resamplers).

## Change

- `docs/source/howtos/swath.rst`: replaced the `ImageContainerNearest` example with the modern `KDTreeNearestXarrayResampler` (dask/xarray-based API), fixed the `ImageContanerBilinear` typo, and updated the remaining prose reference to the deprecated class.
- `docs/source/howtos/preproc.rst`: added the module deprecation note already present in `grid.rst`.

## Verification

- The new doctest block was extracted and run with `python -m doctest` under `-W error` (warnings-as-errors): passes, no deprecation warnings.
- `result.shape` in the doctest asserts the (800, 800) output grid.

### DIFF
--- a/docs/source/howtos/preproc.rst
+++ b/docs/source/howtos/preproc.rst
@@ -3,6 +3,10 @@
 Preprocessing of grids
 ======================
 
+.. note::
+   The pyresample.image module is deprecated.  Please use `pyresample.kd_tree` or `pyresample.bilinear`
+   instead.
+
 When resampling is performed repeatedly to the same grid significant execution time can be save by
 preprocessing grid information.
 
--- a/docs/source/howtos/swath.rst
+++ b/docs/source/howtos/swath.rst
@@ -17,12 +17,16 @@ Resampling can be done using nearest neighbour method, Guassian weighting, weigh
 
 pyresample.image
 ----------------
-The ImageContainerNearest and ImageContanerBilinear classes can be used for resampling of swaths as well as grids.  Below is an example using nearest neighbour resampling.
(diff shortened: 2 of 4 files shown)


djhoese (commented):
It looks like your EWA fix commit is in this PR too. Could you fix it so it is only the documentation changes?

Otherwise, I'm not sure how I feel about these changes. I didn't realize we had the deprecation warnings on the documented ways of doing things until they were already released, but I haven't had time to do anything about it. I'm also not sure telling users to use the "future" classes is the right way to go, but I don't like the other "legacy" Resampler-based classes that much either. The updated examples in this PR also switch from using plain numpy to using dask and xarray which I think is not the right way to go. Users will still want to use numpy.

I think my preference is for this PR to sit for now while the rest of pyresample gets figured out and we can finalize all of this documentation and use this PR as a basis. I'll let other maintainers comment if they have opinions.

PR #747, opened 2026-09-09
Add meson-python dep to unstable CI

djhoese:
CI is failing in the unstable environment. I think it just needs this missing dependency (meson-python) for the mesonpy build backend.

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented


djhoese:
Ok the unstable CI gets further but still fails as it is waiting on https://github.com/SciTools/cartopy/pull/2690 to be released.

PR #749, opened 2026-09-11
Deprecate 'get_area_extent_for_subset' and fix incorrect return value

djhoese:
Sphinx was complaining about bad type definitions in the docstring of 'get_area_extent_for_subset'. I decided to move the types to normal type annotations since we use that elsewhere. This revealed that this method hasn't returned a tuple of scalars since 2018. This method was added in 2018 in #44 and then later we changed how the `get_proj_*` methods work and dropped this scalar use case. Since no tests were ever added for this method the breakage was never caught. In this PR I fix the type annotations and deprecate the method in favor of the slicing logic available for `AreaDefinition` to create a subset area and then get the area extent. Oh yeah, Claude pointed out that this method was on the `BaseDefinition` but was using methods that only existed on the `AreaDefinition`.

I also added type annotations to a helper function when I was confused about how I wanted to replace the internals of this method.

Bottom line: This method has been broken for years, it is being removed.

CC @meteoswiss-mdr who originally added in in #44 (Ulrich)

 - [ ] Closes #xxxx 
 - [ ] Tests added 
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/pyresample/geometry.py
+++ b/pyresample/geometry.py
@@ -192,34 +192,6 @@ class BaseDefinition:
         """Test for approximate inequality."""
         return not self.__eq__(other)
 
-    def get_area_extent_for_subset(self, row_LR, col_LR, row_UL, col_UL):
-        """Calculate extent for a subdomain of this area.
-
-        Rows are counted from upper left to lower left and columns are
-        counted from upper left to upper right.
-
-        Args:
-            row_LR (int): row of the lower right pixel
-            col_LR (int): col of the lower right pixel
-            row_UL (int): row of the upper left pixel
-            col_UL (int): col of the upper left pixel
-
-        Returns:
-            area_extent (tuple):
-                Area extent (LL_x, LL_y, UR_x, UR_y) of the subset
-
-        Author:
-            Ulrich Hamann
-        """
-        (a, b) = self.get_proj_coords(data_slice=(row_LR, col_LR))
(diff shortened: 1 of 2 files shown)


pyresample/geometry.py
```diff
@@ -2380,6 +2357,43 @@ def get_xy_from_proj_coords(self, xm, ym):
 
         return self.get_array_indices_from_projection_coordinates(self, xm, ym)
 
+    def get_area_extent_for_subset(
+            self,
+            row_LR: int,
+            col_LR: int,
+            row_UL: int,
+            col_UL: int,
```
pnuu: Although this is easier to read than all lowercase, it still looks a bit weird to have mixed-case variable names.

pnuu (approved):
LGTM. One comment inline.

pyresample/geometry.py
djhoese: They have been lowercases but Claude pointed out that anyone using this and passing arguments as keyword arguments would get a failure. Overall not too big of a problem since it is being deprecated.

pnuu: But they're not kwargs but plain positional arguments? But yeah, deprecation and all.

djhoese: Unless you use the newer `def func(a, *, b):` any normal positional argument can be passed as `func(a=1)`.

pnuu: :astonished:

